PR #14, opened 2024-08-02
fix: add block height and time

stringhandler:


stringhandler:
DNM something is broken here. The app does not start

stringhandler:
> DNM something is broken here. The app does not start

Fixed. Still can't get it onto the frontend for some reason

PR #30, opened 2024-08-07
chore(ui): implement listening to outside app activity

Misieq01:
Description
---
Change autominer behavior so it would listen to user mouse movement outside app context

Motivation and Context
---

How Has This Been Tested?
---
Manually 

---
Turn on autominer and position mouse outside app window then be idle for 30 seconds, minining should by then start and if you move mouse it should stop

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/App.tsx
```diff
@@ -19,10 +19,31 @@ import useAppStateStore from './store/appStateStore.ts';
...
     const view = useUIStore((s) => s.view);
+    const setUserMousePosition = useUIStore((s) => s.setUserMousePosition);
+    const setError = useAppStateStore((s) => s.setError);
     const startupInitiated = useRef(false);
     const setSetupDetails = useAppStateStore((s) => s.setSetupDetails);
     const settingUpFinished = useAppStateStore((s) => s.settingUpFinished);
 
+    useEffect(() => {
+        const intervalId = setInterval(() => {
+            invoke('check_user_mouse_position')
```
kowalski, line 30: Please add a file like we have in v2: https://github.com/tari-project/tari-universe/blob/main/src/types/invoke.ts

This allows you to type custom handlers in one place

src/App.tsx
```diff
@@ -19,10 +19,31 @@ import useAppStateStore from './store/appStateStore.ts';
 function App() {
     const background = useUIStore((s) => s.background);
     const view = useUIStore((s) => s.view);
+    const setUserMousePosition = useUIStore((s) => s.setUserMousePosition);
+    const setError = useAppStateStore((s) => s.setError);
     const startupInitiated = useRef(false);
     const setSetupDetails = useAppStateStore((s) => s.setSetupDetails);
     const settingUpFinished = useAppStateStore((s) => s.settingUpFinished);
 
+    useEffect(() => {
```
kowalski, line 28: perhaps it would be good idea to limit amount of lines we have in App.tsx

You could easily move the lines you've added to a hook named `useUserMousePosition`

src/App.tsx
Misieq01: Good point, I forgot about that since we last talk about it. Adding it now

stringhandler (commented):
I'm not really a fan of doing this in javascript. How about rather having a task in rust check the mouse movement and then send an event to the javascript "User is idle". Then the autominer can carry on from there. You could probably actually just do the whole start/stop mining in rust as well

Misieq01:
> I'm not really a fan of doing this in javascript. How about rather having a task in rust check the mouse movement and then send an event to the javascript "User is idle". Then the autominer can carry on from there. You could probably actually just do the whole start/stop mining in rust as well

So you would like to leave only switching auto mining state in javascript and rest of the current logic move to rust ?

stringhandler (approved):
tested and working

PR #42, opened 2024-08-09
feat(ui): implement display for applications versions

Misieq01:
Description
---
Implement display for all versions of our applications

Motivation and Context
---
We want to easily see which versions are we running on

How Has This Been Tested?
---
Manually

---
Open settings tabs and you should see all versions

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler:
tested.

Please also add the actual version of the app as well (currently v0.1.26)

PR #61, opened 2024-08-20
feat: add timeout option to settings

Misieq01:
Description
---
-  Added new field in app config for user inactivity
- Added Number input component
- Added input field in settings for changing inactivity threshold

Motivation and Context
---
We wanted to add option for users to adjust time after they are considered inactive

How Has This Been Tested?
---
Manually

---
Go to settings and change value of timeout

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src-tauri/src/app_config.rs
```diff
@@ -48,6 +50,7 @@ impl AppConfig {
             config_file: None,
             mode: MiningMode::Eco,
             auto_mining: false,
+            user_inactivity_timeout: 5000,
```
stringhandler, line 53: let's make this higher

stringhandler: Also please use `Duration` type

stringhandler: or change to `user_inactivitiy_timeout_seconds`

stringhandler (approved):

PR #78, opened 2024-08-21
fix(modes): restart mining when mode changed

mmrrnn:
Description
---
Mode change had no implication on the running mining process. We need to restart mining.

Motivation and Context
---

How Has This Been Tested?
---
Manually, unfortunately I'm not able to attach screen recording since it does not work on my local along with mining

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -57,11 +57,13 @@ mod setup_status_event;
 #[tauri::command]
 async fn set_mode<'r>(
     mode: String,
-    _window: tauri::Window,
+    window: tauri::Window,
     state: tauri::State<'r, UniverseAppState>,
-    _app: tauri::AppHandle,
+    app: tauri::AppHandle,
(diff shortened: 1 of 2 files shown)


src-tauri/src/main.rs
```diff
@@ -57,11 +57,13 @@ mod setup_status_event;
 #[tauri::command]
 async fn set_mode<'r>(
     mode: String,
-    _window: tauri::Window,
+    window: tauri::Window,
     state: tauri::State<'r, UniverseAppState>,
-    _app: tauri::AppHandle,
+    app: tauri::AppHandle,
 ) -> Result<(), String> {
+    let _ = stop_mining(state.clone()).await;
```
stringhandler, line 64: Please add `warn!()` on any errors here

mmrrnn:
done

stringhandler (approved):

PR #90, opened 2024-08-21
fix(ui): handle mining when no internet connection

mmrrnn:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler:
tested. It kinda works, but not exactly what I was hoping for. Let's pause this for a bit and let me think of how best to do this

PR #108, opened 2024-08-22
feat(ui): floor height component

NovaT82:
Description
---
Moved the floor height to the right hand side of the screen and created the ruler component

Motivation and Context
---
UI changes as per figma

How Has This Been Tested?
---
Manually

---
Screenshot 2024-08-22 at 20 10 14

Breaking Changes
---
x

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


shanimal08 (approved):
lgtm

PR #139, opened 2024-08-26
fix(ui): copy seed words only when they are visible

mmrrnn:


mmrrnn:
[Screencast from 26.08.2024 10:34:23.webm](https://github.com/user-attachments/assets/3e33735a-6039-41f2-9aa6-31c029674807)

stringhandler (approved):

PR #185, opened 2024-08-27
fix(ui): clear animation triggers and display values on window focus

shanimal08:
Description
---

- add `contextmenu` listener to disable right click if not on local dev mode
- added missing `useEffect` and `useMemo` deps in `useMiningControls`
- added `persist` to UI store so `isMiningEnabled` isn't lost on reload (so we can actually remove the `contextmenu` listener if we want)
- updated `useVisualisation` so we only set the animation state if the window is not minimised, and focused or visible. (for fail and success states)
- neatened up `useBalanceInfo` a bit

Motivation and Context
---

failure and success block animations would run once returning to the app window after a while of being away since the animation states would be queued up


src/hooks/mining/useMiningControls.ts
```diff
@@ -86,30 +85,33 @@ export function useMiningControls() {
...
             console.info('useEffect: handleVisual stop');
-            handleVisual('stop');
-            isMiningInProgress.current = false;
+            handleVisual('stop').then(() => {
+                isMiningInProgress.current = false;
+            });
         }
 
         if (!isMining && isMiningInProgress.current) {
+            console.info('useEffect: handleVisual pause');
```
PanchoBubble, line 111:
```suggestion
```

PanchoBubble (approved):

src/hooks/mining/useMiningControls.ts
PanchoBubble: Duplicated console info here :P

shanimal08: https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExNWZ2aXAxcmh4NWowaWVrMjZ2eDhhbzNtazc5Mmpxa3Y2NXN0NXFrbiZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/qVVVfmHDMBZug/giphy.gif

PR #187, opened 2024-08-27
fix(ui): horizontal scroll on windows

mmrrnn:
Description
---
Scrollbar appears because font size is too large in the `BlockHeightBg` container - it's horizontal large text.
I tried to handle it as much responsive as possible - for different screen sizes and block number length.

Motivation and Context
---

How Has This Been Tested?
---
Tested on Windows Virtual Machine and Ubuntu

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


NovaT82:
The sizing looks good. The background number just needs to move to the right a bit. On the design the bottom of the text is 4px from the right hand side of the screen. Attaching the design here so you can see:
Success

NovaT82 (approved):
lgtm

mmrrnn:
i found out it can't be that close since the scroll appears

mmrrnn:
Text is now practically on the exact right. It just looks like numbers are spaced from the right, but take a look at the comma.

NovaT82:
Looks good :)
Could you just add a z-index to the ruler, so it doesn't go in underneath the background number?

PR #191, opened 2024-08-27
feat(airdrop): user token integration

PanchoBubble:
Description
---

Integration for token set and refresh

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/hooks/useGetStatus.ts
```diff
@@ -25,7 +25,7 @@ export function useGetStatus() {
             invoke('status')
                 .then((status) => {
                     if (status) {
-                        console.info('Status:', status);
+                        // console.info('Status:', status);
```
BalazsSevecsek, line 28: maybe remove this

src/hooks/airdrop/useAirdropTokensRefresh.ts
```diff
@@ -0,0 +1,42 @@
...
+// import { useInterval } from "../useInterval";
+import { useCallback, useEffect } from "react";
+import { invoke } from '@tauri-apps/api/tauri';
+
+export function useAirdropTokensRefresh() {
+    const { airdropTokens, setAirdropTokens } = useAirdropStore();
+
+    // Handle refreshing the access token
+    const handleRefresh = useCallback(() => {
+        const expired = airdropTokens?.expiresAt && new Date(airdropTokens?.expiresAt * 1000) > new Date();
```
BalazsSevecsek, line 11: Are you sure we need to multiply it by 1000?

PanchoBubble: yes we need to multiply by 1000 so its in ms and not seconds

src-tauri/src/telemetry_manager.rs
```diff
@@ -149,17 +170,48 @@ impl TelemetryManager {
...
+                                    None
+                                }
+                            };
+
+                            let now = std::time::SystemTime::now();
+                            let now_unix = now.duration_since(std::time::UNIX_EPOCH).unwrap_or_default().as_secs();
+
+                            if let Some(claims) = claims {
+                                if claims.exp < now_unix {
+                                    error!("Access token has expired");
```
BalazsSevecsek, line 200: I think we can use warn here as it is not technically an error, but something that we wait here.

src-tauri/src/telemetry_manager.rs
```diff
@@ -149,17 +170,48 @@ impl TelemetryManager {
...
+
+                        let airdrop_access_token_clone = airdrop_access_token.read().await.clone().and_then(|t| {
+                            let key = DecodingKey::from_secret(&[]);
+                            let mut validation = Validation::new(Algorithm::HS256);
+                            validation.insecure_disable_signature_validation();
+
+                            let claims = match decode::<AirdropAccessToken>(&t, &key, &validation) {
+                                Ok(data) => Some(data.claims),
+                                Err(e) => {
+                                    error!("Error decoding access token: {:?}", e);
```
BalazsSevecsek, line 190: use warn! here

src-tauri/src/telemetry_manager.rs
```diff
@@ -130,17 +145,23 @@ impl TelemetryManager {
         self.node_network = network;
     }
 
-    pub async fn initialize(&mut self) -> Result<()> {
+    pub async fn initialize(
+        &mut self,
+        app_state: tauri::State<'_, UniverseAppState>,
```
BalazsSevecsek, line 150: I think it is better if we don't give over the entire state, but only the access token (Arc<RwLock<>>) because that would make it more strict

src-tauri/src/main.rs
```diff
@@ -153,7 +165,12 @@ async fn setup_inner<'r>(
     let last_binaries_update_timestamp = state.config.read().await.last_binaries_update_timestamp;
     let now = SystemTime::now();
 
-    state.telemetry_manager.write().await.initialize().await?;
+    state
+        .telemetry_manager
+        .write()
+        .await
+        .initialize(state.clone())
```
BalazsSevecsek, line 172: here only give over the part of the state that is needed.

src/hooks/airdrop/useAirdropTokensRefresh.ts
BalazsSevecsek: okay

PR #201, opened 2024-08-28
chore: fix state flickering when mode change

Misieq01:
Description
---

Adjusted state transitions and changing mode behavior

Motivation and Context
---

Issues:
-  When not mining and changing mode it would start mining
-  Mining and changing mode would cause weird state flickering's 

How Has This Been Tested?
---

Manually

---

Play around with mining and changing mode

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src/store/useAppStatusStore.ts
+++ b/src/store/useAppStatusStore.ts
@@ -2,14 +2,12 @@ import { create } from 'zustand';
 import { ApplicationsVersions, AppStatus } from '../types/app-status.ts';
 import { modeType } from './types.ts';
 import { persist } from 'zustand/middleware';
-import { invoke } from '@tauri-apps/api/tauri';
 
 type State = Partial<AppStatus>;
 interface Actions {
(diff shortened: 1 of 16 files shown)


src/store/useAppStatusStore.ts
```diff
@@ -37,13 +37,15 @@ export const useAppStatusStore = create<AppStatusStoreState>()(
             setApplicationsVersions: (applications_versions) => set({ applications_versions }),
             setMainAppVersion: (main_app_version) => set({ main_app_version }),
             setMode: (mode) => set({ mode }),
-            setConfigMode: async (mode) => {
+            setConfigMode: async (mode, callback) => {
```
mmrrnn, line 40: What is that callback?

Misieq01: Leftover from first iteration

PR #206, opened 2024-08-28
chore(log): set log mining activity permission

mmrrnn:
Rust part for Erica's task

### DIFF
--- a/src-tauri/src/app_config.rs
+++ b/src-tauri/src/app_config.rs
@@ -20,7 +20,6 @@ pub struct AppConfigFromFile {
     pub last_binaries_update_timestamp: SystemTime,
     pub allow_analytics: bool,
     pub anon_id: String,
-    pub log_mining_activity_permission: bool,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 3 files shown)


src-tauri/src/app_config.rs
```diff
@@ -87,6 +90,7 @@ impl AppConfig {
                     self.allow_analytics = config.allow_analytics;
                     self.anon_id = config.anon_id;
                     self.version = config.version;
+                    self.log_mining_activity_permission = config.log_mining_activity_permission;
```
mmrrnn, line 93: it transforms to false, checked it

src-tauri/src/app_config.rs
```diff
@@ -20,6 +20,7 @@ pub struct AppConfigFromFile {
     pub last_binaries_update_timestamp: SystemTime,
     pub allow_analytics: bool,
     pub anon_id: String,
+    pub log_mining_activity_permission: bool,
```
stringhandler, line 23: there is already a config: "allow_analytics"

mmrrnn: Sorry, I thought it's for something else

PR #230, opened 2024-08-30
feat(settings): reset settings and all remove related data

mmrrnn:
It doesn't work in dev mode due to the hot reload. Try on built app

See attached video:
[Screencast from 30.08.2024 18:07:59.webm](https://github.com/user-attachments/assets/89b1d0b9-4156-4806-aa9e-2b23e83bee68)


src-tauri/src/main.rs
```diff
@@ -627,6 +628,43 @@ fn log_web_message(level: String, message: Vec<String>) {
...
+
+    dirs_to_remove.iter().for_each(|dir| {
+        // check if dir exists
+        if dir.clone().unwrap().exists() {
+            info!(target: LOG_TARGET, "[reset_settings] Removing {:?} directory", dir);
+            remove_dir_all(dir.clone().unwrap()).unwrap();
+        }
+    });
+
+    state.shutdown.clone().trigger();
```
stringhandler, line 658: This needs to happen before removing the dirs above. I suggest moving this to line 637

stringhandler (approved):

PR #237, opened 2024-09-02
feat: implement systray

Misieq01:
Description
---

Implement systray icon with informations about cpu, gpu and estimated earnings 

Motivation and Context
---

How Has This Been Tested?
---

Manually

---

Open app and check system tray

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/systemtray_manager.rs
+++ b/src-tauri/src/systemtray_manager.rs
@@ -1,9 +1,12 @@
 use std::sync::LazyLock;
+use log::{error, info};
 use tauri::{AppHandle, CustomMenuItem, SystemTray, SystemTrayMenu, SystemTrayMenuItem};
+
+use crate::hardware_monitor::HardwareStatus;
 
 
 const LOG_TARGET: &str = "tari::universe::notification_manager";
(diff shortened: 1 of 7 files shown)


Misieq01:
Windows: 
image

From what I understand there is limit of 64 characters for tooltip on windows so i had to make it shorter

Misieq01:
Linux: 
image

Tooltip is not supported there so instead it shows parameters on click

Misieq01:
macOS: 
image

src-tauri/src/systemtray_manager.rs
```diff
@@ -0,0 +1,234 @@
+use log::{error, info};
+use std::sync::LazyLock;
+use tauri::{AppHandle, CustomMenuItem, SystemTray, SystemTrayMenu, SystemTrayMenuItem};
+
+use crate::hardware_monitor::HardwareStatus;
+
+const LOG_TARGET: &str = "tari::universe::notification_manager";
```
stringhandler, line 7:
```suggestion
const LOG_TARGET: &str = "tari::universe::systemtray_manager";
```

PR #262, opened 2024-09-03
ci(fix): resolve osx universal builds, enable linux builds

leet4tari:
Description
Resolve OSX universal builds by installing all rust components for both arch from the src-tauri folder
Enabled linux builds

Motivation and Context
Fix OSX builds
Add Linux builds

How Has This Been Tested?
Builds in local fork


stringhandler (approved):
let's try it

PR #306, opened 2024-09-05
chore(update): custom auto update

mmrrnn:
Description
---
Currently auto update asks you to 1) Download and then 2) Restart. It was a request to merge it into one(first) step. 

Motivation and Context
---
Unfortunately, tauri updater does not allow to do so and it does not show that second step on windows. I tried to catch "DOWNLOADED" and "DONE" events, but it took some time to restart the app so restart dialog was displayed on linux/mac anyway. I decided to create our own dialog with auto update and restart logic.

How Has This Been Tested?
---
Manually

---
if you're using linux, it's quite challenging for now. You have to build your own (old)version, create updater json, host it somewhere and place it inside `tauri.conf.json`. Ask me for help if you need any! :)

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src/containers/AutoUpdateDialog/AutoUpdateDialog.tsx
+++ b/src/containers/AutoUpdateDialog/AutoUpdateDialog.tsx
@@ -46,7 +46,7 @@ function AutoUpdateDialog() {
         return () => {
             unlistenPromise?.then((unlisten) => unlisten());
         };
-    }, []);
+    }, [setIsAfterAutoUpdate]);
 
     const handleUpdate = async () => {
         setIsLoading(true);
--- a/src/containers/AutoUpdateDialog/AutoUpdateDialog.tsx
+++ b/src/containers/AutoUpdateDialog/AutoUpdateDialog.tsx
@@ -1,0 +1,89 @@
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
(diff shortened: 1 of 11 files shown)


mmrrnn:
[Screencast from 05.09.2024 10:59:25.webm](https://github.com/user-attachments/assets/17a6f89b-c0f0-4fed-985b-ac9b4c025015)

stringhandler (approved):

stringhandler (approved):
Happy to merge, but please sort out the Github React warning

src/containers/AutoUpdateDialog/AutoUpdateDialog.tsx
```diff
@@ -0,0 +1,89 @@
...
+            } catch (error) {
+                console.error(error);
+            }
+        };
+
+        checkUpdateTariUniverse();
+        return () => {
+            unlistenPromise?.then((unlisten) => unlisten());
+        };
+    }, []);
```
PanchoBubble, line 49: I would suggest to always keep the dependency array updated or add ` // eslint-disable-next-line react-hooks/exhaustive-deps` together with a reason why if any variable was not added on purpose.
I'll make a PR trying to make eslint strict for these cases if you agree
```suggestion
    }, [setIsAfterAutoUpdate]);
```

PR #310, opened 2024-09-05
fix(config): remove app config versioning

mmrrnn:
Description
---
Remove versioning and create `serde` default values to handle all properties at once.

Motivation and Context
---
I've noticed something strange that was happening for a while:
When we load older version of app_config it always throws an error and uses default config due to the json and AppConfigFromFile types inconsistency:
```
Failed to parse app config: missing field `p2pool_enabled` at line 1 column 46
App config does not exist or is corrupt. Creating new one
```

How Has This Been Tested?
---
unit tests + manually all cases

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src-tauri/src/app_config.rs
```diff
@@ -12,16 +12,24 @@ const LOG_TARGET: &str = "tari::universe::app_config";
...
-    pub allow_telemetry: bool,
-    pub anon_id: String,
-    pub monero_address: String,
-    pub gpu_mining_enabled: bool,
-    pub cpu_mining_enabled: bool,
+    #[serde(default = "default_mode")]
+    mode: String,
+    #[serde(default = "default_false")]
+    auto_mining: bool,
+    #[serde(default = "default_false")]
```
stringhandler, line 19:
```suggestion
    #[serde(default = "default_true")]
```

src-tauri/src/app_config.rs
```diff
@@ -12,16 +12,24 @@ const LOG_TARGET: &str = "tari::universe::app_config";
...
+    p2pool_enabled: bool,
+    #[serde(default = "default_system_time")]
```
stringhandler, line 21:
```suggestion
    #[serde(default = "default_0")]
```
This needs to be a long time in the past

src-tauri/src/app_config.rs
```diff
@@ -12,16 +12,24 @@ const LOG_TARGET: &str = "tari::universe::app_config";
...
+    last_binaries_update_timestamp: SystemTime,
+    #[serde(default = "default_true")]
```
stringhandler, line 23:
```suggestion
    #[serde(default = "default_false")]
```
This must be false unless they have accepted

src-tauri/src/app_config.rs
stringhandler: I think you can also just say `default = "true"`

mmrrnn: unfortunately no

stringhandler (approved):

stringhandler (commented):
Looks good, but the test fails on windows. I think we need to change the last update time to just store seconds from the system epoch and we convert it into time.

```
assertion `left == right` failed
  left: "SystemTime { intervals: 133700189673790786 }"
 right: "SystemTime { tv_sec: 1725545367, tv_nsec: 379078628 }"
 ```

PR #325, opened 2024-09-06
fix(ui): cleanup animation states

shanimal08:
Description
---

- updated initial animation tower rotation
- used framer `LayoutGroup` and `layout` prop to prevent janky re-renders on animations that affect the layout (majority of file changes)
- use `useShallow` when grabbing data from zustand stores which are updated every second in `useGetStatus` (majority of file changes)
- refactored the block height components:
  - split out the ruler and accent
  - added `BlockHeightAccent` component which handles the big number resizing in isolation
  - new ruler components with better centering just focusing on the marks and main block height number
- remove canvas hiding in telementary mode toggle implementation
- renamed `useBalanceInfo` to `useBalanceChanges` and updated logic a bit:
  - make sure to check if mining or not (if not mining just update the display block height immediately)
  - check for block_height and balance changes separately
- split out animation states setting (so we don't need to keep referencing the same hook, `useVisualisation`, in different places):
  - only use `useCallback` for `success` and `fail` animations since they depend on the widow being in view + fail animation has the timing deps
  - split success and fail so we can call either func more specifically
  - use `setAnimationState` directly for start, stop, and pause
- update zustand stores:
  - removed session storage for persist (so defaults to localStorage)
  - added `create` helper so we can reset to initial states on shut down

Motivation and Context
---

- there was a weird screen flicker when the block height changed (due to my hanky accent implementation)
- ruler and accent weren't always vertically centered
- display block height wasn't set if not mining
- 

How Has This Been Tested?
---

- locally:

https://github.com/user-attachments/assets/b9e7f6d1-3599-4d21-898c-6c723acee74a

https://github.com/user-attachments/assets/5f8c8f93-0e86-4144-8f8b-27341c1c3761


eslint.config.js
```diff
@@ -36,7 +36,10 @@ export default [
             ...hooksPlugin.configs.recommended.rules,
             'react/jsx-uses-react': 'off',
             'react/react-in-jsx-scope': 'off',
-            'no-console': ['warn', { allow: ['info', 'warn', 'debug', 'error'] }],
+            'no-console': [
+                'warn',
+                { allow: ['info', 'warn', 'debug', 'error', 'group', 'groupCollapsed', 'groupEnd'] },
```
stringhandler, line 41: seems wrong

PR #337, opened 2024-09-06
fix(telemetry): missing data

### LINKED ISSUE #335: Telemetry not logging correct app_id or mode and missing app_version
**Describe the bug**
This is definitely me in miner logs: SIAuDaF!nrhrkfiNlQ55

but it is not my app_id, which is app.config  `"anon_id":"e7TIu7JY1hg51%fTdP1z"`

mode is not being logged correctly--says eco when i switched to ludicrous

and we're missing app version, which is needed.

[tu_miner_debugging_2024-09-06T17_42_21.917786Z.csv](https://github.com/user-attachments/files/16911407/tu_miner_debugging_2024-09-06T17_42_21.917786Z.csv)

Krakaw:
Description
---
Fixes #335 

Motivation and Context
---
The telemetry data was missing the version, mode and app_id. This was because of an errant instance of the AppConfig being initialized then left unused except for the telemetry_manager.

How Has This Been Tested?
---
I `println`'d the telemetry data to check the correct values were being sent.

---
Run the app, opt into telemetry, view the telemetry data being sent matches what you see on screen.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


BalazsSevecsek (approved):

stringhandler:
Seems fine

stringhandler (approved):
I don't understand how it fixes this, but happy to merge anyway

Krakaw:
> I don't understand how it fixes this, but happy to merge anyway

The config initialization was being shadowed. So the first initialization was being cloned and moved into the tokio worker for the telemetry. Then the actual config that the main app was using and updating was being initialized and passed into the main app.

I just removed the useless first initialization and moved the main config init to above the telemetry init.

PR #372, opened 2024-09-09
feat(ui): block win sound

MCozhusheck:
Description
---
Add sound effect when block is mined.

Motivation and Context
---
More feedback to the user.

How Has This Been Tested?
---
Replaced condition to run audio and tested manually.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler:
Marking as controversial because of the permissions change

MCozhusheck:
> I want a block win sound. What exactly is the problem with permissions?

We need permissions for reading buffered file but I had problem with assign correct scope for just a single file that we needed and I ended up with `"scope": ["**"],` that works fine.

I tried using something like `"$RESOURCE/block_win.mp3"` and `assets/block_win.mp3` to no use.

MCozhusheck:
Managed to reduce the read access to just the audio file so now it's much less controversial.

Misieq01:
Tested on **Ubuntu** | **MacOS** | **Windows** 11 and works on each one

brianp:
I can't currently test this because every time I win I get lose animations.

Once we fix the animation timing, I'll come back to this.

brianp:
We need the block win sound to also play on the block win replay.

brianp:
Can you also add an event item that tracks when the sound is toggled?

MCozhusheck:
Added boolean in the `BlockchainVisualization` for tracking when the audio is playing `isPlayingAudio` and a setter `setIsPlayingAudio`.
Also we send telemetry event that looks like this:
```js
{
  "app_id": "hJr@t#FsvG30Mn2QLAS$",
  "cpu_name": "11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz",
  "created_at": {
    "nanos_since_epoch": 597927466,
    "secs_since_epoch": 1738765756
  },
  "event_name": "audio-enabled",
  "event_value": false,
  "gpu_name": "Intel(R) Iris(R) Xe Graphics [0x9a49]",
  "os": "Linux",
  "user_id": "v3,TFKEtEunDgrKrSkVYG54zU6x4tuJ,0.9.2,47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU",
  "version": "0.9.2"
}
```

PR #402, opened 2024-09-10
fix(settings): monero address editing & refactor

mmrrnn:
Description
---
* Removed `MoneroAddressInput` since it won't be reused so it's not a component.
* Added copy to clipboard logic
*  apply, cancel buttons instead of ssa
* Added monero address getter
* Used app config monero address when app starting.


mmrrnn:
[Screencast from 10.09.2024 13:45:39.webm](https://github.com/user-attachments/assets/0676bd65-d3e3-4ec2-961f-0cf9d2484735)

Krakaw (approved):
This works functionally, however when the error message is displayed it re-positions all of the other components on the page, causing them to shift (which has it's own issues, but aren't from this PR).

It looks good and does what it is meant to, if we could set the height of the component to include the error height that would be amazing.

mmrrnn:
Merged in #409

PR #409, opened 2024-09-10
chore(settings): refactor settings view

mmrrnn:
Description
---
* Split markups to separate components - they should rerender individually without affecting anything else
* Some performance improvements

Motivation and Context
---
Settings page is really slow, it refreshes multiple times per second. Some changes couldn't be done due to the refresh. Component is also too complex, bad managed what affect performance issues


stringhandler (approved):
tested and working, thanks!

Issue #381, opened 2024-09-09
0.2.8 Mac x86 GPU does not mine


metalaureate:
**Describe the bug**
After 20+ minutes GPU still doesn't have a hash rate.

![image](https://github.com/user-attachments/assets/d21305ba-73a1-43cc-9c09-a90e2047a3ed)
![image](https://github.com/user-attachments/assets/972a5440-9234-4b73-a340-a4e1386a5936)
![image](https://github.com/user-attachments/assets/659d72e6-5689-4af7-beb5-73329c20e890)

metalaureate:
[com.tari.universe.zip](https://github.com/user-attachments/files/16935589/com.tari.universe.zip)

stringhandler:
@karczuRF another instance. I suspect this is due to not having opencl installed maybe?

karczuRF:
Perhaps, I'm gonna investigate it

karczuRF:
@metalaureate Could you please check which version of OpenCl and OpenCL do you have? You can find it like this:
- Click on the Apple icon in the top-left corner of your screen and select "About This Mac."
- Click "System Report," then go to the "Graphics/Displays" section. 
- The OpenGL and OpenCL version will be listed under the graphics card details.

Issue #322, opened 2024-09-06
v0.2.5 no GPU with spinner in GPU panel


leet4tari:
Running a computer without a GPU, has the GPU panel showing a spinner, which would be confusing for a user.

**Expected behavior**
Thinking greying out the panel when there is no GPU, might be an idea

**Screenshots**
<img width="1621" alt="Screenshot 2024-09-06 at 12 13 48" src="https://github.com/user-attachments/assets/1fed9c1a-c1af-4f10-832f-89e5c858ce9a">

**Desktop (please complete the following information):**
 - Windows 11 - v0.2.5

stringhandler:
Similar to #358

PR #442, opened 2024-09-11
feat(gpu): detect gpu

### LINKED ISSUE #358: Add command to xtrgpuminer to test if it can run. If it returns an error, disable gpu mining
On TU startup, after getting/checking the latest version of xtrgpuminer, run a command to test if it can work on the system.

e.g.

`xtrgpuminer --detect`

This should try initialize the context, but not actually run any hashing. We are mostly trying to test for integrated GPU chips here.

If the resulting exit code is not 0, then we should disable GPU mining and show an error underneath the toggle saying "GPU disabled because your hardware does not support it"

karczuRF:
Description
While setting up the TU detect if gpu is available and can start mining. If any error and exit_status != 0 disable gpu mining

Motivation and Context
Resolves:
- #358
- #322 
- #381

How Has This Been Tested?
Ubuntu 22.04 manually 

Check with the newest gpu_miner build on Windows and macOS

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


karczuRF:
Just found small bug, will fix it now

PR #471, opened 2024-09-13
fix(ui): systray format

karczuRF:
Description
Fix system data formatting

Motivation and Context
Resolves #419 

How Has This Been Tested?
Manually

What process can a PR reviewer use to test or verify this change?
**MUST BE YET TESTED ON WINDOWS**

Breaking Changes
---

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

PR #483, opened 2024-09-16
fix(ui): format billion amount

### LINKED ISSUE #473: for money, the amount should be B for billion but for hashes, the amount should be G for gigahash

karczuRF:
Description
For money, the amount should be B for billion but for hashes, the amount should be G for gigahash

Motivation and Context
Resolves #473 

How Has This Been Tested?
Manually

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/utils/formatNumber.ts
```diff
@@ -1,10 +1,12 @@
...
     //TODO: add props for customisation
-    return Intl.NumberFormat(undefined, {
+    const formatted = Intl.NumberFormat(undefined, {
         notation: 'compact',
-        maximumFractionDigits: maxDigits || 3,
+        maximumFractionDigits: maxDigits,
         style: 'decimal',
     }).format(value);
+    // workaround because there is no "hashrate" unit
+    return isHashrate ? formatted.replace('B', 'G') : formatted;
```
stringhandler, line 9: This will only work for english locale. B is different in other locales

stringhandler (commented):
Thanks for the fix, but I don't think this will work in every locale

src/utils/formatNumber.ts
karczuRF: You're right. I'll find better solution to support locales and hashrate

brianp:
You introduced a backend human formatter in https://github.com/tari-project/universe/pull/471/files#diff-4f21708790727e4d29b64fe804fa59ee483bc9433176faa898f2944ab9830d11R1

Is this PR still relevant as is. And is this implementation still preferred?

karczuRF:
Because we use lower case for stats component units looks imo weird, and eg instead of `GH/s` we have `g h/s`. But this is out of scope for this PR and is more like UI thing
image

Issue #89, opened 2024-08-21
Block time readout doesn't start at t-0


metalaureate:
**Describe the bug**
Users are confused why block times start at non-zero time indexes such as 30 to 2 mins in.

**Expected behavior**
Normalize the block time start time on the UI.

stringhandler:
basically change block time from actual block time

to "last time I started mining or showed you an animation"

please put the current block time in the settings as "Last block added to chain time"

PR #493, opened 2024-09-16
chore(stores): split get status command & refactor stores

mmrrnn:
# Rust Commands Cleanup
* Split `status` to `get_p2pool_stats`, `get_tari_wallet_details`, `get_app_config`, `get_miner_metrics`
* Remove `get_monero_address, `get_telemetry_mode` since we've got one `get_app_config`
# Adjust Stores to fit new commands
* Split `useAppStatusStore` to `useMiningStore`, `useAppConfigStore`, `useP2poolStatsStore`
* Move `useBaseNodesStatusStore`, `useCpuStatusStore` and `useGpuStatusStore` to `useMiningStore` since all fetches/updates are made all together and logic is mainly shared
# General improvements
* ***Moved `useAirdropSyncState` to `AirdropLogin, `useMining` and `useUiMiningStateMachine` to the `Miner` since it every dependency update(quite frequent) leads to rerendering whole App tree(along with all Providers and canvas - it's really expensive)***
* Fix memory leak with setInterval in `useSetUp` hook
* Separated `SettingsDialog` and moved it to the top of the tree so it's not rerender 6 times/seconds leading to issues with scrolling and changing values
* Move some components logic to Zustand actions - it minimizes components complexity and unwanted rerenders. Stores' logic should be placed in store actions. It also improves actions ordering(debugging) and sharing logic.
* Removed `useShallow` abuse and modified state getters to be more efficient
* Memoized stuff inside components
* We fetch stuff individually now and have more control over react tree
* Removed `useMiningControls` and `useVersions`, `useApplicationsVersions` - some logic moved to stores, some is not needed anymore
* Improved error handling inside stores
* FIxed Cpu/Gpu Tile statuses in `Miner`
* Fixed GPU/CPU markups disabled flag in `Settings`
* Fixed calculating time from previous block


src/containers/SideBar/components/Settings/DebugSettings.tsx
```diff
@@ -1,14 +1,13 @@
...
-import { useAppStatusStore } from '@app/store/useAppStatusStore.ts';
 import { Stack } from '@app/components/elements/Stack.tsx';
-import { useShallow } from 'zustand/react/shallow';
+import { useMiningStore } from '@app/store/useMiningStore';
 
 export default function DebugSettings() {
     const { t } = useTranslation(['common', 'settings'], { useSuspense: false });
 
-    const blockTime = useAppStatusStore(useShallow((state) => state.base_node?.block_time));
+    const blockTime = useMiningStore((state) => state.base_node?.block_time);
```
shanimal08, line 10: not sure if there's it's an issue with the store/setting the value, but the settings counter doesn't updaye

https://github.com/user-attachments/assets/47284f3f-8525-4575-8bfb-3f88c345ce65

src/containers/SideBar/components/Settings/Markups/CpuMiningMarkup.tsx
```diff
@@ -1,47 +1,32 @@
...
 import { MinerContainer } from '../../../Miner/styles';
 import { useTranslation } from 'react-i18next';
+import { useAppConfigStore } from '@app/store/useAppConfigStore';
 
 const CpuMiningSettings = () => {
     const { t } = useTranslation(['common', 'settings'], { useSuspense: false });
+    const isCPUMining = useMiningStore((s) => s.cpu.mining.is_mining);
+    const isGPUMining = useMiningStore((s) => s.gpu.mining.is_mining);
+    const isCpuMiningEnabled = useAppConfigStore((s) => s.cpu_mining_enabled);
+    const setCpuMiningEnabled = useAppConfigStore((s) => s.setCpuMiningEnabled);
```
shanimal08, line 15: might want to also set `miningControlsEnabled` when updating gpu/cpu_mining_enabled in `useAppConfigStore` - i turned both off in settings and got stuck in 'start mining' witht he button disabled but obviously mining will never start 😅

src/store/useMiningStore.ts
```diff
@@ -35,17 +44,106 @@ const initialState: State = {
...
+            const metrics = await invoke('get_miner_metrics');
+            set(metrics);
+        } catch (e) {
+            console.error(e);
+        }
+    },
+    handleBlockMined: () => {
+        set({
+            postBlockAnimation: true,
+            timerPaused: true,
```
shanimal08, line 85:
```suggestion
            timerPaused: false,
```

was testing and after winning everything would freeze up 😅 button and mode switch disabled, block height wouldn't change, got sooo worried, but it's thankfully just this little value - timer needs to be unpaused hehe

src/store/useMiningStore.ts
```diff
@@ -35,17 +44,106 @@ const initialState: State = {
...
 };
 
-export const useMiningStore = create<MiningStoreState>()((set) => ({
+export const useMiningStore = create<MiningStoreState>()((set, getState) => ({
     ...initialState,
-    setDisplayBlockTime: (displayBlockTime) => set({ displayBlockTime }),
-    setDisplayBlockHeight: (displayBlockHeight) => set({ displayBlockHeight }),
-    setEarnings: (earnings) => set({ earnings }),
+    fetchMiningMetrics: async () => {
+        console.info('Fetching mining metrics...');
```
shanimal08, line 74: do we need to log this every second? 😅 

also, related to this in `useMiningMetricsUpdater` - i noticed my cpu mining almost never works, so if i disabled gpu mining and hit 'start' the `fetchMiningMetrics` just goes on foreeeverr and my button is stuck on loading. 
maybe it'd help to check here which mining mode is enabled and if the metrics response ever includes 'is_mining' then stop the interval and reset mining state after like 100 tries? just an idea (i'm not sure what the best approach for that might be maybe @stringhandler has some ideas:D?)

shanimal08 (commented):
just a couple comments, most important being unpausing the timer after winning a block, but other than those looks good! 

i'm not a huge fan of making calls inside the store set actions, but i think it makes sense when using invoke that just updates the same stuff over and over, or once :D 

i tested and do think it feels a lot better :D i still see a random flicker/re-render around that sidebar every now and then, but a lot less frequently (i think it has something to do with the hardware stats or GPU hashrate but don't know what:( )

src/store/useMiningStore.ts
mmrrnn: Was this happening before for you?

src/containers/SideBar/components/Settings/Markups/CpuMiningMarkup.tsx
mmrrnn: Disabled mining button when both cpu and gpu mining are disabled in settings

src/containers/SideBar/components/Settings/DebugSettings.tsx
mmrrnn: Fixed

src/store/useMiningStore.ts
mmrrnn: Played with this a bit. Refactored calculating `displayBlockTime`. But why do we need to pause timer in a first place? I calculate time: now - last_block_time - it seems like we don't need this anymore

src/containers/SideBar/components/Settings/DebugSettings.tsx
mmrrnn: Hey, it doesn't work on current main.

src/store/useMiningStore.ts
shanimal08: yeah the cpu_mining has always been spotty for me locally, it either takes ages to start up or never does. so that wasn't introduced here, just think it may be a good time to add a fallback/fail on too many retries?

src/store/useMiningStore.ts
shanimal08: we had to add a delay to wait for balance changes because the balance only comes in a bit after the block height changes, without the delay we might only display a win after it moves on to the next block. timer is paused just to allow all the animations to run through before updating the display block height and block time

mmrrnn: Was it like requirement to reset timer, update balances and display animation in exact the same time? If yes i still think it's better way to handle it: check whether block has changed in store

shanimal08: yeah, it was 😅 - https://github.com/tari-project/universe/issues/89 - but that's from a while back, maybe we can change things now? @stringhandler pls advise :D

mmrrnn: I see, so I can also handle it using current logic, should be better with current flow

stringhandler:
Tested and working well. Needs conflict resolved, but can be merged

PR #500, opened 2024-09-17
refactor: store the passphrase without keyring

brianp:
Description
---
In the event the user doesn't give access to the system keychain, then simply write the passphrase to the config file. It is less secure, but the user opted out of security when denying access to the keyring. Ideally, we would let them know at some point but this happens to early in the initialization to prompt them with anything.

Motivation and Context
---
Recover, and don't allow a middle ground where someone can create a wallet and lose access to their seedphrase.

How Has This Been Tested?
---
Manually

closes: #467


stringhandler:
We need to surface this error to the users in the front end and give them a chance to try again

PR #510, opened 2024-09-18
feat: show system notifications

Misieq01:
##  [ New Feature ] => Notifications
Implemented reusable method for triggering display of system notifications with desired title and body message

#### Linux:
- Working on both dev and production environments
- Missing Tari icon for notification

### Windows
- Working on both dev and production environments
- Notifications displayed as if they have been send from powershell

### MacOS
- Working only on production environment
- To make it work user have to enable notifications for Tari app and go to notifications center and switch notficiation display from banner to alert for Tari

--------

### [ @tauri-apps/plugin-notification ]
Currently implemented and replaced old implementation with official tauri plugin
After testing it looks like it work similar as my implementation or even slightly worse

#### Linux:
- Working only on production environment ( build locally )
- Missing Tari icon for notification

### Windows
- Working on both dev and production environments
- Notifications `ON DEV` displayed as if they have been send from powershell
- Notifications `ON BUILD` displayed with proper tari icon

### MacOS
- Working only on production environment
- To make it work user have to enable notifications for Tari app and go to notifications center and switch notficiation display from banner to alert for Tari


src-tauri/src/notification_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+use log::info;
+
+const LOG_TARGET: &str = "tari::universe::notification_manager";
+static INSTANCE: LazyLock<NotificationManager> = LazyLock::new(NotificationManager::new);
+
+pub enum CurrentOperatingSystem {
+    Windows,
+    Linux,
+    MacOS,
+}
```
brianp, line 13: This enum exists in the `SystemtrayManager` can we define it in a single place and reuse it?

src-tauri/src/notification_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+        if cfg!(target_os = "windows") {
+            CurrentOperatingSystem::Windows
+        } else if cfg!(target_os = "linux") {
+            CurrentOperatingSystem::Linux
+        } else if cfg!(target_os = "macos") {
+            CurrentOperatingSystem::MacOS
+        } else {
+            panic!("Unsupported OS");
+        }
+    }
```
brianp, line 65: Maybe the repeated OS level stuff could get moved into a trait as default implementations?

src-tauri/src/notification_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        match Self::detect_current_os() {
+            CurrentOperatingSystem::Linux => {
+                notification.auto_icon().appname("Tari Universe").finalize()
+            }
+            CurrentOperatingSystem::MacOS => notification.finalize(),
+            CurrentOperatingSystem::Windows => notification.finalize(),
+        }
+    }
+    fn detect_current_os() -> CurrentOperatingSystem {
```
brianp, line 55:
```suggestion
    }
    
    fn detect_current_os() -> CurrentOperatingSystem {
```

brianp (changes requested):
Code looks good. It's at least the second time we're using this pattern so it could be good to move the repeated bits to a trait if possible.

Once the branch is updated I'll test.

brianp (approved):
👏🏻  Excellent stuff

src-tauri/src/notification_manager.rs
```diff
@@ -0,0 +1,55 @@
...
+            CurrentOperatingSystem::MacOS => {
+                #[cfg(target_os = "macos")]
+                notification.show().unwrap();
+            }
+            CurrentOperatingSystem::Windows => {
+                #[cfg(target_os = "windows")]
+                notification.show().unwrap();
+            }
+        }
+    }
```
brianp, line 38: Three unwraps here, within a function that doesn't return a result could do a bit of unexpected damage. If we could clean that up, it would be looking quite brilliant.

src-tauri/src/notification_manager.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        match PlatformUtils::detect_current_os() {
+            CurrentOperatingSystem::Linux => {
+                #[cfg(target_os = "linux")]
+                let handle = notification.show().inspect_err(
+                    |e| warn!(target: LOG_TARGET, "Failed to show notification: {:?}", e),
+                )?;
+                handle.on_close(|notification| {
+                    info!(target: LOG_TARGET, "Notification closed: {:?}", notification);
+                });
```
brianp, line 30: Line 28 is outside of the `cfg` block

```suggestion
                #[cfg(target_os = "linux")]
                {
                    let handle = notification.show().inspect_err(
                        |e| warn!(target: LOG_TARGET, "Failed to show notification: {:?}", e),
                    )?;
                    handle.on_close(|notification| {
                        info!(target: LOG_TARGET, "Notification closed: {:?}", notification);
                    });
                }
```

brianp (changes requested):
I tried running in both dev, and with a local build. MacOS doesn't seem to know what to do with the notifications. And prompts with a confusing message Saying to Choose an Application and "Where is use_default".

image

Misieq01:
> I tried running in both dev, and with a local build. MacOS doesn't seem to know what to do with the notifications. And prompts with a confusing message Saying to Choose an Application and "Where is use_default".
> 
> image

This feature does not work on macos in dev mode. Did it happen also in local build version ?

brianp:
Sorry I didn't re-read the notes after the weekend and forgot it was prod mode only.

We should disable this in the dev environment. It's quite a large nuisance if it can't be used. 

It worked fine with a prod build. Although we need to format the win value correctly:
image

Even in the prod build. If notifications aren't enabled yet we receive the dialog to Choose Application.

Misieq01:
> Sorry I didn't re-read the notes after the weekend and forgot it was prod mode only.
> 
> We should disable this in the dev environment. It's quite a large nuisance if it can't be used, as it pops up those dialogs on every block win.
> 
> When notifications aren't enabled yet we still receive the dialog to Choose Application even on the prod build. Once enabled we need to format the win value correctly: image

I should be able to disable triggering of showing notifications on dev  environment and also format the amounts, but as for the dialog on macos for choosing application I'm not sure if I will be able to do something with it

brianp:
> I should be able to disable triggering of showing notifications on dev environment and also format the amounts, but as for the dialog on macos for choosing application I'm not sure if I will be able to do something with it

It seems strange. It's a very terrible UX experience. But this isn't normal. Shouldn't we be able to prompt for notifications permission or something similar instead of trying to use notifications, and showing the Choose Application dialog. I feel like there's a step missing somewhere.

brianp:
It may be useful to use Tauri's existing plugin for system notifications: https://github.com/tauri-apps/tauri-plugin-notification

It should allow us to perform actions like ensuring we have permission before sending. It also appears to handle the OS dependent sending switches.

brianp:
on hold until the tauri v2 upgrade is done, as we suspect the API will become easier to deal with.

brianp:
Moving this into the Todo list. Once V2 is merged let's revisit the tauri notifications plugin, and see if we can get rid of the weird "Choose application" dialog in mac os.

Misieq01:
@brianp Could you decide what do we want to do with it ? Described in description how it behaves on each platforms. If you are okay with how it works then I can go and finish implementing it

PR #529, opened 2024-09-19
feat(ui): airdrop add gems animation to gems stats pill

peps:
This adds a "gem animation" when the user's gems value changes. This is for user's who are logged into Airdrop. 

https://www.loom.com/share/90cda46674c4434fbecc43680aaa7133?sid=9788e9e2-883d-4b1b-a3a8-dcb4fceec844

I also cleaned up the design for the stats pill and user menu. The previous contrast was difficult to read with the background. 

CleanShot 2024-09-19 at 17 14 23

Note: Right now this only triggers when the Gem value changes. It does play the animation on initial load, but it's usually covered by the splash screen so it is not visible. To fix this, we'll need to disable the UserInfo component until the splash screen is hidden.


PanchoBubble:
Merged your changes to my branch so we can release them together

PR #555, opened 2024-09-20
fix(gpu): prevent app crash if gpu detect fails

MCozhusheck:
Description
---
On my integrated intel gpu the gpu miner would fail to detect the gpu causing crash of the entire application. This PR only makes it log error and continue working as gpu mining will be disabled.

Motivation and Context
---
Stop crashing app with unsupported GPUs.

How Has This Been Tested?
---
Tested manually on my machine (Ubuntu with integrated intel gpu)

---
same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -358,7 +358,15 @@ async fn setup_inner(
             .ensure_latest(Binaries::GpuMiner, progress.clone())
             .await;
 
-        state.gpu_miner.write().await.detect().await?;
+        drop(
+            state
+                .gpu_miner
+                .write()
+                .await
+                .detect()
+                .await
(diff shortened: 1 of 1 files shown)


src-tauri/src/main.rs
```diff
@@ -358,7 +358,15 @@ async fn setup_inner(
             .ensure_latest(Binaries::GpuMiner, progress.clone())
             .await;
 
-        state.gpu_miner.write().await.detect().await?;
+        drop(
```
Misieq01, line 361: Could you move it outside if statement so its called every time app starts ?

mmrrnn (approved):

stringhandler (approved):

PR #601, opened 2024-09-25
fix(ui): updated airdrop claim modal UI 2

peps:
image


peps:
closing since its out of sync with the base branch

Issue #299, opened 2024-09-04
0.2.2 System Tray Icon on Mac is not compliant


metalaureate:
**Describe the bug**
![image](https://github.com/user-attachments/assets/3fa861f8-4cab-43f7-9b09-014365e8f48e)

Design to provide.

metalaureate:
Designs: https://www.figma.com/design/Tyh0UcWVo44z66gYHd0TBs/Tari-Universe-Planning-%26-Wires?node-id=2358-10281&node-type=FRAME&t=Ac8ACclav4xKKMeM-0

<img width="871" alt="Screenshot 2024-09-04 at 4 26 24 PM" src="https://github.com/user-attachments/assets/2101f9d3-78e4-496e-916c-1259d9e9571c">

PR #609, opened 2024-09-25
fix(icons): update system tray icons

MCozhusheck:
Description
---
fixes: #299 
Update system icons according to designs.

Motivation and Context
---
Fix incorrect icons.

How Has This Been Tested?
---
Just launch app and check icons in system menu.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler:
I think you are changing all the icons here, when only the systray icon needs to be replaced

MCozhusheck:
> I think you are changing all the icons here, when only the systray icon needs to be replaced

Reverted change for apply new system tray icon in all cases. Now only it's applied to system tray depending to OS of the user according to [tauri docs](https://tauri.app/v1/guides/features/icons/)

stringhandler:
I don't think these are the correct icons.

MCozhusheck:
> I don't think these are the correct icons.

Yep I mixed up icons from designs. Now should be correct.

brianp:
I can not build this PR. @MCozhusheck will investigate.

image

PR #624, opened 2024-09-26
fix: cargo lint error

shanimal08:
Motivation and Context
---

there's a cargo-lints error causing all the new PR clippy checks to fail, just added another unused var per the `help` suggestion, and then dropped it immediately after the new help suggestion 😬 

image


shanimal08:
closing cos it's already been fixed

PR #625, opened 2024-09-26
fix: remove all unwraps in hardware monitors

MCozhusheck:
Description
---
There are instances of app crash happening caused by `unwrap` in `hardware_monitor.rs`. This PR adds default values for those unwraps.

Motivation and Context
---
Prevent app crash if some data about cpu or gpu is unavailable.

How Has This Been Tested?
---
Just run tari universe.

---
same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (approved):
I'm going to merge it, but in future, please add `target: LOG_TARGET` to `warn!` so that we can redirect it to the correct log file

PR #630, opened 2024-09-26
fix(ui): systray earnings and minimize

mmrrnn:
* Only CPU earnings were displayed in the systray(not CPU + GPU),
* Systray's "unminimize" button was enabled even though the app was unminimized -> On mac and linux


brianp:
The issue needs a description. I do not know what it is trying to achieve.

brianp (changes requested):
The issue needs a description. I do not know what it is trying to achieve.

brianp (approved):
Tested. Unminimize button is not active when unminized, but becomes available when minimized.

PR #663, opened 2024-09-30
fix: allow beta builds on same computer as main builds

stringhandler:
Stores data in `<data>/com.tari.universe.beta` instead of `com.tari.universe` for development and beta builds, allowing developers to run both without confusion. 

Seed words are also stored in a different keychain location for beta and release.

Note: requires release builds to be built with `feature = "release-ci"`


leet4tari (approved):
utACK

Issue #502, opened 2024-09-17
User can right-click and select a "Reload" option from the main Universe window which then restarts the interface


SolfataraEmit:
**Description**
When running Tari Universe, the user can right-click on the interface which then presents a "Reload" option. This appears to reload the interface, starting the initial boot up sequence that a user encounters when first running Tari Universe.

**To Reproduce**
1. Have Tari Universe Installed.
2. Run Universe, then right-click anywhere within the interface. You should have a "Reload" option
3. Select reload

**Expected behavior**
The expected behaviour should be that the user is not able to reload the interface in this manner, based on feedback.

**Desktop (please complete the following information):**
 - OS & Version: macOS 14.6.1

SolfataraEmit:
This issue appears to have been resolved as of version 0.3.7. Closing for now.

SolfataraEmit:
Noted that the issue appears to have returned in 0.4.6. Reopening issue.

PR #665, opened 2024-09-30
feat: disable refresh on non dev build

MCozhusheck:
Description
---
fixes: #502 
Disabled whole context menu (menu on right click) in non dev build.

Motivation and Context
---
Disable refresh action for normal user which in current state introduces bugs to the application.

How Has This Been Tested?
---
Launch app in dev and release mode and right click to check if context menu appears.

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/App.tsx
```diff
@@ -18,8 +18,25 @@ import CriticalErrorDialog from './containers/CriticalErrorDialog/CriticalErrorD
...
+    });
+
+    document.addEventListener('contextmenu', function (event) {
+        event.preventDefault();
+    });
+};
+
 export default function App() {
     useLangaugeResolver();
+    disableRefresh();
```
mmrrnn, line 39: Use hook which returns fn cleaning at every rerender(removes this event listener)

MCozhusheck: done

brianp (approved):
Tested and on prod builds the menu is disabled.

Tas4tari:
Tested on Windows 11, app version 0.5.60 and when I right click within the release build, I don't see a menu

Issue #633, opened 2024-09-26
Wallet shows 0 when it is undefined--it should show "-"


metalaureate:
**Describe the bug**
Users report their balances fluctuating between 0 and its value. Users are disconcerted to see wallet balances at zero when app starts.

stringhandler:
So on startup, it should show "-" and once started, if it is 0, then show 0?

PR #667, opened 2024-09-30
fix: wallet balance display if no tXTM

shanimal08:
Description
---
- set balance initial state to `null` and check specifically for `null` instead of `0` to display the `-`

Motivation and Context
---
- misunderstood the requirement for #633 and set it to always be `-` if 0

NOTE: 

- this undoes the initial fix for #633 since the BE returns 0 while it's loading anyway so the `-` hardly has time to show 😅

### DIFF
--- a/src/containers/SideBar/components/Wallet.tsx
+++ b/src/containers/SideBar/components/Wallet.tsx
@@ -12,12 +12,12 @@ import { useWalletStore } from '@app/store/useWalletStore.ts';
 function Wallet() {
     const { t } = useTranslation('sidebar', { useSuspense: false });
     const balance = useWalletStore((state) => state.balance);
-    const formatted = formatBalance(balance);
+    const formatted = formatBalance(balance || 0);
     const sizing = formatted.length <= 6 ? 50 : formatted.length <= 8 ? 44 : 32;
     const [showBalance, setShowBalance] = useState(true);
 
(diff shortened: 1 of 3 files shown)


src/containers/SideBar/components/Wallet.tsx
```diff
@@ -12,12 +12,12 @@ import { useWalletStore } from '@app/store/useWalletStore.ts';
...
     const { t } = useTranslation('sidebar', { useSuspense: false });
     const balance = useWalletStore((state) => state.balance);
-    const formatted = formatBalance(balance);
+    const formatted = formatBalance(balance || 0);
     const sizing = formatted.length <= 6 ? 50 : formatted.length <= 8 ? 44 : 32;
     const [showBalance, setShowBalance] = useState(true);
 
     const toggleBalanceVisibility = () => setShowBalance((prev) => !prev);
-    const displayValue = balance && balance > 0 ? (showBalance ? formatted : '*****') : '-';
+    const displayValue = balance === null ? '-' : showBalance ? formatted : '*****';
```
mmrrnn, line 20:
```suggestion
    const displayValue = balance ?? (showBalance ? formatted : '*****');
```

mmrrnn (approved):

stringhandler:
Requires a backend change to work fully

PR #673, opened 2024-09-30
chore(log): logging improvements

mmrrnn:
* Unify dirs structure - every binary(except xmrig) has config and log dir. All are nested at the same level.
Screenshot from 2024-09-30 23-48-51
* limit logs size to 2mb, use level warn
* Create log4rs config on every run and save it to the configs dir for each binary.


PikoroNamekiano (approved):

src-tauri/src/main.rs
```diff
@@ -1663,19 +1665,23 @@ fn main() {
...
                 &app.path_resolver()
                     .app_log_dir()
                     .expect("Could not get log dir"),
-                include_str!("../log4rs_sample.yml"),
+                include_str!("../log4rs/universe_sample.yml"),
             )
             .expect("Could not set up logging");
+            let config: RawConfig = serde_yaml::from_str(&contents)
+                .expect("Could not parse the contents of the log file as yaml");
+            log4rs::init_raw_config(config).expect("Could not initialize logging");
```
brianp, line 1684: What exactly is happening differently that `tari_common::initialize_logging` wasn't doing that required duplicating this functionality?

brianp (changes requested):
I might be overlooking the benefit of duplicating this module. What exactly is being done differently? And is it something `tari_utilities` could support, so we can keep the code in one place instead of copying it over?

mmrrnn:
Check out current logs structure. All binaries` logs are placed randomly, without specific structure

brianp:
I understand what the goal of the PR is. But the original code wasn't simply dropping files in random places. It was configurable. So why did we have to duplicate the function. What changed in the function fundamentally that didn't work before? 

If you had corrected the paths as this pr does, could we continue to use the original `tari_common::initialize_logging` ?

mmrrnn:
default tari logs path hardcode `/log` nesting

PR #612, opened 2024-09-25
feat: gpu devices list and management

karczuRF:
Description
Resolves
-  #611 
-  #497 

Gpu turning on/off devices works only [if the associated tarigpuminer PR](https://github.com/stringhandler/tarigpuminer/pull/18) is merged. 

Motivation and Context
Refactor hardware parameters to return not only the single (first) one GPU device, but all detected ones, allowing for better gpu devices management (e.g. turning on/off specified) and controlling params like hashrate.

How Has This Been Tested?
Manually on Linux and macOS

What process can a PR reviewer use to test or verify this change?
Check manually

Breaking Changes
---

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

PR #675, opened 2024-10-01
fix: gpu devices list

### LINKED ISSUE #610: GPU usage for MacOS
The hardware monitor needs to be refactored to return correct gpu parameters. At the moment [it returns cpu usage and cpu label ](https://github.com/tari-project/universe/blob/ab6e3b794f33e4d0cee6c2e465d4e9dd7a067505/src-tauri/src/hardware_monitor.rs#L445)

Image

E.g. on MacBook is should get two gpu devices (sth like `Intel UHD Graphics 630` and `Radeon Pro 555X`) but instead I'm getting cpu brand name (so `Intel Core i7... GPU`) which is not correct. 
Also usage seems to be taken from `global_cpu_usage()` but this needs to be verified if it's correct

karczuRF:
Description
Resolves #610 

Motivation and Context
Gpu devices are not listed on macOS so the info needs to be taken from gpu_status.json file which is created while gpu device detection by gpuminer.
This solution was chosen to use `opencl3` crate which supports different os, includes macOS.

Solution will work only if https://github.com/stringhandler/tarigpuminer/pull/19 is merged.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/gpu_miner_adapter.rs
+++ b/src-tauri/src/gpu_miner_adapter.rs
@@ -35,6 +35,14 @@ pub(crate) struct GpuMinerAdapter {
     pub(crate) coinbase_extra: String,
 }
 
+#[derive(serde::Deserialize)]
+pub struct GpuStatusFromFile {
+    #[serde(default = "default_num_devices")]
+    num_devices: u32,
+    #[serde(default = "default_device_names")]
+    device_names: Vec<String>,
+}
+
 impl GpuMinerAdapter {
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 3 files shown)


src-tauri/src/gpu_miner_adapter.rs
```diff
@@ -218,6 +229,27 @@ impl StatusMonitor for GpuMinerStatusMonitor {
 
     async fn status(&self) -> Result<Self::Status, anyhow::Error> {
         let client = reqwest::Client::new();
+        let config_path = PathBuf::from(""); //TODO add app config_path
```
stringhandler, line 232: if you move the config path in the gpuminer to after the config file, you will not have this problem

stringhandler (commented):
Should I not merge #612 first?

src-tauri/src/gpu_miner_adapter.rs
karczuRF: Thanks! I'll refactor this after merge this and 612 - I'll have to add some code to integrate these two PRs anyway

brianp (approved):

Tas4tari:
Tested on versions: 0.5.20 & Release version 0.5.22:
Ticket logged:
https://github.com/tari-project/universe/issues/843

Issue #662, opened 2024-09-30
Minotari Node continues in background if you close during initial sync


stringhandler:
v0.4.6

To reproduce:
* close Tari Universe.
* Start Tari Universe
* Close during setup screen when it is connecting (0/3 nodes)
![image](https://github.com/user-attachments/assets/3e5787b9-14fd-4f81-a34d-caa1986a009a)

Expected: 
* App closes and all binaries are stopped

Actual:
* minotari node continues
* This can be seen in logs and in Task Manager on windows

PR #684, opened 2024-10-01
fix: graceful exit

MCozhusheck:
Description
---
fixes: #662 
Exiting app during initialization (especially during connecting to other peers) caused app to run some processes in the background.

Motivation and Context
---
Fix issue describe above.

How Has This Been Tested?
---
Run TU and exit during initialization while looking for other peers in the network.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


MCozhusheck:
> Tested locally but this does not solve the problem

Found and fixed and another commit issue with p2pool locking RWLock which caused dangling p2pool process. It was due p2pool trying to get status from status monitor which was already killed.

src-tauri/src/process_watcher.rs
```diff
@@ -68,7 +68,9 @@ impl<TAdapter: ProcessAdapter> ProcessWatcher<TAdapter> {
             loop {
                 select! {
                       _ = watch_timer.tick() => {
+                            info!(target: LOG_TARGET, "Checking if {} is running", name);
```
stringhandler, line 71:
```suggestion
                            debug!(target: LOG_TARGET, "Checking if {} is running", name);
```

src-tauri/src/process_watcher.rs
```diff
@@ -68,7 +68,9 @@ impl<TAdapter: ProcessAdapter> ProcessWatcher<TAdapter> {
...
                             if child.ping() {
+                                info!(target: LOG_TARGET, "{} is running", name);
```
stringhandler, line 73:
```suggestion
                                debug!(target: LOG_TARGET, "{} is running", name);
```

src-tauri/src/p2pool_manager.rs
```diff
@@ -147,11 +147,12 @@ impl P2poolManager {
             .await?;
         process_watcher.wait_ready().await?;
         if let Some(status_monitor) = &process_watcher.status_monitor {
-            loop {
+            for attempt in 1..=5 {
                 sleep(Duration::from_secs(5)).await;
```
stringhandler, line 151: 5 seconds is very long here, rather make it 1 second but run more attempts

brianp:
Not working for me. The base node is still being left orphaned after I've shut down the app.

image

stringhandler:
Think this is fixed in 0.5.20

PR #692, opened 2024-10-01
chore(ci): add local builds as alpha, ci for beta and release

leet4tari:
Description
Add local defulat builds as alpha, ci for beta and release

Motivation and Context
Give users that fork the repo and build local an build labeled alpha
Ci none release builds as beta ... found that we can use [], nor ## nor dashes

How Has This Been Tested?
Builds locally and in local fork with CI and uploads assets as expected for all OS (Windows/OSX/Linux)

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -85,8 +85,10 @@ const LOG_TARGET_WEB: &str = "tari::universe::web";
 
 #[cfg(feature = "release-ci")]
 const APPLICATION_FOLDER_ID: &str = "com.tari.universe";
-#[cfg(not(feature = "release-ci"))]
+#[cfg(feature = "release-ci-beta")]
 const APPLICATION_FOLDER_ID: &str = "com.tari.universe.beta";
+#[cfg(not(any(feature = "release-ci", feature = "release-ci-beta")))]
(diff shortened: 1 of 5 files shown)


stringhandler (approved):

src-tauri/src/main.rs
```diff
@@ -85,8 +85,10 @@ const LOG_TARGET_WEB: &str = "tari::universe::web";
 
 #[cfg(feature = "release-ci")]
 const APPLICATION_FOLDER_ID: &str = "com.tari.universe";
-#[cfg(not(feature = "release-ci"))]
+#[cfg(feature = "release-ci-beta")]
```
stringhandler, line 88:
```suggestion
#[cfg(and(feature = "release-ci-beta", not(feature="release-ci"))]
```
This might fix the clippy

stringhandler (approved):

src-tauri/src/main.rs
leet4tari: I made a change, but now I think we need to put in a faulty state, if both features ```release-ci``` and ```release-ci-beta```, will give us a consts ```note: `APPLICATION_FOLDER_ID` must be defined only once in the value namespace of this module```?

PR #693, opened 2024-10-01
fix:  overflowing scrolling numbers on linux

Krakaw:
Description
---

Scrolling numbers in the wallet were not hiding

Motivation and Context
---

Before:
image
After:
image

How Has This Been Tested?
---

I ran it on linux.

---
Run it on the platform make sure the numbers aren't displayed except for the actual wallet value.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


shanimal08 (approved):
not exactly sure how only this z-index change fixed it 😅 but can't test on Linux, so tested these changes on my mac and nothing's borked so 👍🏼 :D

PR #705, opened 2024-10-02
fix(ui): update block timer when no mining

mmrrnn:


brianp:
Sounds like this PR is already fixed in https://github.com/tari-project/universe/pull/807

> updated the debug settings block time counter to use block_time from base_node instead of the display block height (also updates when not mining)

Closing in favour of the already merged #807.

If you feel like this closing is in error. Please update the PR, re-open, and let me know why.

PR #710, opened 2024-10-03
fix: add default values for p2pool during initialization

MCozhusheck:
Description
---
Add default values for p2pool stats.

Motivation and Context
---
Fix `undefined` displayed in experimental setting during initialization.

How Has This Been Tested?
---
Run app and check if there is any `undefined` in experimental settings.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


Tas4tari:
Checked on Beta version 0.5.10 & there is no undefined visible within experimental settings. Screenshots attached
Screenshot 2024-10-07 at 20 54 29
Screenshot 2024-10-07 at 20 54 22
Screenshot 2024-10-07 at 20 54 34

PR #717, opened 2024-10-03
feat(i18n): adds french language support

firebomb1:
Description
---

Adds French language support and translations.

Motivation and Context
---

Better i18n support.

How Has This Been Tested?
---

Locally.

---

Attempt to use the website after selecting the French language.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


Tas4tari:
Tested on Beta version 0.5.10:
Thee following text are still in English when I check the translations:
`Show History`, `Hide History`, `Testnet`

firebomb1:
> Tested on Beta version 0.5.10: Thee following text are still in English when I check the translations: `Show History`, `Hide History`, `Testnet`

Thanks for the testing, without more context it's hard to tell which page/section you're referring to.

Although "Testnet" will usually be translated "Testnet" in French.  A really literal translation, "Réseau de test" would work, but from experience the term "Testnet" is known especially for users who want to mine a token/coin, so "Testnet" is preferable.

I can't find any "Show History" / "Hide History" reference in the English files or even hardcoded in the rest of the code, you'll have to be more precise about where you've found them.

firebomb1:
Oh my bad, Hide and Show History are strings that have been added After this PR has been merged, that's why they aren't translated. So maybe next time the translations are updated they will be translated.

PR #726, opened 2024-10-04
fix: reset settings

MCozhusheck:
Description
---
Fix issue with reset setting getting stuck and also if set remove only wallet_config of current network.

Motivation and Context
---
Fix reset settings.

How Has This Been Tested?
---
Run application and reset setting after setup both with reset wallet and and without.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler (approved):
utACK

PR #729, opened 2024-10-04
fix(ui): removed a bad import from the airdrop login modal

peps:
This is just a lint fix, removing an unused import from the ClaimModal component.


stringhandler:
Done in a different pr

PR #749, opened 2024-10-08
feat(ui): added connection indicator to version chip

peps:
- Moved the chip in the sidebar header into it's own `VersionChip` component.
- Added a `ConnectedPulse` indicator component that shows yellow while connecting and green once connected.
- Updated version number to grey to match designs.

Question: Im using the `isSettingUp` value from `useAppStateStore` to check the status of the connection, are there other / different checks that should be made for proper integration? 

CleanShot 2024-10-08 at 15 48 08

### DIFF
--- a/src/containers/SideBar/components/VersionChip/VersionChip.tsx
+++ b/src/containers/SideBar/components/VersionChip/VersionChip.tsx
@@ -1,0 +1,24 @@
+import { useTranslation } from 'react-i18next';
+import ConnectedPulse from './ConnectedPulse/ConnectedPulse';
+import { Divider, Wrapper } from './styles';
+import { useAppStateStore } from '@app/store/appStateStore';
+
+interface Props {
+    version: string;
+}
+
+export default function VersionChip({ version }: Props) {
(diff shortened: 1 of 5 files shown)


src/containers/SideBar/components/VersionChip/VersionChip.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useAppStateStore } from '@app/store/appStateStore';
+
+interface Props {
+    version: string;
+}
+
+export default function VersionChip({ version }: Props) {
+    const { t } = useTranslation('common', { useSuspense: false });
+
+    const isAppSettingUp = useAppStateStore((s) => s.isSettingUp);
```
shanimal08, line 13: re your question - yeah this should be `base_node.is_connected` instead, already stored in `useMiningStore` :D example: used here in [`DebugSettings.tsx`](https://github.com/tari-project/universe/blob/68fa3c6711f9e0613ba5c12397d85fbf8a6c10f6/src/containers/Settings/sections/experimental/DebugSettings.tsx#L18)

shanimal08 (approved):

PR #766, opened 2024-10-08
feat: validate wallet conf matches existing wallet

brianp:
Description
---
If the config file doesn't match the wallet, then ditch them both and start fresh. This should be rare for users but happens frequently in development.

Avoid waiting for the wallet to start by reading the needed keys directly from the dir before the startup sequence occurs.

Motivation and Context
---
Prevent mismatches of the displayed settings and running wallet. If the config previously got deleted a new one would generate displaying a new address, but the running wallet wouldn't reflect that address.

How Has This Been Tested?
---
Manually

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -87,13 +87,12 @@ zip = "2.2.0"
 [target.'cfg(windows)'.dependencies]
 winreg = "0.52.0"
 
-# needed for keymanager. TODO: Find a way of creating a keymanager without bundling sqlite
 chrono = "0.4.38"
 device_query = "2.1.0"
-libsqlite3-sys = {version = "0.25.1", features = ["bundled"] }
 log = "0.4.22"
 nvml-wrapper = "0.10.0"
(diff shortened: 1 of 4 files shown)


src-tauri/Cargo.toml
```diff
@@ -74,8 +74,7 @@ xz2 = { version = "0.1.7", features = ["static"] }
 open = "5"
 thiserror = "1.0.26"
 
-# needed for keymanager. TODO: Find a way of creating a keymanager without bundling sqlite
-libsqlite3-sys = { version = "0.25.1", features = ["bundled"] }
```
stringhandler, line 78: I think this still needs to be bundled

stringhandler (commented):
happy, but I think we need to add the bundled feature to rusqlite

Zack435 (approved):

Zack435 (approved):

PR #790, opened 2024-10-09
fix(ui): simplified sidebar wallet show/hide history animation

peps:
This PR simplifies the wallet show/hide history animation. I basically removed most of the `layout` props so we have more control over the animation and there's less weirdness.

I changed it so the height of the history container is animated instead of the parent container of the wallet. The parent just adjusts to the new height of the animation automatically as it grows in height. 

I also removed some of the layout animations from the big number, just keeping it on the "tXLM" text.  Less stretching this way, and the show/hide number still looks good

https://www.loom.com/share/a9ee052711f34e429e106394cdc69f2d?sid=2c496820-66b9-4dd1-a726-3739db9d3498

### DIFF
--- a/src/components/CharSpinner/CharSpinner.tsx
+++ b/src/components/CharSpinner/CharSpinner.tsx
@@ -38,7 +38,6 @@ export default function CharSpinner({ value, variant = 'large', fontSize }: Char
         if (!isNum) {
             return (
                 <Characters
-                    layout
                     $decimal={isDec}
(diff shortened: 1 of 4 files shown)


stringhandler:
looks great

shanimal08 (approved):
slick! tested & lgtm!

Tas4tari:
Unable to test on Windows 11 as I have not won a block yet, when I do, I will confirm

PR #798, opened 2024-10-09
feat(binaries): prioritize github releases over mirror cdn

MCozhusheck:
Description
---
Reverse the order of choosing releases source of binaries.

Motivation and Context
---
CDN is being overwhelmed with downloads requests so we can just first try github releases to minimize traffic on CDN.

How Has This Been Tested?
---
Deleted all binaries and checked from which source we are downloading binaries.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/binaries/adapter_tor.rs
+++ b/src-tauri/src/binaries/adapter_tor.rs
@@ -22,7 +22,10 @@ impl LatestVersionApiAdapter for TorReleaseAdapter {
         let mut dist_responded = false;
 
         for _ in 0..3 {
-            let response = reqwest::get(dist_tor_bundle_url).await;
+            let response = reqwest::Client::new()
+                .head(dist_tor_bundle_url)
+                .send()
+                .await;
             if let Ok(resp) = response {
                 if resp.status().is_success() {
                     dist_responded = true;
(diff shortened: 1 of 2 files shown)


src-tauri/src/binaries/adapter_tor.rs
```diff
@@ -17,37 +17,37 @@ pub(crate) struct TorReleaseAdapter {}
...
 impl LatestVersionApiAdapter for TorReleaseAdapter {
     async fn fetch_releases_list(&self) -> Result<Vec<VersionDownloadInfo>, Error> {
+        let dist_tor_bundle_url = "https://dist.torproject.org/torbrowser/13.5.6/tor-expert-bundle-windows-x86_64-13.5.6.tar.gz";
         let cdn_tor_bundle_url = "https://cdn-universe.tari.com/torbrowser/13.5.6/tor-expert-bundle-windows-x86_64-13.5.6.tar.gz";
-        let mut cdn_responded = false;
+        let mut dist_responded = false;
 
         for _ in 0..3 {
-            let response = reqwest::get(cdn_tor_bundle_url).await;
+            let response = reqwest::get(dist_tor_bundle_url).await;
```
stringhandler, line 25: You are actually downloading the whole file here. Try a smaller file or use a method other than GET. Maybe HEAD or somethign?

MCozhusheck: fixed

stringhandler:
Closing. In 0.5.20, it will find in both, first choosing the mirror

PR #804, opened 2024-10-10
fix(ui): use correct font family for reward

shanimal08:
Description
---
- just used `DrukWide` in the earnings animation text


stringhandler (approved):

Tas4tari:
Unablle to test on v0.5.17on Windows 11 as I have won any blocks yet

Issue #742, opened 2024-10-07
Add a warning when mining under 250 h/s


metalaureate:
**Describe the issue**
Use the connection warning affordance to warn if mining is cpu hash rate is >0 but less than 250 h/s

Warning: CPU power too low to mine.

stringhandler:
"Your hash rate is very low. Try closing all apps other than Tari Universe"

@metalaureate please provide better wording when you can

PR #809, opened 2024-10-10
feat: add low hash rate warning

MCozhusheck:
Description
---
implements: #742 
Adds warning under the mine button about low hash rate.
Screenshot from 2024-10-10 15-45-40

Motivation and Context
---
Miner may enter into very slow mining mode if there is not enough free RAM. 

How Has This Been Tested?
---
Increased hash rate above my CPU capability to trigger warning.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/public/locales/af/settings.json
+++ b/public/locales/af/settings.json
@@ -59,5 +59,6 @@
   "use-tor": "Gebruik Tor",
   "importing-wallet": "Importing wallet",
   "action-requires-restart": "This action requires restart to apply changes",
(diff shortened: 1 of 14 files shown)


brianp (approved):
LGTM

Tested by running the tari_miner before starting Universe. Dialog showed up as expected under 250h/s and after turning tari miner off the hash increased and the warning went away.

Tas4tari:
Tested successfully on release version 0.5.22 on Windows 11, version 23H2
12th Gen Intel(R) Core(TM) i7-1255-U 1.70GHz

Issue #611, opened 2024-09-25
Gpu hardware stats device list


karczuRF:
Refactor hardware parameters to return not only the single (first) one GPU device, but all detected ones, allowing for better gpu devices management (e.g. turning on/off specified) and controlling params like hashrate.

Example of use:
![Image](https://github.com/user-attachments/assets/d58e3b5a-f373-427f-9d91-e3e7b911037f)

uforiaio:
After v0.4.x (?) the Intel on chip gpu is the only used one. Earlier versions didn't have the same behavior, but to be fair I wasn't watching close. Not sure if the gpu stats list has a guid/uuid that uses to order the cards. It's obviously not alphabetical. 

![image](https://github.com/user-attachments/assets/15fb572d-e326-4656-a9af-b380327d6d68)

brianp:
Improvements made. Closing issue.

Issue #497, opened 2024-09-17
GPU miner: turn on/off a specific device


karczuRF:
- Detect which devices work and which ones are not supported. Pass and show that info to the user. 

- People have asked for the ability to turn on and off devices, so maybe we can pass in a string like `xtrgpuminer --use-devices=1,3,4` which skips 2 device. We should  also add `--exclude-devices=2` which is another way of saying the same thing

PR #824, opened 2024-10-11
feat: gpu general improvements

karczuRF:
Description
General improvements for gpu miner.
**Notice! Because of the changes in tarigpuminer repo this works only with version `xtrgpuminer >=0.1.13`**

Resolves:
- #610  
- #611 
- #497 

Motivation and Context
This PR includes changes #612  and #675  and was made to merge the code and use `gpu_status.json` file generated by tarigpuminer while devices detection.

image

How Has This Been Tested?
Manually on Linux with single integrated Intel gpu device. 

Test on Windows and macOS to see if devices are listed properly and it is possible to turn them on/off.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (changes requested):
This PR is looking really good. Tested on M1 with two versions of the gpuminer. One supporting M1 and one not. The experimental settings reflected expectations. Older version of GPUMiner does not support M1 and would not allow for GPU mining. New version of GPU miner allowed for mining.

## Two issues that need fixing:

1. When the miners are not running they still show a Usage percentage 
image

Expectation: Show `N/A` for usage when either miner is not running.

2. When GPU mining and we switch to `Ludicrous` mode the stats never reload. The spinner remains indefinitely.

https://github.com/user-attachments/assets/894913cd-5bf9-4ed4-951c-b7e0f31c17b2

Expectation: The stats should reload

karczuRF:
Thanks for feedback. I'll implement your points

Tas4tari:
Tested on versions: 0.5.20 & Release version 0.5.22:
Unable to test as GPU does not run at all as there are no GPU devices detected
Ticket logged:
https://github.com/tari-project/universe/issues/843

PR #835, opened 2024-10-14
feat(ui): and fix: restart dialog component | re-usable state item | add restart trigger …

shanimal08:
…to tor toggle

Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/public/locales/en/settings.json
+++ b/public/locales/en/settings.json
@@ -1,70 +1,92 @@
 {
+    "action-requires-restart": "This action requires restart to apply changes",
     "applyInviteCode": "Apply Invite Code",
-    "settings": "Settings",
-    "seed-words": "Seed words",
-    "send-logs": "Submit an Issue & Logs",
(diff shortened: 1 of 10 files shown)


brianp (approved):
If you cancel the restart dialog the settings are persisted, but not applied. When reviewing the settings it will not look like that setting is applied even though it isn't. 

I don't think this is a blocker for this issue, as the dialog is mostly about the need to restart, but it's noteworthy and maybe something that needs to be revistited later.

https://github.com/user-attachments/assets/46480991-3ad4-4070-815a-59b31590b070

Tas4tari:
@shanimal08 Tested on Windows 11, app version 0.5.22 successfully. Reeceuved the same issue as @brianp and have logged a seperate ticket for it:
https://github.com/tari-project/universe/issues/849

PR #841, opened 2024-10-14
fix: add mmproxy monitoring and better health checking

stringhandler:


brianp:
Didn't want you to feel left out. So I tested it. It ran. Didn't see any errors.

Tas4tari:
Tested the ffg way on Windows 11, app version 0.5.22:
I had the universe app open and running, then I went to task manager and ended all minotari processes, when i ended them they all started up again almost immediately

Issue #133, opened 2024-08-24
Dark mode


metalaureate:
**Describe the bug**
I blew my eyeballs out.

brianp:
Closing as won't fix. I'll never give you dark mode. White screens for everyone!

PR #850, opened 2024-10-15
feat(ui): dark mode

shanimal08:
## Description

it's the same, but dark.

**theming (ui):**

- added `palettes` directory and created dark/light theme files
- standardized colours
- added better types for theme objects
- ensured that we're using `theme.palette.x` in styling where there is or would be a switch (majority of the file changes 😅)
- updated components and sections where necessary to work with theming changes from design QA, e.g.:
  - removed the background cloud image from the Soon setup screen
  - adjusted the `LinearProgress` component styling
- tower animation:
  - new build with added Dark Mode colour scheme
  - added `setAnimationProperties` function for the toggle

**theming (logic):**

- added `theme` and `set_theme` to app_config (`system`, `dark`, and `light`)
- added UI store theme field, so we get the saved theme setting from app_config and use dark/light based on save settings or system
- added `useDetectMode` helper hook to listen for (system) theme changes

**other necessary updates:**

- added `very_e`(xperimental) flag to app_config and Very Experimental settings section
- cleaned up our Button components:
  - changed `ButtonBase` which had the new styling for buttons (rounded) to `Button` as our default button
  - split out the old `Button` to `SquaredButton`, `IconButton`, and `TextButton` instead of having 600 props for variants
  - updated styling options for button colours
- added new `RadioButton` component
- cleaned up Settings Navigation and added show/hide of super experimental settings

## Motivation and Context

- https://github.com/tari-project/universe/issues/133

## How Has This Been Tested?

**switching modes (incl. system mode):**

https://github.com/user-attachments/assets/85b3fe96-7237-4497-88c9-c3c7d94b4fdc

**dark mode screens:**

https://github.com/user-attachments/assets/80cfc997-40ee-4a70-9e95-6c955d975c36

https://github.com/user-attachments/assets/fb09b873-76c7-4ab7-b207-c880d4544960

https://github.com/user-attachments/assets/7f20ba6e-7d38-4f77-b563-90ee2f50ceee


shanimal08:
@PanchoBubble ty for testing :D 

> I know these are experimental and need to be changed regardless

think i should rather just use the normal green toggle here? :D 


> this component has black shadow over black bg 

i'm not sure what to change/fix? 😅  for box shadows we're keeping them black even on dark mode because the contrast colour (white:p) looks really silly

mmrrnn:
I played with this a bit, it looks cool! :D

PR #861, opened 2024-10-16
feat(ui): staged security modals ui

peps:


brianp:
> The only pieces not integrated are...
    1. The trigger to show the modal. I think it's supposed to pop up when the user reaches a certain threshold. I'm just not sure what that is specifically.
    2. We should probably store in local storage if the has seen the modal... so it only pops for them once.

Are there plans to add this, does this need assistance? 

1. As for the threshold. Aurora does this at 10000 MicroTari 
2. There's a heavy lean against using localstorage. Instead, make it an appconfig field.

Once those have been completed I'd happily test and merge this.

peps:
> > The only pieces not integrated are...
> > 1. The trigger to show the modal. I think it's supposed to pop up when the user reaches a certain threshold. I'm just not sure what that is specifically.
> > 2. We should probably store in local storage if the has seen the modal... so it only pops for them once.
> 
> Are there plans to add this, does this need assistance?
> 
> 1. As for the threshold. Aurora does this at 10000 MicroTari
> 2. There's a heavy lean against using localstorage. Instead, make it an appconfig field.
> 
> Once those have been completed I'd happily test and merge this.

Thanks @brianp! I've passed this over to @PanchoBubble to complete the integrations, converting to Draft for now while it's still being worked on. 

@PanchoBubble we might also be able to just merge this, then complete integration in a separate PR. Right now this isn't triggered by anything, so it wont pop up for anyone and we would just have these components ready to go in `main`. WDYT?

brianp:
@stringhandler how do you feel about the base of this being merged before it's ready for use? 

I have not actually reviewed the changes yet, but will if we want to merge the bulk first.

PanchoBubble:
@brianp yes I agree
I just confirmed this is not meant to go live until main net, I'd say lets get this one merged and we can integrate the functionality later down the line when we get information about at what threshold it should get triggered and stuff 
 @stringhandler

brianp (approved):
LGTM utAck

PR #865, opened 2024-10-16
fix: ensure single init

### LINKED ISSUE #864: Fix double initialization

MCozhusheck:
Description
---
Fixes #864 
Double initialization caused invalid config in which port of the mmproxy was invalid and also caused sometimes progress bar to go back to 0% 

Motivation and Context
---
Fix mining

How Has This Been Tested?
---
Run app and check if xmrig is using resources.

---
Check if there are 2 warning logs coming from initialization with description how much time it took. In my case these were:
```
14:55:11 WARN  setup_application took too long: 66.029057277s
14:55:11 WARN  setup_application took too long: 66.252814494s
```
This indicates that there were 2 processes running along. After the fix there should be only 1 log with time it took.

Another giveaway of invalid init is progress bar moving backward sometimes - this indicates that's there is another process messing with progress bar. This also shouldn't happen anymore.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (approved):
I couldn't replicate the double init issue locally. Maybe it's OS dependent? But I'd be happy to merge this as the mmproxy issue was a blocker.

PR #875, opened 2024-10-17
fix: use dynamic ports instead of fixed for wallet and upgrade tor

stringhandler:
Uses dynamic ports where we were using fixed ports.

This should make running beta and production versions at the same time better.

Also upgrades tor


Anziskavds:
Tested on a MacBook Air 13 inch M1 2020 on Testnet v0.5.31 and Beta v0.5.26. I can see a improvement on running both prod and beta builds on my machine.

PR #880, opened 2024-10-17
chore(logging): optimize logging and initialization

Krakaw:
Description
---
The frontend was controlling the application initialization process, which means we got caught with some lifecycle issues where a useEffect hook was triggering the `setup_application` invocation multiple times. I added a little debounce mechanism to the call.
We had unnecessary logging, moved some to trace as well as only emit a log if a state has changed.

Motivation and Context
---
The logs were overflowing with useless information. This just streamlines it a little.
The app was needlessly resetting the application state.

How Has This Been Tested?
---
I ran the app, the logs are cleaner and there is less duplication of initialization.

---
Run the app, make sure nothing is broken.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/hardware_monitor.rs
+++ b/src-tauri/src/hardware_monitor.rs
@@ -1,7 +1,7 @@
 use std::{fs, ops::Deref, path::PathBuf, sync::LazyLock};
 
 use anyhow::anyhow;
-use log::{debug, warn};
+use log::{debug, trace, warn};
 use nvml_wrapper::{enum_wrappers::device::TemperatureSensor, Nvml};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


stringhandler (approved):

Anziskavds:
Tested by doing a reset, then start up and looked if the log files are smaller. It tested unsuccessful. The log files was bigger after the reset.

Anziskavds:
Message from Keith 
image. I followed steps received from Mike but it's not the correct requirements. As per the requiremnt Keith provided it tested successfully on a MacBook Air 13 inch M1 2020 on Testnet 0.5.31.

PR #888, opened 2024-10-18
feat: add id for block in telemetry_manager

BalazsSevecsek:


brianp (changes requested):
See DM's

brianp (approved):
Mining works now. Will wait for confirmation and merge if good.

PR #712, opened 2024-10-03
fix(firewall): handle firewall rules on windows

mmrrnn:

PR #897, opened 2024-10-21
feat: include vc in installer

### LINKED ISSUE #1555: Missing MSVCP140.dll file crashes the app on Windows
**Describe the bug**
Tari Universe is crashing when system doesn't have `Microsoft Visual C++ Redistributable` installed.

We handled this issue before, I believe it's because of upgrade to Tauri v2. Missing dependecy checker is triggered too late. 

**To Reproduce**
1. Uninstall all Microsoft Visual C++ packages from your windows(restart)
2. Run the app
3. See error

**Expected behavior**
This error should be handled by us, we should prevent from the app crash.

**Screenshots**

Image

 - Windows 11

Misieq01:
##  [ Improvment ] => Microsoft Visual C++ Redistributable [ WINDOWS ONLY ]
Extended installer with setup for mentioned package
------------
### [ Features ]
- Created new bundler in => `tari-win-bundler` directory which combines Tari universe installer.msi and vc_redist.exe
- Extended CI workflow `release` with setup required for building mentioned bundle and attaching it to release
- Add some styling to new installer
- Fix autolaunch button at the end of installer

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Windows installer for a more intuitive setup experience with a modern, localized user interface.
  - Introduced dynamic naming that clearly differentiates between beta and official release builds.
  - Improved dependency handling and installer signing to ensure a smoother and more secure installation process for end-users.


leet4tari:
This is a pretty radical change for a very small number of users, thou it should be addressed some how. Other users might see this as a download tax, that they don't need.

I was thinking we could possible do this build option as a full offline installer, but would make this package option pretty big.  
  
Could also include the tor package, as this has been requested a few times to be bundled. Windows Bundle options for Tauri can be found - https://v1.tauri.app/v1/guides/building/windows/#webview2-installation-options.

Not sure why anybody would use an offline installer as internet is needed for any useful usage of the package, but it's an idea to put all this in one place.

Misieq01:
> This is a pretty radical change for a very small number of users, thou it should be addressed some how. Other users might see this as a download tax, that they don't need.
> 
> I was thinking we could possible do this build option as a full offline installer, but would make this package option pretty big.
> 
> Could also include the tor package, as this has been requested a few times to be bundled. Windows Bundle options for Tauri can be found - https://v1.tauri.app/v1/guides/building/windows/#webview2-installation-options.
> 
> Not sure why anybody would use an offline installer as internet is needed for any useful usage of the package, but it's an idea to put all this in one place.

I don't think its affect very small number of users. It's affects all future windows installation when users does not have vc_redist already on the machine so this number can get big when tari will be released to wider audience. 

Also the documentation that you provided does not refer to this problem as its only WebView2 installation configuration, its not like we can use it to include more packages

As for including others packages as tor, gpu miner I think we could do it with tauri options for configuring wix termplate as long as there is no extra installer in place.

If we want to include vc_redist then all work done in this PR is required

brianp:
@leet4tari I can look at the changes here but most of it won't mean anything to me. So I'm going to lean on your review if you could officially make one.

And testing... Any chance you can test it?

brianp (approved):
LGTMy untrained windows knowledge 👍🏻 🙃

Misieq01:
> @leet4tari I can look at the changes here but most of it won't mean anything to me. So I'm going to lean on your review if you could officially make one.
> 
> And testing... Any chance you can test it?

@brianp if you would like to check the installer, the only way now is to download it from releases on my fork https://github.com/Misieq01/universe/releases => Tari.Universe_0.5.22_x64_en-US.exe

leet4tari:
I think this can be paused for now, but should be in place by release as this will be something that comes up alot. My view on combining two problems into one solution would mean we get to solve many problems with one download.  ie: include vc_redist, solve this problem. Include tor, solve that download problem. Include other tools, solve offline single-shot download and install. Down side is huge download, but I only seeing that happen for initial install, else download updates as needed.

brianp:
Related to #712 

Both will wait until next week for updates, and firewall logic may be moved into the installer logic. In any case we probably need to wait for p2pool to be complete before this will get updated/merger.

leet4tari:
Should we rather not download the latest ```VC_redist.x64.exe``` to include when building? Instead of keeping an old version in our repo. Do we really need ```VC_redist.x86.exe``` as it's for the 32bit Windows?

Misieq01:
> Should we rather not download the latest `VC_redist.x64.exe` to include when building? Instead of keeping an old version in our repo. Do we really need `VC_redist.x86.exe` as it's for the 32bit Windows?

With downloading or using one included in files I'm fine either way. I can remove the 32 bit version and switch to online download of vc_redist if you want

leet4tari (approved):
btACK

PR #898, opened 2024-10-21
feat: detect orphan chain

MCozhusheck:
Description
---
Every 30 seconds fetch block from current tip minus 1000, 2000 and 3000 from both local node and text explorer and compares these 2. If they are different for just 1 case, send event to frontend which displays warning.

Motivation and Context
---
Some miners are stuck at orphan chains. This will check if it's a case for them and display information.

How Has This Been Tested?
---
Change condition to always return true in `check_if_is_orphan_chain` in `node_manager.rs` and check if alert under mining button is displayed.

Also replicated orphan chain by changing in experimental settings to not use tor and manually ban tcp peers in `minotari node` to make sure you are that there are no connected peers.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


stringhandler:
tested, but needs to be the other way around. It should read blocks from the explorer and see if they are in the current chain

mmrrnn:
Resolved conflicts on separate branch due to the missing write permissions: https://github.com/tari-project/universe/pull/1018

mmrrnn (approved):

brianp (changes requested):
Mining can't start when I'm on this branch. 

I remember now I had actually tested this Friday but didn't comment.

image

brianp:
I was testing an outdated branch. Disregard the last message. Even the screenshot reads v0.6.6. 

Once I was on the correct branch it was notable the i18n may be missing or misconfigured:

image

Issue #842, opened 2024-10-14
Miner stat display


brianp:
## Two issues that need fixing:

1. When the miners are not running they still show a Usage percentage 
<img width="344" alt="image" src="https://github.com/user-attachments/assets/1e5b3b91-90c6-4747-aa22-97c9732b7ebc">

Expectation: Show `N/A` for usage when either miner is not running.

2. When GPU mining and we switch to `Ludicrous` mode the stats take about 2minutes to reload, where they show up immediately when on eco. 

https://github.com/user-attachments/assets/894913cd-5bf9-4ed4-951c-b7e0f31c17b2

Expectation: The stats should reload

PR #900, opened 2024-10-21
fix(ui): hide chip if c/gpu enabled | trigger interval fetch on mode change

shanimal08:
Description
---

- only display CPU/GPU usage chip if that mining type is enabled and mining
- check for est. earnings balance so we don't render `0`
- ~~updated `useMiningStatesSync` to check for mining mode changes so the interval items are changed sooner than waiting for the interval - probably a MINOR improvement if any, as it's just a 1 second interval. i couldn't reproduce point two on the issue (neither are instant for me:p) but thought i'd add it in case it may help)~~ reverted this since it's such a quick interval

Motivation and Context
---
- resolves the first part of https://github.com/tari-project/universe/issues/842

How Has This Been Tested?
---
- locally

image

---
- switch mining modes with GPU mining only
- toggle GPU/CPU mining and see whether the usage chips are still there or not


brianp (approved):
LGTM

Tested and works as expected.

Tas4tari:
Tested on Windows 11, app version 0.5.60 and works correctly for CPU and . Unable to test GPU as GPU is not detected

PR #911, opened 2024-10-22
fix: app freeze on app exit

Misieq01:
### [ Solution Description ]
From what I understand trying to shutdown app can cause processes to get `is_healthy` flag marked as false and start restarting mechanism with retries and ignoring other branches of `select!`
Adding checking of `is_triggered` flag for shutdown handlers seems to solve this problem

------------
### [ Testing ]
- Stopping and starting mining should work as expected
- shutting down app should take no longer then 5 - 10 seconds


Anziskavds:
Tested successfully on Testnet v0.5.45 on a MacBook Air 13 inch M1 2020.

PR #912, opened 2024-10-22
fix: populate-user-agent-header

Misieq01:
### [ Content ]
Added User-agent header to request that had it missing or was not descriptive


src-tauri/src/binaries/adapter_tor.rs
```diff
@@ -71,7 +71,14 @@ impl LatestVersionApiAdapter for TorReleaseAdapter {
             .join(format!("{}.asc", asset.name));
         let checksum_url = format!("{}.asc", asset.url);
 
-        match download_file_with_retries(&checksum_url, &checksum_path, progress_tracker).await {
+        let user_agent = format!(
+            "universe {}({}) | {}",
+            env!("CARGO_PKG_VERSION"),
+            std::env::consts::OS,
+            "github adapter"
```
brianp, line 78: Should this say "tor adapter" ?

```suggestion
            "tor adapter"
```

brianp (changes requested):
I don't think we need different user-agents per adapter github/tor/binary download. All the agents could just say universe with version and platform, and if this is the case it could be a static var instead of defined per file.

/cc @leet4tari opinions?

src-tauri/src/binaries/adapter_tor.rs
leet4tari: Agree, only TU version and platform info I think is useful. Getting reports on TU version or platform usage from CDN provider is possible. Different ```adapter``` would not request resources from different services.

Misieq01:
Closing in favor of https://github.com/tari-project/universe/pull/935

PR #917, opened 2024-10-23
chore(ci): bump tari_crypto and tari_utilities to drop old ring v0.16 dependency

leet4tari:
Description
bump tari_crypto and tari_utilities to drop old ring v0.16 dependency

Motivation and Context
Remove older dependencies and improve cross-compile options.

How Has This Been Tested?
Builds in local fork for all current platforms
Also builds in docker on Apple M1 for Linux ARM64


brianp (approved):
LGTM

Tested, successfully mined blocks, addresses stayed the same.

PR #936, opened 2024-10-25
fix(ui): update floating ui elements

shanimal08:
Description
---
- updated the `Dialog` and `ErrorSnackbar` components so their dismiss functionalities don't conflict with each other
- styling: moved snack bar to middle of screen

Motivation and Context
---
noticed if a dialog/modal was open and you got an error, you couldn't dismiss the error snack with a click (not super important because they auto-close, but this structure is better for if we ever change that)

How Has This Been Tested?
---
- locally

---

if running locally you can comment out the auto-close useEffect in `ErrorSnackbar` and set `error` to any string so it shows

- open settings modal
- make sure you can dismiss the error

extra: 

- open settings modal
- open send logs modal in settings
- see that: 
    1. you can dismiss the error 
    2. dismissing the error doesn't dismiss the send logs modal
    
if _not_ running locally you need to somehow get an error to display, and then follow the above steps quicker than 5 seconds 😅


brianp (approved):
LGTM. Untested. Testing instructions required too much JS.

#yolo

PR #945, opened 2024-10-26
feat(i18n): script generating json translations

mmrrnn:
Motivation and Context
---
We spend so much time on adding new i18n translations to the JSON files so I decided to automate this process a bit.

General Info
---
We can now generate json boilerplate using npm script. It also can be used to solve conflicts related to the i18n.

Description
---
Script updates translation files(for all nationalities - english translation by default to be translated in future - maybe we could use ai to automate further??) by setting a nested value in JSON files located in the **public/locales** directory. It takes a file name and a key-value pair as arguments, where the key is a dot-separated string representing the nested property to update, and the value is the new value to set.

Usage
---
```npm run translate <fileName> <key="value">```
```npm run translate sidebar hide-history="Hide history"```
```npm run translate settings idle-timeout.title="Some timout title"```


brianp:
Went to test. Have conflcits. Am unsure how to use the changes to resolve the conflicts.

mmrrnn:
Done, conflicts in this PR will be painful since reodering, unifying formatting. But once merge it should be handy

PR #948, opened 2024-10-28
fix(ui): cannot rotate the 3D canvas with mouse drag

peps:
I noticed a bug where the drag events on the 3D canvas element were being blocked. `pointer-events: none` was not being respected in the browser. Looks like this Global `all: unset;` declaration was overwriting the `pointer-events` rule on div elements. 

Removing `div` from this reset fixes the issue. `div` elements should not need a reset like this anyways since they don't really have any default styles to begin with. 

https://www.loom.com/share/ed56286a903d4c98b54745732c66a092?sid=31821a02-5d33-47f6-b803-93149a6d2bb1

### DIFF
--- a/src/theme/GlobalStyle.ts
+++ b/src/theme/GlobalStyle.ts
@@ -19,7 +19,6 @@ export const GlobalReset = createGlobalStyle`
     fieldset,
     textarea,
     dialog,
-    div,
     button,
     input {
         all: unset;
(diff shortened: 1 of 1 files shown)


shanimal08 (approved):
thx 😅 sorry added it because the modals were being weird on click - you could probably remove the `button,` from there too - that's what borked the cursor:pointer stuff 🙈

peps:
cool, for the record, ill do a full sweep of the app adding proper tab capture navigation later this week.

PR #954, opened 2024-10-28
fix: release and build workflow for sentry

shanimal08:
fix releases for sentry:

- add `max_old_space_size` to release.yml
- add build.sourcemaps back to vite.config
- add more specific config for the sentry plugin souremap config

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,237 +2,237 @@
 name: Release
 
 'on':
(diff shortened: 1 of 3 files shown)


stringhandler (approved):

Tas4tari:
Spoke to @shanimal08 on how to confirm this and she has confirmed that the source maps are on Sentry

PR #976, opened 2024-10-30
fix(tor): restaring issue on unix

mmrrnn:
BACKGROUND
---
Tor might be already running and using control/socks port. Tor ends in infinite restart loop.

SOLUTION
---
Check if control/socks port are already used and use random one which is available.
This check is done when loading tor's config. Port replacement is done per session, we don't overwrite it in the config file


mmrrnn:
TESTING ON MULTIPLE DEVICES, created PR to just pull changes

brianp:
Closing. This PR did not fix the issue and we reverted bridges on mac: https://github.com/tari-project/universe/pull/982 && https://github.com/tari-project/universe/pull/991

PR #982, opened 2024-11-01
fix: disable tor bridges settings for macos

shanimal08:
Description
---

- update Tor settings:
  - check for OS
  - disable extra Tor settings UI if macOS, with explainer copy
- other changes:
  - fix text button colour check
  - rm console.gebug

How Has This Been Tested?
---

- locally: 

https://github.com/user-attachments/assets/bd8f598e-ebf5-4bda-bed4-0214dabadcfc

---

- run on a mac 
- try use Tor settings
- only off/on should be allowed on mac


Anziskavds:
Tested successfully on MacBook Air 13 inch M1 2020 on Testnet v0.5.63.

PR #985, opened 2024-11-01
fix: implement better tor settings

stringhandler:
Uses a random tor port when starting.
Disabled tor settings in code for mac

### DIFF
--- a/src-tauri/src/node_adapter.rs
+++ b/src-tauri/src/node_adapter.rs
@@ -109,8 +109,8 @@ impl ProcessAdapter for MinotariNodeAdapter {
             if cfg!(target_os = "windows") {
                 // No need
             } else {
-                args.push("-p".to_string());
-                args.push("use_libtor=false".to_string());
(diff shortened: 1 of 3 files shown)


src-tauri/src/node_adapter.rs
```diff
@@ -109,8 +109,8 @@ impl ProcessAdapter for MinotariNodeAdapter {
             if cfg!(target_os = "windows") {
                 // No need
             } else {
-                args.push("-p".to_string());
-                args.push("use_libtor=false".to_string());
+                // args.push("-p".to_string());
+                // args.push("use_libtor=false".to_string());
```
mmrrnn, line 113: Do we use libtor on linux too? I see inconsistency with main.rs

mmrrnn: I fixed this issue on my separate branch. We use libtor only for macos with missing ui changes:
https://github.com/tari-project/universe/pull/991

PR #1005, opened 2024-11-05
feat: customize msi installer

Misieq01:
### [ Summary ]

- **User access control** => from what I saw its probably tied to certificate that we generate and could not be change inside wixtoolset
- **Icon of a installer** => Windows do not provide option for changing icon of `.msi` installer

- Current implementation of installer provides very little options for theming, we will be able to do more and also change installer icon on installer version from this PR: #897

------------
### [ What can be done  ]

- We can change all texts that all displayed through installer
- We probably can change fonts colors, sizes etc.
- We can change background image on the starting page (It needs to be in **`.bmp format and exactly 493x312px`**)
 image_720
- We can change banner image on all pages that displays it, can only be one and also in **`.bmp format with 493x58px`**
 image_480


Misieq01:
Closing in favor of #897

PR #1007, opened 2024-11-05
fix: power levels adjustments

PanchoBubble:
We need to:

Fix the overlapping marker and labels

Change slider to % increments but say x of y cores or x of y threads on slider description so that it communicates both a % progress and the core/thread increments

Revise the save state to be explicit, since it takes about 30 seconds to apply

It currently defaults to 1 irrespective of whether Eco or Ludicrous is set, which is confusing: it should default to whatever the user had selected (the Eco level or the Ludicrous level)

________________

https://www.loom.com/share/9e7dc4c1123c4701b8032c55a2861f83

Fixes: #1016


src-tauri/src/main.rs
```diff
@@ -171,7 +171,7 @@ async fn set_mode(
     info!(target: LOG_TARGET, "set_mode called with mode: {:?}, custom_max_cpu_usage: {:?}, custom_max_gpu_usage: {:?}", mode, custom_cpu_usage, custom_gpu_usage);
 
     fn f64_to_isize_safe(_value: Option<f64>) -> Option<isize> {
-        let value = _value.unwrap_or(-1.0);
+        let value = _value.unwrap_or_default();
```
brianp, line 174: `_value` when used should be renamed to `value` as the underscore is used to denote variables that are unused.

brianp (changes requested):
Tested. The swapping to Ludircus bug is fixed. But overall the sliders now make the UX incredibly buggy/jumpy, and the settings appear to be saving before explicitly selecting save.

https://github.com/user-attachments/assets/d8b12149-8ce1-4911-a2d1-3a47849e2113

src-tauri/src/main.rs
PanchoBubble: checking

PanchoBubble:
@brianp 
https://www.loom.com/share/aa2af6eed0c246e6aecafcfb51265f58

Let me know if you still face these ux/ui issues, looks like styled components prop updates was causing a lot ofissues for some reason

brianp:
Looking much better

Anziskavds:
Tested successfully on MacBook Air 13 inch M1 2020 on Testnet v0.6.4.

PR #1009, opened 2024-11-06
fix(ui): update the transaction share link querystring to use blockheight

peps:
- Updated the share link to use `bh` instead of `tx_id`.
- Added a `max-height` media query to the modal so it fits on small screens. 

CleanShot 2024-11-06 at 15 05 26


src/store/useShareRewardStore.ts
```diff
@@ -13,7 +13,7 @@ interface Actions {
 }
 
 const initialState: State = {
-    sharingEnabled: false,
+    sharingEnabled: false, // set this to true to enable sharing
```
peps, line 16: TO view the sharing buttons UI, change this to true

Issue #1012, opened 2024-11-06
0.5.63: Windows 11: Telemetry toggle showing on start up off the app


Tas4tari:
**Describe the bug**
As a user after initial installation of the app I find that whenever I restart, I see the Telemetry option visible on start up bug is.

**To Reproduce**
1. User has tari universe installed and open
2. Close the app
3. Ensure all processes have ended
4. Click on the app icon to open the app
5. Observe the start screen when app is open

**Expected behavior**
Telemetry option should only be visible to the user upon initial installation of the tari universe app

**Screenshots**
![Screenshot (190)](https://github.com/user-attachments/assets/7a82ec38-8dfa-459f-882e-33b5a487884d)

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11, version 23H2

PR #1015, opened 2024-11-06
fix(ui): add flag to config & fix permissions toggle init

shanimal08:
Description
---
- remove sharingEnabled flag from js and move to app_config
- fixed setting the iniiiiiitial `seenPermissions` value and check if already set in `useAirdropStore`

Motivation & context
---

- resolves https://github.com/tari-project/universe/issues/1012
- easier testing

How Has This Been Tested?
---

- locally

---

- update `sharing_enabled` app_config and check if true/false work as expected
- if it's the very first install permissions toggle should be visible on the setup screen, else it shouldn't be

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -9,12 +9,12 @@ version = "0.5.63"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [build-dependencies]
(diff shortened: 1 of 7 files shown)


brianp (approved):

Tas4tari:
Tested unsuccessfully on Windows 11, app version 0.6.4. I still see the Telemetry option visible during every start up or restart of the app

PR #1018, opened 2024-11-06
feat(node): detect orphan chain

mmrrnn:
I needed to recreate https://github.com/tari-project/universe/pull/898 because I don't have write permissions.


brianp:
It could be an issue being outdated from main, *but* on this branch mining won't start for me.

mmrrnn:
Maciek is back so I think he'll continue with this PR here: https://github.com/tari-project/universe/pull/898

PR #1019, opened 2024-11-06
fix: unresponsive state when cpu gpu mining are off

### LINKED ISSUE #635: [0.4.5 MacBook Air M1 2020] Start Mining button enters busy, unresponsive state if a user turns off GPU/CPU mining while Universe is mining
**Describe the bug**
When mining, if a user goes into the settings and toggles off the GPU and CPU mining without pausing the miner first, the button  changes to "Start Mining", but with a spinning busy icon which prevents the user from interacting with the button. Turning either back on reverts back to the "Pause Mining" state.

**To Reproduce**
1. Make sure Universe is running and mining.
2. Go directly into the Settings and toggle off both GPU and CPU mining. Note the result on the Mining Button (shifts from Pause Mining to Start Mining, enters spin/busy state).
3. Start up either GPU or CPU mining, note the state change on the button.

**Expected behavior**
I cannot speak to the intended behaviour, but I would assume that the button should, when it transitions to "Start Mining", that it either remains active (in that a user could press the button then receive some sort of error message to indicate that either CPU or GPU mining must be active in the settings as a minimum to begin) or greyed out so the user cannot interact with it until they've activated either GPU or CPU mining in the settings.

OS & Version: macOS 14.6.1

Misieq01:
### [ Content ]
- Adjusted if statements in `setGpuMiningEnabled` and `setCpuMiningEnabled`


src/store/useAppConfigStore.ts
```diff
@@ -163,10 +165,16 @@ export const useAppConfigStore = create<AppConfigStoreState>()((set, getState) =
         if (miningState.cpu.mining.is_mining || miningState.gpu.mining.is_mining) {
             await miningState.pauseMining();
         }
+
+        console.log('miningState.miningInitiated', miningState.miningInitiated);
+        console.log('miningState.cpu.mining.is_mining', miningState.cpu.mining.is_mining);
+        console.log('miningState.gpu.mining.is_mining', miningState.gpu.mining.is_mining);
```
mmrrnn, line 171: remove console.logs

Anziskavds:
Tested successfully on MacBook Air 13 inch M1 2020 on Testnet v0.6.4.

PR #1036, opened 2024-11-07
fix: gpuDevices undefined causes crash in settings

peps:
There's a bug in the `Settings / Mining` section that causes a UI crash.  When gpuDevices is `undefined` it breaks. This PR adds an extra check to fix this. 

CleanShot 2024-11-07 at 17 13 29

CleanShot 2024-11-07 at 17 21 32

CleanShot 2024-11-07 at 17 17 08

CleanShot 2024-11-07 at 17 24 09


PanchoBubble (approved):

stringhandler:
fixed in a different pr

PR #1038, opened 2024-11-07
fix: remove requirement for opencl and change gpu power levels to use grid size instead of percent

stringhandler:


src-tauri/src/main.rs
```diff
@@ -220,31 +202,8 @@ async fn get_max_consumption_levels() -> Result<HashMap<String, i32>, String> {
...
-        Ok(i32::try_from(max_wg_size).unwrap_or(0))
-    })() {
-        Ok(threads) => threads,
-        Err(e) => {
-            warn!(target: LOG_TARGET, "GPU detection failed: {}", e);
-            0 // Explicit default when GPU/OpenCL is not available
-        }
-    };
+    // At some point we should split this per card and allow the user to choose
+    let gpu_threads = 256;
```
PanchoBubble, line 206: don't you want to call `LUDICROUS_MODE_GPU_GRID_SIZE` here instead?

PanchoBubble (approved):

src-tauri/Cargo.toml
```diff
@@ -30,27 +30,27 @@ keyring = {version = "3.0.5", features = [
...
+] }# Required for tari_wallet
 log = "0.4.22"
 log4rs = "1.3.0"
 minotari_node_grpc_client = {git = "https://github.com/tari-project/tari.git", branch = "development"}
 minotari_wallet_grpc_client = {git = "https://github.com/tari-project/tari.git", branch = "development"}
-nix = {version = "0.29.0", features = ["signal"]}
+nix = {version = "0.29.0", features = ["signal"] }
 nvml-wrapper = "0.10.0"
 open = "5"
 opencl3 = "0.9.5"
```
brianp, line 44: We can drop this.

brianp (changes requested):

PR #1039, opened 2024-11-07
fix: conflicting ports

Misieq01:
### [ Content ]
- Implemented PortAllocator that ensures allocated port is free


mmrrnn (approved):
Tested, works fine

brianp (changes requested):
The PortAllocator doesn't contain any state. Which means we don't need to make it a singleton or use the `current()` pattern. It could just be a namespaced collection of functions. Reducing a tiny bit of memory overhead retaining the singleton, and some complexity. You could still use the struct to house methods if you want, but it could just be called via `PortAllocator::new().get_port()` directly.

PR #1050, opened 2024-11-08
fix(ui): setup screen nav - simplified the progress animation and added a hover style

peps:
This PR simplifies the steps navigation on the setup screen nav. It now animates smoothly from 0-100% over 15 seconds. The trigger to jump to the next step is now triggered onAnimationComplete. This way we don't need to sync up timers with the animation ourselves. 

I also added a hover state and made a few minor CSS tweaks along the way. 

CleanShot 2024-11-08 at 17 37 01

https://www.loom.com/share/7cb2b49c927f40309e4f6faf0a3bfeea?sid=cc844d7b-8b09-4428-b2b4-85419320271b


shanimal08 (approved):
<3

Issue #1046, opened 2024-11-07
As soon as the miner opens there is one single block flickering(blue/red) the jumps off. 


Anziskavds:
**Describe the bug**
Intermittently as soon as the miner opens there is one single block, flickering and blue and red for a few times and then jumps off before the blocks starts building.

**To Reproduce**
1. Open the miner
2. Wait for the miner to open
3. Look at the blocks
4. As soon as the miner opens there is one single block, flickering and blue and red for a few times and then jumps off before the blocks starts building

**Expected behavior**
Block should start building immediately after start up.

**Video**
https://www.loom.com/share/1ad7e030187c481da3b3e83cda915ad5?sid=0d1224f8-ee9e-46a1-85b3-54fb8a2dcbfc

**Desktop (please complete the following information):**
 - OS & Version: MacBook Air 13 inch M1 2020 on Sequoia v15.1
 - Browser & Version Tsetnet v0.6.1

Tas4tari:
Tested on Windows 11, app version 0.6.2 and saw this occur after a fresh install

Tas4tari:
I've seen this on Windows 11, app version 0.6.4

Tas4tari:
Tested on Mac Intel Core i5, app version 0.6.4 and this occurred even though the update accurred automatically

Tas4tari:
Tested on this on Windows 11, app version 0.6.6 and this still occurs. Verified this with a fresh install

Tas4tari:
Tested successfully on Windows 11, app version 0.6.15

PR #1057, opened 2024-11-12
fix(ui): animation end on startup

shanimal08:
Description
---

the main issue was that the _failed_ block animation has been showing right on startup because it already started the mining animation and then we'd get the mining metrics and move on to a 'new' block (the first one really) and the animation would move on to "fail" state because we'd check against your last tx block height to see if you won it, but it goes to "fail" for two reasons - 1. a brand new user wouldn't have any transactions to check against, and 2. even if it's an existing user with a wallet, the initial mining metrics fetch would be checked against the initial 0 or undefined block height, so it wouldn't be a win 

this PR changes things so 1. we don't call `handleNewBlock` until we're getting the actual block height from the base_node, and 2. we only show the tower and start the mining animation at the same time as setting up is complete and the mining metrics are fetched for the first time 

(this won't prevent it from happening _ever again_ though, as there's a chance you really could lose the latest block right as you've started up and it moves onto solving for the next :p)

also did a bunch of cleanups re. the set up flow 

---

**animation specific:**

- update glApp build with a version that doesn't automatically "showVisual" so we have better control over that on our side
- only set showVisual when setup is complete and we start mining (if auto mining is on, which it is on a completely new install)

**setup flow cleanups/improvements:**

- moved external dependency listener to own hook and call in `AppWrapper`
- fixed AutoUpdate setState error (split out listener and update logic, also moved listeners to useEffects)
- added check for if miner metrics is currently being fetched or not
- cleaned up `useSetup` hook and removed old deps form the listener dependency array that are no longer needed

Motivation and Context
---
- resolves https://github.com/tari-project/universe/issues/1046
- resolves the auto update dialog console error i kept seeing

How Has This Been Tested?
---
- locally, with a clean install: 

https://github.com/user-attachments/assets/549410dc-b13e-43c5-9f9b-aa714a7a94dc

---
- if you're running this locally to test, please remove caches, application support etc folders so you have to wait for headers to sync and also have nothing in your wallet:D (that's how you can reproduce on current `main`) - see that is doesn't happen
- if testing on a built version, be sure to reset settings and wallet before (if you had a balance, else it should be fine to just run and test)

Breaking Changes
---

- [x] None


Tas4tari:
Tested successfully on Windows 11, app version 0.6.15

Issue #821, opened 2024-10-11
0.5.18: Windows: When toggling Pool Mining on/off and the app is restarted, I see flashing black screens 


Tas4tari:
**Describe the bug**
As a user I find that when I toggle Pool Mining off or back on and the app restarts, there is a black screen that flashes for less than aa second. This happens until the app is opened and the miner starts

**To Reproduce**
1. Go to Settings
2. Select Experimental settings
3. Go to Pool Mining
4. Turn the toggle off
5. Click on `Restart` CTA on the pop up that appears
6. Observe as the app restarts

**Expected behavior**
The app should restart with no issue, and there should be no window/screen that flashes during this process

**Video**
https://www.loom.com/share/bdf4f479a87b4d27acd3d4477542b651?sid=16f53af8-50ca-4093-a0e5-b9361647803a

**logs:**
[logs.zip](https://github.com/user-attachments/files/17341104/logs.zip)

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11, version 23H2

Tas4tari:
This is still occurring on NextNet version 0.5.20

Tas4tari:
On nextNet version 0.5.22, I find that this is moe persisteent now. wheneever I'm requested to resstart the app and confirm, I see the flashing black screens

Tas4tari:
This is still an issue on Release version 0.5.24

Tas4tari:
Tested on Nextnet version 0.5.25 on Windows 11 and this is still an issue. Especially when I closee of the aapp and restart it

Tas4tari:
Tested on Nextnet version 0.5.27 on Windows 11 and this is still an issue.

Tas4tari:
Tested on v0.5.28, isssue is still occurring

mmrrnn:
There must be "creation_flags(PROCESS_CREATION_NO_WINDOW)" missing somewhere

Tas4tari:
Testeed on Windows 11, app version 0.5.56 and this issue is still occurring

Tas4tari:
Tested on Windows 11, app version 0.5.58 and this is still occurring

Tas4tari:
Tested on Windows 11, app version 0.5.60 and issue is still occurring

Tas4tari:
Tested on Windows 11 on app version 0.5.61 and this is still an issu

Tas4tari:
Tested on Windows 11 on app version 0.5.63 and this is still an issue

[...]

PR #1058, opened 2024-11-12
fix(ui): prevent from displaying cli window

mmrrnn:
#821

Prevent the command line window from appearing on Windows when killing previous instances and installing dependencies. I haven't utilized taskkill because it doesn't allow for completely suppressing the popup window.

### Code cleanup:

* [`src-tauri/src/xmrig_adapter.rs`](diffhunk://#diff-f23e650ace5cbb6e4f0795f97caf21270734697fde4e31a45a650c14088ca1f0L69-L70): Removed the duplicated call to `kill_previous_instances`. It's already callled by `process_watcher`


Tas4tari:
Tested successfully on Windows 11, app version 0.7.0 and I no longer see any flashing screens

Tas4tari:
Tested successfully on Windows 11, app version 0.7.0

PR #1060, opened 2024-11-12
fix(restart): optional stopping miners for restart application

mmrrnn:
Added a flag for terminating processes before restarting the app. We don't want to kill processes during auto-updates, but in other cases we want to kill them and prevent from hanging processes and killing them after the restart.

### DIFF
--- a/src/components/elements/LinearProgress.tsx
+++ b/src/components/elements/LinearProgress.tsx
@@ -41,6 +41,7 @@ const Bar = styled(m.div)<{ $variant?: 'primary' | 'small' | 'tiny' }>`
     border-radius: 50px;
     background: ${({ theme }) => theme.palette.contrast};
     height: ${({ $variant }) => ($variant ? '5px' : '10px')};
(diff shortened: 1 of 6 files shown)


brianp:
Why don't we want to kill processes during auto-update?

mmrrnn:
App is stopped until user decides whether he wants to update the app or not - in that case we would freeze the app trying to kill miners that hasn't even started yet

Tas4tari:
Tested on Windows 11, app version 0.6.15. I find that when I click on `Restart Now` nothing occurs for a few seconds and gives me the impression that the CTA is not working before the app eventually closes. As a user this experience is not ideal for me as I almost closed the app manually thinking that perhaps it may have frozen etc

Anziskavds:
Tested on MacBook Air 13 inch M1 2020 and I also experience the delay when clicking on the ``Restart now`` button. I've experienced this on Testnet v0.6.16

PR #1062, opened 2024-11-13
feat: cache get_network_difficulty response

mmrrnn:
Cache `get_network_difficulty`  response since it's too heavy to call it every 1sec

### DIFF
--- a/src-tauri/src/node_manager.rs
+++ b/src-tauri/src/node_manager.rs
@@ -171,10 +171,10 @@ impl NodeManager {
     pub async fn get_network_hash_rate_and_block_reward(
         &self,
     ) -> Result<(u64, u64, MicroMinotari, u64, u64, bool), NodeManagerError> {
-        let status_monitor_lock = self.watcher.read().await;
+        let mut status_monitor_lock = self.watcher.write().await;
         let status_monitor = status_monitor_lock
             .status_monitor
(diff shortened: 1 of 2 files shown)


brianp:
Should it be noticeable via the logs that we receive less timeout, or already fetching notifications?

mmrrnn:
I tested it with timer locally and I went from like 30/40ms to 2/3ms -> Not a big deal but maybe in some cases it's more significant. This is called once per second.

src-tauri/src/node_manager.rs
```diff
@@ -171,10 +171,10 @@ impl NodeManager {
     pub async fn get_network_hash_rate_and_block_reward(
         &self,
     ) -> Result<(u64, u64, MicroMinotari, u64, u64, bool), NodeManagerError> {
-        let status_monitor_lock = self.watcher.read().await;
+        let mut status_monitor_lock = self.watcher.write().await;
         let status_monitor = status_monitor_lock
             .status_monitor
-            .as_ref()
+            .as_mut()
```
brianp, line 177: Why did this need to become mutable?

mmrrnn: We save cached properties

brianp (approved):
Tested, no new issues found. Seems to be working.

PR #1063, opened 2024-11-13
chore: taplo for toml

brianp:
Description
---
Enforce good toml files and provide the tooling to do it

Motivation and Context
---
Toml keeps flip flopping based on peoples editor settings


leet4tari (approved):
utACK

mmrrnn (approved):

PR #1064, opened 2024-11-14
feat(ui): setup slides reading speed timer

peps:
This PR adds a function to calculate the reading speed of each slide on the setup screens based on text length. 

https://www.loom.com/share/df378764a9ac4d10b5fe2ab4a85d2616?sid=cdabc628-7b6d-4e0d-81b4-ad1b4746dd43

### DIFF
--- a/src/components/elements/LinearProgress.tsx
+++ b/src/components/elements/LinearProgress.tsx
@@ -41,6 +41,7 @@ const Bar = styled(m.div)<{ $variant?: 'primary' | 'small' | 'tiny' }>`
     border-radius: 50px;
     background: ${({ theme }) => theme.palette.contrast};
     height: ${({ $variant }) => ($variant ? '5px' : '10px')};
+    will-change: width;
 `;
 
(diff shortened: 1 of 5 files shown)


brianp (changes requested):
I enjoy that it's *much* faster than before.

but now I find it just a *smidge* too fast. Like maybe the reading time is just right, but there's no breathing room between start and finish of the next slide where that initial second or two is lost.

peps:
> I enjoy that it's _much_ faster than before.
> 
> but now I find it just a _smidge_ too fast. Like maybe the reading time is just right, but there's no breathing room between start and finish of the next slide where that initial second or two is lost.

Ok, I added +1 more second pause. We can adjust and dial it in more if needed.

Tas4tari:
Tested successfully on Windows 11, app version 0.6.15

Anziskavds:
Tested successfully on MacBook Air 13 inch on M1 2020 on Testnet v0.6.16.

PR #1066, opened 2024-11-14
feat(ui): admin ui menu

peps:
This PR adds a `Admin UI` menu for controlling and debugging various UI elements. Each button in the menu is a toggle so each element can be turned on and off. 

This will allow us to work on various UI elements, and debug them more easily. 

This menu will only be accessible when running the app in dev mode. 

CleanShot 2024-11-14 at 18 12 33


shanimal08 (approved):
neat!! tested and LGTM :D

brianp:
It's worth noting that the button doesn't appear until you hover the corner.

Also the dialogs are "real" in that they will do things like reset, restart, etc. They are not demo dialogs, they're forced real dialogs.

peps:
> It's worth noting that the button doesn't appear until you hover the corner.
> 
> Also the dialogs are "real" in that they will do things like reset, restart, etc. They are not demo dialogs, they're forced real dialogs.

The purpose of this menu is to provide easy access to these UI elements for development and UI testing. More for the visual side of things than the actual functionality. 

The button is hidden until hover on purpose so it doesn't distract visually from the design of the page. It's there for us to use if we need it, so it doesnt need to be visible at all times.

brianp:
Don't get me wrong, I don't disagree. I just said it's worth noting :D 

If you start the app expecting a button and it's not there, someone needs to know how to find it.

PR #1068, opened 2024-11-15
feat(ui): release notes section in settings modal

peps:
This PR adds a Release Notes section to the settings modal. The release notes themselves are stored in a markdown file in the public folder called `CHANGELOG.md`

**package.json update** - I added a new dependancy `react-markdown` to parse the markdown file.

The first version in the list is open by default, the rest of the entries are collapsed and can be toggled open. 

A`upgrade` button is shown if there is a new version of the app available. 

CleanShot 2024-11-18 at 16 48 05

CleanShot 2024-11-25 at 18 40 39

CleanShot 2024-11-25 at 18 51 11


brianp:
That scroll bar will become infinitely long as we continue. Would it be possible to collapse older sections to create a list of release versions and dates that can be expanded?

peps:
> That scroll bar will become infinitely long as we continue. Would it be possible to collapse older sections to create a list of release versions and dates that can be expanded?

Sure, its possible, ill add some custom parsing to make it happen

src/store/useUIStore.ts
```diff
@@ -38,6 +40,7 @@ const initialState: State = {
     dialogToShow: null,
     showExperimental: false,
     showExternalDependenciesDialog: false,
+    enableReleaseNotes: false,
```
peps, line 43: Set this to `true` to enable the section in Settings

public/ReleaseNotes.md
peps, line 1: This file was generated using AI, it needs real content before merging.

shanimal08: also re. this file - i wonder if we shouldn't have translated version of it too? 😬

shanimal08 (approved):
LGTM & tested - looks great:D

public/ReleaseNotes.md
peps: @shanimal08 hm probably... It's not my call, but I personally think we should launch with english first then add multi-lang in a future iteration. Im not sure how complicated it would be to set up this `ReleaseNotes.md` file with auto-translations and don't want that holding this PR up.

brianp:
Additional item talked about today: If we're linking to a remote file it's possible someone on an older version would receive the updated file seeing notes for a version they don't have. 

If the version they are on, is older than that of the notes lets add a big Upgrade button/banner.

Anziskavds:
Tested successfully on a MacBook Pro M1 2020 13 inch Sequoia 15.1.1 on Testnet v0.8.7.

Tas4tari:
Tested successfully on Windows 11, app version 0.8.7
Screenshot (443)
Screenshot (444)
Screenshot (445)
Screenshot (446)
Screenshot (447)

PR #1071, opened 2024-11-15
fix(ui): add animation build sans long block animation

shanimal08:
animation build without the error block holding things up 
thx to @mmrrnn for finding the issue in glApp

the issue was the error (looong block) block lifecycle messing with the animation state manager moving on, so we just removed the 'errorBlock' logic for now

- if you get to 20 of the blue blocks that's when an error block would've come up and messed everything up, so if you try stop mining now after 20 blue blocks it should actually stop


brianp (approved):
Tested and block animation stopped when mining stopped.

Anziskavds:
Tested successfully on Testnet v0.6.17 on a MacBook Air 13 inch M1 2020.

PR #1075, opened 2024-11-15
fix(ui): loading indicator for restart dialog

mmrrnn:
Restart application takes 5+ secs because we have to kill all related processes. Without loading indicator it looked like restarting does not work at all.
Screencast from 11-15-2024 05_15_43 PM


Tas4tari:
Tested successfully on Windows 11, app version 0.7.0

Anziskavds:
Tested successfully on MacBook Air 13 inch M1 2020 on Testnet v0.7.0.

PR #1090, opened 2024-11-20
feat(ui): reward replay

shanimal08:
Description
---
- config updates:
  - ~~added replayed items to app config (since the requirements only allow for one replay per app install)~~
  - removed old config flags we don't need any more 
- UI updates:
  - added replay button to history item and missed item counter to wallet button
  - added individual replay text to `Earnings.tsx` for the animatin
- logic updates:
  - add check to `useEarningsRecap` to set reward counter
  - added `handleWinReplay` action to `useBlockchainVisualisationStore` to handle individual replay animations (whether mining or not)

Motivation and Context
---
- https://tari.atlassian.net/browse/TC-47

How Has This Been Tested?
---
- locally:

https://github.com/user-attachments/assets/fa704cb0-8965-406c-a2e4-fde2dc472c97

(when not mining animation state has to be started, then stopped after)

https://github.com/user-attachments/assets/3c008eb1-6dc8-40f5-b1d4-f8cb9c5e1108

---
- open wallet and replay any history item animation
- if you leave it mining for a while unfocused/minimised and come back with recaps, there should be a notification with a counter of the wind you missed.
- opening the wallet to go replay should clear the counter

Breaking Changes
---

- [x] None


src/store/useAppConfigStore.ts
```diff
@@ -41,7 +44,7 @@ const initialState: State = {
     config_version: 0,
     config_file: undefined,
     mode: 'Eco',
-    auto_mining: true,
```
brianp, line 44: Was this the feature gate for `mine_on_app_start`?

brianp (commented):
I don't love storing the id's in the app config. As it's not config. Do we have any other storage options that aren't local storage, and also not config?

brianp:
This needs a "clear all" button. If we're adding a notification style red number. I personally don't want a notification left there, and also don't want to have to click and watch every single replay.

src/store/useAppConfigStore.ts
shanimal08: nope that'd be [`mine_on_app_start`](https://github.com/tari-project/universe/pull/1090/files#diff-cd1133f102d8444a5d366fbd1e7d5878e3908ba7f1d41ea778a6893426af8a63R48) `auto_mining` is the flag from aaages ago

shanimal08:
> This needs a "clear all" button. If we're adding a notification style red number. I personally don't want a notification left there, and also don't want to have to click and watch every single replay.

^ @metalaureate pls advise:D

brianp:
If you are away, and see the notification that you have 1 or 2 rewards on played. But since being back have won additional rewards. There is no way to know which reward is the one that can be played / is new.

Tas4tari:
Tested successfully on Windows 11, app version 0.7.1:

- open wallet and replay any history item animation

Unable to test the following as I have not yet won any tXTM to confirm notifications and replay:

- if you leave it mining for a while unfocused/minimised and come back with recaps, there should be a notification with a counter of the wind you missed.
- opening the wallet to go replay should clear the counter

PR #1097, opened 2024-11-21
fix(ui): history item

shanimal08:
Description
---
- moved the parent hover style reference and added to the right side items too
- noticed while testing the `showShareButton` check meant the replay button would also be hidden if you weren't logged in with airdrop so fixed up those checks too

Motivation and Context
---

| half of the history item styling wasn't updated on hover |
| :---: |
| image |

How Has This Been Tested?
---

- locally:

https://github.com/user-attachments/assets/e9a39322-addd-4c2f-ba04-eb06816bcf34

https://github.com/user-attachments/assets/fbf12ee8-2f34-41bf-be67-965e7dabd52e

https://github.com/user-attachments/assets/923a4a46-34ec-43ed-bf8e-17599258f612

**(only to show animation change:)**

https://github.com/user-attachments/assets/17c61044-4fb4-4750-af68-23aeb6e3589a

---

- check the history items when logged in and out with airdrop
- make sure the history item styling isn't peeking through anywhere on hover:3

Breaking Changes
---

- [x] None

### DIFF
--- a/src/containers/main/SideBar/components/Wallet/HistoryItem.styles.ts
+++ b/src/containers/main/SideBar/components/Wallet/HistoryItem.styles.ts
@@ -94,18 +94,12 @@ export const HoverWrapper = styled(m.div)`
     position: absolute;
     inset: 0;
     z-index: 4;
-    background-color: rgba(255, 255, 255, 0.1);
-    align-items: center;
-    display: flex;
-    flex-direction: row;
-    justify-content: flex-end;
-    height: 100%;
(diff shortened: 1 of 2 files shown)


peps:
Nice, one little thing, it's weird that the share button animates in on the `x` and the history button doesn't. 

To make it less distracting, I recommend removing the `x` animation on the share button and have them both fade in at the same speed on hover.

peps (approved):

PR #1098, opened 2024-11-21
fix(ui): bring fonts inline to optimize loading and remove more unused fonts

peps:
This PR is an effort to improve animation and font performance, especially on windows. 

- Load fonts inline in the `index.html` head instead of from `styled-components`
- Remove all unused fonts
- Remove GTAmerica font since it was only used in 1 place ( the history items ). Poppins looks fine here as well.


shanimal08 (approved):
LGTM, d'ya wanna just delete the stuff in `theme/fonts` too though :D

peps:
> LGTM, d'ya wanna just delete the stuff in `theme/fonts` too though :D

done

brianp (approved):
Game changer on windows.

Issue #851, opened 2024-10-15
Release version: 0.5.22: WebGL errors still prevalent on Sentry


Tas4tari:
**Describe the bug**
When checking Sentry, I founnd that there are still WebGL issues that are showing up:
https://tari-labs.sentry.io/issues/?end=2024-10-15T11%3A31%3A59&environment=production&groupStatsPeriod=auto&project=4507979991285760&query=release%3Atari-universe%400.5.22&sort=freq&start=2024-10-14T17%3A06%3A00
**Screenshot:**
<img width="1552" alt="Screenshot 2024-10-15 at 1 40 25 PM" src="https://github.com/user-attachments/assets/bcae6ec1-c595-4fec-b3a7-194033cbbd87">

mmrrnn:
![Image](https://github.com/user-attachments/assets/53ec061a-c878-4823-8cd2-aae8531ccfc2)


```
 2024-11-07 12:38:00.624873601 ERROR "THREE.WebGLRenderer: Error creating WebGL context."
 2024-11-07 12:38:01.226404014 INFO  "Updater event" "Log Item" "UPTODATE"
 2024-11-07 12:38:01.422842358 INFO  "Updater event" "Log Item" "UPTODATE"
 2024-11-07 12:39:10.836055560 INFO  "Mining starting...."
 2024-11-07 12:39:10.836213986 INFO  "Fetching applications versions"
 2024-11-07 12:39:10.836265539 INFO  "Fetching applications versions"
 2024-11-07 12:39:10.850584119 INFO  "Mining started."
```

all the time on my appImage build on ubuntu. It works fine on dev and manual build

mmrrnn:
LINUX
--
During testing, I have noticed that there is a difference in behavior between my local build and the public build in terms of losing the WebGL context. This issue seems to be related to the fact that the local build was created on different hardware than the public build, causing compatibility issues with the WebGL context.

brianp:
Does this context error actually prevent the visuals from rendering?

If so, is there any way we can check for disabled hardware acceleration then
- not initialize the visuals
- disable visual mode by default, and not allow it to be enabled

PR #1102, opened 2024-11-22
fix: webgl not supported

mmrrnn:
[#851](https://github.com/tari-project/universe/issues/851)

When WebGL not supported, disable visual mode, display snackbar and send info to the sentry
Screenshot from 2024-11-22 00-20-42
Screenshot from 2024-11-22 00-24-31
Screenshot from 2024-11-22 00-22-33


src/App/App.tsx
```diff
@@ -15,6 +19,19 @@ import AppContent from './AppContent';
 export default function App() {
     const isShuttingDown = useShuttingDown();
     const isSettingUp = useAppStateStore((s) => s.isSettingUp);
+    const setError = useAppStateStore((s) => s.setError);
+    const setIsWebglNotSupported = useUIStore((s) => s.setIsWebglNotSupported);
+    const { t } = useTranslation('common', { useSuspense: false });
+
+    useEffect(() => {
+        if (!window.WebGL2RenderingContext && !window.WebGLRenderingContext) {
+            Sentry.captureMessage('WebGL not supported by the browser', { extra: { userAgent: navigator.userAgent } });
```
brianp, line 28: > and send info to the sentry

I think if we've identified WebGL is not supported, we've prevented the context error, there's no need to send any info to sentry. It's just noise at that point. Maybe best to log it, so it's recorded somewhere local to the instance.

brianp (changes requested):

src/App/App.tsx
mmrrnn: Error won't be sent to sentry since I prevented from animation init. Decided do send messages to sentry to check what browsers users use when it happens

mmrrnn: So I can maybe improve this

Tas4tari:
Checked on Sentry and did not see any webgl errors for app version 0.7.7:
https://tari-labs.sentry.io/issues/?cursor=1%3A10%3A0&end=2024-11-26T08%3A51%3A59&environment=production&groupStatsPeriod=auto&page=1&project=4507979991285760&query=release%3Atari-universe%400.7.7&referrer=issue-list&sort=freq&start=2024-11-24T10%3A25%3A00

PR #1108, opened 2024-11-22
chore: dedupe sentry logging | ignore already getting x warns | fix lints

shanimal08:
Description
---

noticed the actual error from BE wasn't logged in sentry re. paper_wallet details, then noticed the amount of duplications

- removed `Sentry.captureException(`s in each `catch` because we are already using the `captureConsoleIntegration` 
- added ignore for all the `"Already getting ...."` warnings from BE which are being logged as errors clogging up sentry
- made sure all `console.error`s have a short description and are passing the error through
- fixed lints

### DIFF
--- a/src/App/AppWrapper.tsx
+++ b/src/App/AppWrapper.tsx
@@ -1,3 +1,4 @@
+import { IGNORE_FETCHING } from '@app/App/sentryIgnore';
 import { useDisableRefresh } from '@app/hooks/useDisableRefresh';
 import { useListenForExternalDependencies } from '@app/hooks/useListenForExternalDependencies';
 import { useUpdateListener } from '@app/hooks/useUpdateStatus';
@@ -16,7 +17,7 @@ import App from './App.tsx';
 const environment = import.meta.env.MODE;
 const sentryOptions = {
     dsn: 'https://edd6b9c1494eb7fda6ee45590b80bcee@o4504839079002112.ingest.us.sentry.io/4507979991285760',
-    integrations: [Sentry.captureConsoleIntegration({ levels: ['warn', 'error'] })],
+    integrations: [Sentry.captureConsoleIntegration({ levels: ['warn', 'error'] }), Sentry.extraErrorDataIntegration()],
     release: packageInfo.version,
     environment,
     // Set tracesSampleRate to 1.0 to capture 100%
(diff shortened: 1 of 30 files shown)


src/App/AppWrapper.tsx
```diff
@@ -16,14 +17,15 @@ import App from './App.tsx';
 const environment = import.meta.env.MODE;
 const sentryOptions = {
     dsn: 'https://edd6b9c1494eb7fda6ee45590b80bcee@o4504839079002112.ingest.us.sentry.io/4507979991285760',
-    integrations: [Sentry.captureConsoleIntegration({ levels: ['warn', 'error'] })],
+    integrations: [Sentry.captureConsoleIntegration({ levels: ['warn', 'error'] }), Sentry.extraErrorDataIntegration()],
```
brianp, line 20: Couldn't we just turn this down to `'error'` and remove `'warn'` and then Sentry wouldn't get warnings, or am I oversimplifying it?

shanimal08: we don't actually have that many `console.warn`s, none we've added specifically, usually they come from the webview console and that's where a couple of the webgl warnings come from so figured it'd be good to leave in... 

when i said here 

> added ignore for all the "Already getting ...." warnings from BE which are being logged as errors clogging up sentry

the "warn"ings are warns from BE which come to the FE as errors in the catch, separate from `console.warn()` warnings 😅 

those three warnings/errors are the main ones clogging up sentry, and i don't think we need them logged really... certainly not in sentry (but didn't want to remove the source logs entirely:D)

brianp: hmm, I guess i'm okay with it for now. Would be nice if the BE and FE could negotiate the difference between a warning and an error by type instead of by message but that can be a future enhancement.

shanimal08: think the tauri v2 sentry plugin  will help a bunch;)

Tas4tari:
Checked sentry : Open Issues for v0.7.7: 
https://tari-labs.sentry.io/issues/?cursor=1%3A10%3A0&end=2024-11-26T06%3A56%3A59&environment=production&groupStatsPeriod=auto&page=1&project=4507979991285760&query=release%3Atari-universe%400.7.7&referrer=issue-list&sort=freq&start=2024-11-24T10%3A25%3A00
And confirmed that I could not see any `Already getting miner metrics` messages

PR #1110, opened 2024-11-22
fix: app_config overrides

mmrrnn:
Description
---
We've found with @Misieq01 what's causing the `app_config` overrides. It's because older versions have: `custom_max_gpu_usage: Option<u32>`(which is `null` as default). It was later changed to `custom_max_gpu_usage: Vec<GpuThreads>`(which is `[]` by default). It was treated as types inconsistency by serde which led to using and saving default app config

### DIFF
--- a/src-tauri/src/app_config.rs
+++ b/src-tauri/src/app_config.rs
@@ -52,8 +52,11 @@ pub struct AppConfigFromFile {
     reset_earnings: bool,
     eco_mode_cpu_threads: Option<u32>,
     ludicrous_mode_cpu_threads: Option<u32>,
+    #[serde(default = "default_vec_string")]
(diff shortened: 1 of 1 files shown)


MCozhusheck (approved):

brianp:
As a testing note: This PR doesn't address the problem of configs being reset to default. A config created with 0.7.1 will get *entirely* reset when upgrading even with this fix.

If I set my monero address in 0.7.1, and then update to this branch. My monero address will be reset back to default, as the old config is still considered corrupt. Users will still suffer the issue on all versions before 0.7.3.

Tas4tari:
Tested successfully on Windows, app version 0.7.7 by checking app config which shows as:
`custom_max_gpu_usage:[]``

Issue #1105, opened 2024-11-22
Re-enable long block animation


brianp:
# Description

We disabled the long block animation as it was conflicting with multiple other animations, and causing weird states.

Work out the strange state issues with other animations, and re-enable the long block.

Related to https://github.com/tari-project/universe/pull/1071

PR #1113, opened 2024-11-22
fix: fixed animation build | add "long block" back & fix queue

shanimal08:
Description
---

adds fixed animation build

**long block changes:**

- added "long block" animation back
- added reset to error block on status change so the queue doesn't get clogged up waiting for the long block animation lifecycle
- updated `useMiningUiStateMachine` so we don't try to start or stop again if the statusIndex changes - made the checks stricter

**setup/main/shutdown screens animation changes:**

  - hide canvas on shut down
  - sped up durations
  - removed `AppContent` wrapper and just used `AnimatePresence` (no "wait" mode) so with `AppContentContainer` wrappers as direct children so there's no weird error about keys on shut down
  - add ui field for admin UI toggle of show setup (so we don't set up again)

**other changes:**

- check if macOS before trying to fetch Tor entry guards (saw the error locally and plenty in sentry) + small style tweak

Motivation and Context
---

- previous issue was that the animation either wouldn't stop if you stopped mining (most noticeable one), or the other states like fail or success might also not come up at the correct time after a "long block" had been added to the queue.
- resolves https://github.com/tari-project/universe/issues/1105

How Has This Been Tested?
---
- locally:

https://github.com/user-attachments/assets/5239b4a1-6554-4ae8-a697-a0cfa81ca687


mmrrnn (commented):
Tested successfully

mmrrnn (approved):

brianp (changes requested):
I did see the long block animation. After winning my first block the win animation played, then the green row "jumped up" a level, and now the green row won't go away.

image

I restarted, and successfully mined a block before the long block animation. The green row cleared as expected.

I continued mining, receiving the long block, and then winning, and got stuck with a green platform again.

brianp (changes requested):
Same replication steps as before. Green platform doesn't clear, and now a block is missing.

image

shanimal08:
i wasn't ready yet @brianp haven't pushed the missing block fix 😬

shanimal08:
https://github.com/user-attachments/assets/695e044d-5357-46d3-b585-2269e786ba0d

oki ready now @brianp :D

brianp:
Sorry @shanimal08 

The issue now seems intermittent. Or the circumstances have become even more specific. If I've won any blocks before a case where long block plays, and then I win, I don't see this problem.

But almost anytime I restart the app, and get long block, followed by winning that block it freezes.

This time I also captured the platform "jump" which might be a clue as to what the problem is.

https://github.com/user-attachments/assets/740d0f89-be27-44b2-b9d6-aea047ee0e52

brianp:
[universe-web.log](https://github.com/user-attachments/files/17919929/universe-web.log)

brianp (approved):
I have not reproduced the issue any longer. Seems like it could be fixed or it's your lucky afternoon 👀

Tas4tari:
Tested successfully thus far on Windows 11, app version 0.8.2 however I only saw 2 long block animations and I could stop mining successfully. I will be testing this further as I go

Issue #856, opened 2024-10-15
Release version 0.5.22: Wallet Settings: When user clicks on copy for Monero address they are prompted to restart the app


Tas4tari:
**Describe the bug**
As a user I find that when I click on the `copy` icon for the Monero address, the `Restart Tari Universe?` pop up appears. It makes no sense to have it on the Copy icon

**A few other issues:**
- It is not intuitive to the user that they are able to edit their Monero addess as there is no edit icon
- When I remove my Monero address and leave it blank, and click cancel within the `Restart Tari Universe?` pop up, I find that it does not re-populate the input with the address that was there prior to me removing it. 
  - As a user I was expecting the  change to not be retained as I selected to cancel the pop up instead of restarting to activate the change

**Loom:**
https://www.loom.com/share/54bc8d2630cb40b18cd46fc94819aec7?sid=b5d87a71-70a4-4e99-a06e-2526a31cbd0a

** Device:**
- Acer 315-59, version 23H2, 12th Gen intel core i7
- Windows 11

Tas4tari:
Tested on Windows 11 on NextNet version 0.5.27 and issue is still occurring

Tas4tari:
Tested on Windows 11, version 0.5.28 and this is still an issue

Anziskavds:
This is happening on MacBook Air 13 inch M1 2020 on Sonoma 14.5 on Testnet v0.5.48.
<img width="1370" alt="image" src="https://github.com/user-attachments/assets/f038c18a-ae77-415d-aeb4-605b5d60243f">

Tas4tari:
Tested on Windows 11, app version 0.5.56 and this is still an issue

Tas4tari:
Tested on app version 0.5.58 on Windows 11 and this is still occurring

Tas4tari:
Tested on Windows 11 on app version 0.5.60 and this is still an issue

Tas4tari:
Tested on Windows 11 on app version 0.5.63 and this is still an issue

Tas4tari:
Tested on Windows 11 on app version 0.6.0 and this is still an issue

Tas4tari:
Tested on Windows 11 on app version 0.6.2 and this is still an issue

Tas4tari:
Tested on Windows 11 on app version 0.6.4 and this is still an issue

Tas4tari:
Tested on Windows 11 on app version 0.6.6 and this is still an issue

[...]

PR #1116, opened 2024-11-25
fix: prevent reset app dialog after copying monero address

MCozhusheck:
Description
---
Prevent reset app dialog popping up when user copies monero address
Closes: #856 

Motivation and Context
---
We should show this dialog only when user changes monero address.

How Has This Been Tested?
---
Run on ubuntu and windows, repeatedly clicked monero address multiple times and the dialog did not show up.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


shanimal08 (approved):

Tas4tari:
Tested successfully on Windows 11, app version 0.7.8

PR #1117, opened 2024-11-25
fix: change only minotari node to stop on error 114

stringhandler:
This exit was added so that we can detect when the base node stops on error 114 so that we can delete the previous database and resync. However, the GPU miner would exit with a weird error code and then it would not be restarted.


brianp (approved):
LGTM

Tas4tari:
Tested on Windows 11, by going to task manager, ending minotari processes and confirming it restarts immediately on app version 0.8.2 
If I tested this incorrectly please let me know

Issue #819, opened 2024-10-11
0.5.18: Windows: Starting up the program and checking if it connects with tor enabled and then with tcp enabled is not working


Tas4tari:
**Describe the bug**
As a user I find when I install the app, disable Tor, close the app, delete the node file and restart, I get stuck at `connecting network peers` for the past 35+ minutes. 
**Additional info:**
I checked on com.tari.universe and the node file had been addded again

**To Reproduce**
Test Starting up the program and checking if it connects with tor enabled and then with tcp enabled:

1. User should have Tari Universe installed on their computer
2. User should have Tari Univers open
3. Click on the Settings icon
4. Go to Experimental Settings
5. Scroll down to the Tor toggle
6. User should see the Use Tor toggle on by default
7. User should be able to toggle the Use Tor toggle on/off
8. When Tor is toggled off, confirm the app asks the user to restart
9. User should say no and close the app
10. Go to com.tari.universe and delete the node data folder: Windows: AppData\Local\node & Mac: Application Support\node
11. User should restart the app
12. Confirm the user is connected to the Tari network and the chain resyncs

**Expected behavior**
On restart of the app, user should be be connected to the Tari network, the chain resynces and the user should be able to mine.

**Screenshots**
![Screenshot (53)](https://github.com/user-attachments/assets/219bdb91-6f04-4c33-929b-add59d1abb32)

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11, version 23H2

Tas4tari:
This is still an issue on 0.5.22

Tas4tari:
This is still an issue on 0.5.24

Tas4tari:
Tested on NextNet: version: 0.5.25 and this is still an issue

Tas4tari:
Tested on NextNet version 0.5.27 and this is still an issue

Tas4tari:
Tested on v0.5.28 and issue is still prevalent

Anziskavds:
When a user toggled Tor off and clicked cancel on the restart pop up and manually restarted the miner, on start up the user received a critical error pop up message. I clicked on the close Tari universe button and tried to open Tari Universe again but I got stuck on 43%. I needed to force quite Tari Universe and set Tor to true on visual studio to get Tari Universe up and running again. This is on a MacBook Air 13 inch M1 2020 on Testnet v0.5.28

mmrrnn:
Related to #874

Anziskavds:
[...]

PR #1118, opened 2024-11-25
fix: when using tcp, use tcp seeds

brianp:
Description
---
When not using tor, use the ip4, and ip6 DNS seed addresses.
Additionally, when the method of connection is changed clear out the peer databases to ensure no wasted efforts on connection types we don't support.

Fixes: #819 

Motivation and Context
---
Currently, our dns entry for seed nodes is overpopulated and can't be read by some ISP's due to the size of the txt records and timeouts.

This makes it nearly impossible to put seeds in for tor, ip4, and ip6 with enough for all. We have now separated out the addresses.
`seeds.esmeralda.tari.com` - with a mixture
`tor.seeds.esmeralda.tari.com` - with only tor addresses
`ip4.seeds.esmeralda.tari.com` - with only ip4 addresses
`ip6.seeds.esmeralda.tari.com` - with only ip6 addresses

When not connecting through tor, use the correct DNS address for seeds.

How Has This Been Tested?
---
Manually

Testing Instructions
---
- Start the application
- Turn off TOR
- See that you can still connect to the Tari network on restart


brianp:
Putting the issue on hold until Devops can propagate the changes to nextnet DNS. /cc @leet4tari

Misieq01 (approved):
Looks good to me

Issue #1121, opened 2024-11-26
Custom mining sliders go transparent while saving when mining


brianp:
**Describe the bug**
The circles denoting where the custom slider selection is set to transparent while saving while the miner is running.

**To Reproduce**
1. Turn the miner on Ludicrous (machine must be mining for it to occur)
2. Go to Custom mining settings
3. Change the settings
4. Select "Save Changes"
5. See the circles go transparent while applying the save

**Expected behavior**
Circles should not go transparent

**Screenshots**
![image](https://github.com/user-attachments/assets/0980919b-8bf5-4948-a050-fb79f4e610de)

https://github.com/user-attachments/assets/a5d8d7f2-9555-4c19-bcaf-9bcf3672c56c

**Desktop (please complete the following information):**
 - Mac OS & Windows

Tas4tari:
Tested successfully on Windows 11, app version 0.8.2

PR #1124, opened 2024-11-26
fix(ui): slider disabled styling

shanimal08:
Description
---

- fixed slider's thumb styling so that isn't all a lighter opacity when `disabled` (`isLoading`) - only lighten the border
- also added a helper to get a "safe" value based on the range slider steps, because i noticed the GPU threads initial value was an odd number, and i could slide to an odd number,  but that should only increment in steps of two

Motivation and Context
---

- resolves https://github.com/tari-project/universe/issues/1121

How Has This Been Tested?
---

- locally:

https://github.com/user-attachments/assets/e40c6653-d74e-4083-a3e5-5592e576800b

---
1. see ticket
2. see that GPU settings can't set to an odd number


Tas4tari:
Tested successfully on Windows 11, app version 0.8.2

Issue #1123, opened 2024-11-26
Improve your Experience toggle is always displayed


brianp:
**Describe the bug**
The analytics toggle to "Improve your Experience" is always displayed. 

This should only be presented on initial startup, and then can be found in the settings if the user would like to switch it off.

<img width="537" alt="image" src="https://github.com/user-attachments/assets/10d9fb9f-f339-41f3-b49f-395c5dbf3dce">

Issue #1114, opened 2024-11-24
Add app anon id under advanced settings


metalaureate:
**Describe the bug**
Currently, users who need their Miner Anon ID for community projects (e.g., Discord bots tracking on-chain activity) must locate this information in app_config. To improve accessibility and support third-party integrations, we should expose this ID in Advanced Settings.
Rationale:

At least one community project has already requested this feature
We anticipate more community projects will need this information
The ID is frequently needed for technical support

brianp:
We will add a read-only and copyable display of the anon id to the General settings pane.

PR #1128, opened 2024-11-26
fix(ui): permissions on intro & updated settings with anon id

shanimal08:
Description
---
- add `created_at` field to app_config to check if it's the first ever run of the app or not (to hide permissions toggle on the Intro screens on subsequent runs)
- added Application Information settings section in General Settings to display the `anon_id` which can be copied
- tightened up General Settings styling a little

Motivation and Context
---
resolves:
- https://github.com/tari-project/universe/issues/1123
- https://github.com/tari-project/universe/issues/1114

How Has This Been Tested?
---

- locally:

https://github.com/user-attachments/assets/410b1168-d926-4fde-b8a2-3aa346fed434

---

re. #1123:

- reset everything/remove app_config and start app - should show permissions toggle on the intro screens
- on any runs after that (not after reset) it shouldn't show on intro screens

re. #1114:

- your app anon_id should be viewable and copy-able(?) from general settings


src/containers/floating/Settings/sections/general/AppDataSettings.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return (
+        <>
+            <SettingsGroupTitle>
+                <Typography variant="h6">{t('application-info')}</Typography>
+            </SettingsGroupTitle>
+            {anon_id && (
+                <SettingsGroupContent>
+                    <Stack direction="row" alignItems="center" justifyContent="flex-start">
+                        {/* eslint-disable-next-line i18next/no-literal-string */}
+                        <Typography variant="p">Anon ID: {anon_id}</Typography>
```
shanimal08, line 24: disabled string check because anon_id is a field name we don't translate, but LMK if that should also be translated

PR #1133, opened 2024-11-27
feat: solve long shutdowns

### LINKED ISSUE #980: Quitting the app before it has finished sync'ing hangs the app on Mac
**Describe the bug**
Wait for 30% then quit--app hangs.

### LINKED ISSUE #1119: closing laptop for some hours and then reopening puts Universe in bad state
**Describe the bug**
Close laptop, wait a few hours. Open--Universe won't stop mining. Closing the app hangs, have to force close.

Misieq01:
Fixes #980 
Fixes #1119  (probably)

### [ Summary ]
- Adjusted `stop_all_miners` methods
- Prevent starting up processes when shutdown is triggered
- exposed is_running methods for all binaries
- add is_pid_file exists metods for all binaries
- fix shutting down screen display


brianp:
I am going to merge this. But I think we might have lost the "Shutting down..." screen when I merged main. I see it briefly but it goes away quite quickly even when the window stays open. So if you can take a look, and see if that's fixable and open another PR it would be appreciated.

PR #1135, opened 2024-11-27
Socket changes

PanchoBubble:


src/hooks/airdrop/useWebsocket.ts
```diff
@@ -7,6 +7,8 @@ import { invoke } from '@tauri-apps/api/tauri';
 
 let socket: ReturnType<typeof io> | null;
 
+const NETWORK = 'esmeralda';
```
brianp, line 10: That's not going to work on nextnet

PanchoBubble: Hey yeah this is for testing purposes won't go live yet

PR #1150, opened 2024-11-28
chore: split out tauri commands for readability

shanimal08:
moved all our tauri commands to their own file 

1. for readability and
2. because i did that in the V2 branch and i think i've missed changes when merging when there are conflicts 😬


mmrrnn (approved):
Looks fine, tested on dev

mmrrnn (approved):

src-tauri/src/commands.rs
```diff
@@ -0,0 +1,1511 @@
...
+use crate::systemtray_manager::{SystemtrayManager, SystrayData};
+use crate::tor_adapter::TorConfig;
+use crate::utils::shutdown_utils::stop_all_processes;
+use crate::wallet_adapter::TransactionInfo;
+use crate::wallet_manager::WalletManagerError;
+use crate::{
+    setup_inner, ApplicationsVersions, BaseNodeStatus, CpuMinerMetrics, GpuMinerMetrics,
+    MaxUsageLevels, MinerMetrics, TariWalletDetails, UniverseAppState, APPLICATION_FOLDER_ID,
+    MAX_ACCEPTABLE_COMMAND_TIME,
+};
```
MCozhusheck, line 24: Remove those imports and paste these struct directly in this file. In a case of `ApplicationsVersions` it's declared in `main.rs` but used only in `commands.rs`. It can be simplified by simply declaring it here.

MCozhusheck (approved):

PR #1159, opened 2024-11-29
Adds referrals request data

PanchoBubble:


src/hooks/airdrop/useWebsocket.ts
```diff
@@ -70,19 +68,7 @@ export const useWebsocket = () => {
...
-                    }
-                    if (
-                        (msgParsed.name === WebsocketEventNames.MINING_STATUS_CREW_UPDATE ||
-                            msgParsed.name === WebsocketEventNames.MINING_STATUS_USER_UPDATE) &&
-                        msgParsed?.data?.totalTimeBonusMs
-                    ) {
-                        setTotalBonusTimeMs(msgParsed.data.totalTimeBonusMs);
-                    }
-                });
+                socket.on(userId as string, handleWsUserIdEvent);
```
PanchoBubble, line 71: Move listener to a handler to keep in cleaner

src/types/ws.ts
```diff
@@ -19,10 +20,19 @@ export interface QuestCompletedEvent {
     };
 }
 
+export interface CrewMember {
+    imageUrl: string | null;
```
BalazsSevecsek, line 24: See the comment at the other PR: many of these information do not change over time, so it would be worth asking them down only once and not every 15 secs

PR #1190, opened 2024-12-05
feat: add ws verification

BalazsSevecsek:


stringhandler (changes requested):
I think this is not the best way of doing it. Let's chat about it offline

PR #1221, opened 2024-12-10
feat: increase autostartup permissions

### LINKED ISSUE #1154: Windows: 0.8.2: Auto start on system boot: When user restarts laptop, app does not automatically start
**Describe the bug**
As a user I find that when I change the setting for `Auto start on system boot` to enabled and restart my laptop. I find that the app does not restart

**To Reproduce**
1. Go to 'Tari Universe' app
2. Click on 'Settings'
3. Scroll down to 'Auto start on system boot'
4. Click on the toggle to enable the setting
5. Setting should be enabled
6. Close the app
7. Restart your laptop
8. Observe

**Expected behavior**
When laptop is restarted, and use unlocks, user should see the Tari Universe app start automatically

**Screenshots**
Screenshot (412)

 - Aspire A315-59
 - Windows 11, 23H2

Misieq01:
Fixes #1154 

### [ Summary ]
- Extended autolauncher strategy for windows with extra step of creating task schedule
- Created task in scheduler is responsible for starting universe app with highest permission when user login first time after booting up machine

--------------

### [ Testing ]
- Build the app from this PR and then install from .msi
- Enable auto launch feature and restart machine, app should start automatically
- Also you should see entry "Tari Universer startup" in task scheduler 

## Summary by CodeRabbit

- **New Features**
  - Enhanced auto-launch functionality with asynchronous updates and dedicated Windows support for a smoother startup experience.
  - Improved blockchain visualization by adding transaction recap tracking for better overview.

- **Chores**
  - Updated the application version and refined configuration formatting for enhanced consistency.


src-tauri/src/auto_launcher.rs
```diff
@@ -146,6 +222,9 @@ impl AutoLauncher {
         let auto_launcher = AutoLauncher::build_auto_launcher(app_name, &app_path)?;
 
         AutoLauncher::toggle_auto_launcher(&auto_launcher, is_auto_launcher_enabled)?;
+        #[cfg(target_os = "windows")]
+        self.toggle_windows_admin_auto_launcher(is_auto_launcher_enabled)
+            .await?;
```
brianp, line 227: Feels like this is maybe something that should be withing `AutoLauncher::toggle_auto_launcher` which already does OS specific things for auto launching?

brianp (commented):
Sorry, i made this suggestion last year but it failed to post. I'll be testing this today.

Misieq01:
@brianp Did you tested on dev or build the app from this PR ?

brianp:
I built the app locally, and installed via msi

Misieq01:
I will look into it but remember that it worked just fine on my windows 11

Misieq01:
@brianp Could you check in windows task scheduler if you got a entry for Tari Universe

brianp:
I uninstalled all Tari versions and cleaned directories. Built from PR, re-installed from msi. Turning "Auto-start on system boot" on did not produce any item I can find in the Task Scheduler. Although their tiny little unordered list doesn't make it easy.

Misieq01:
> I uninstalled all Tari versions and cleaned directories. Built from PR, re-installed from msi. Turning "Auto-start on system boot" on did not produce any item I can find in the Task Scheduler. Although their tiny little unordered list doesn't make it easy.
Task scheduler does not has great UI haha. Okay, thanks for putting time in testing it, I will give it a look today, either something broke after updating branch or there is another case on your setup that need to me handled

Misieq01:
@brianp Tested it and for me it also stopped working, not sure what caused it but removing check of flag from auto_launcher crate and just depending on our app_config flag solved it for me and now it works like it used to before Christmas break

brianp:
I have tested it again. I can now find the entry in the tesk scheduler. But the application does not start on startup. It may be due to the location in the task scheduler?

image

brianp:
It works 😄 

One more problem we'll have to solve: Existing Windows machines that have this option already enabled won't receive the fix. They'd have to toggle auto-start before they would see the change working again.

I think we need to add a config migration for this to handle it in 1-of-2 ways:
1. If the config option is enabled, trigger the new startup mechanism to add the task scheduler
2. If the config option is enabled, disable it on Windows, reflecting the current reality.

Option 1 is preferred. 

I'm happy to merge this PR first, if we can produce a solution for the existing users before the next release (Wednesday).

PR #1226, opened 2024-12-11
chore(ci): revert the removal of openssl/hickory-client/hickory-proto

leet4tari:
Description
Revert the removal of openssl/hickory-client/hickory-proto

Motivation and Context
Fix current build failures

How Has This Been Tested?
Builds in local fork


brianp:
Builds are currently building sans hickory. I'm probably going to close this unless there's a specific need for hickory to be re-introduced.

PR #1229, opened 2024-12-11
fix(ui): animation updates | transaction & earnings fixes | miner metrics and stores cleanups

### LINKED ISSUE #1198: Windows: Won on my pc but now the UI is stuck here post catch up animation
**Describe the bug**
image
in this state, replay doesnt work
and there is no way to get it back to the original block animation
however when miimized it will report on additional winnings, just without the animation

### LINKED ISSUE #1318: 0.8.34 Red long block doesn't animate
**Describe the bug**
On both Windows and Mac, long block is broken. It it winks but never animates out.

https://www.loom.com/share/26a28ed11b094e2c8b2fb0326f6e53d2

shanimal08:
w## Description

**reverted back to actual tauri crate in this PR**

- updated animation build so you can show a replay if not started
- updated on our side to remove starting the animation if not mining (just pass isReplay)
- removed the extra `miner_metrics` listener in [`useSystemTray`](https://github.com/tari-project/universe/pull/1229/files#diff-0f8836d7c431878dea871d881c7beb5bdc2926ab237fe7dba0a0948338e6f77a), and just use the store values instead
- split out miner metrics from the mining store to its own store, [`useMiningMetricsStore`](https://github.com/tari-project/universe/pull/1229/files#diff-3fc3d5af17dd298493c576af53a4508d2495560802e3fcd66d5b447003d4e017)
  - cleaned up mining stores (visualizations, metrics, and ui)
  - moved some actions out of the mining stores to be used outside of hooks ([`miningStoreActions`](https://github.com/tari-project/universe/pull/1229/files#diff-8932d256e2c3fb7ca702d495c4548e785826b529ae16113f71f96534e24a0c78))
- [changed `get_tari_wallet_details` to an emit instead of calling the command from FE at an interval](https://github.com/tari-project/universe/pull/1229/files#diff-2f5e0a90d4195e9986f5e24928dce16b59a80a2cf30f7059b38d55bd7d1eff69R540-R555)
- bump oudated deps
- added splashscreen window to capabilities files
- standardise getting app window - using `getCurrentWindow` whereas before we had a mix of `getCurrentWindow` and `getCurrentWebviewWindow`
- use `app.emit` instead of window in `telemetry_manager`
- removed unnecessary intervals in favour of listeners
- added `useDeferredValue` in places we needed to check for 'stale' data

## Motivation and Context

Fixes #1198
Fixes #1318

## How Has This Been Tested?

https://github.com/user-attachments/assets/2a3faf5d-2d4b-43c2-b990-2ad86fd52574

https://github.com/user-attachments/assets/56310c51-959e-4df7-9955-62fa29eb1378

https://github.com/user-attachments/assets/ee8e5a90-7eb0-4377-b5e7-4318c5261a79

- check that long block and earnings animations work as expected 
  - normal earnings
  - missed earnings
  - and earning replays
- check that the systray is working normally
- check that earnings animations x wallet balance are in sync
- no freezy

## Breaking Changes

\_\_

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/containers/floating/StagedSecurity/sections/SeedPhrase/SeedPhrase.tsx
```diff
@@ -41,11 +42,15 @@ const SeedPhrase = ({ setSection, words }: Props) => {
...
-    }
+    useEffect(() => {
+        const groups: string[][] = [];
+        for (let i = 0; i < words.length; i += 6) {
+            groups.push(words.slice(i, i + 6));
+        }
+        if (groups.length) {
+            setWordGroups(groups);
+        }
+    }, [words]);
```
MCozhusheck, line 53: is this useEffect necessary ? `words` comes from props so changes in `words` will change wordGroups anyway

src/App/AppWrapper.tsx
```diff
@@ -54,15 +33,5 @@ export default function AppWrapper() {
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
```
Misieq01, line 34: We won't be bringing it back ?

shanimal08: yea but it always exists in github :D just a pet peeve of mine to leave in big chunks of commented code, but i don't mind adding it back if you want? :D

src/containers/phase/Setup/components/AppVersion.tsx
```diff
@@ -6,13 +6,20 @@ export default function AppVersion() {
     const [tariVersion, setTariVersion] = useState<string | null>(null);
 
     useEffect(() => {
+        let fetched = false;
```
MCozhusheck, line 9: is this variable necessary ? It's local variable if we want to verify we fetch only once we should use `useState` or `useRef` variable.

src/containers/floating/StagedSecurity/sections/SeedPhrase/SeedPhrase.tsx
```diff
@@ -41,11 +42,15 @@ const SeedPhrase = ({ setSection, words }: Props) => {
         setChecked(!checked);
     };
 
-    const wordGroups: string[][] = [];
-
-    for (let i = 0; i < words.length; i += 6) {
-        wordGroups.push(words.slice(i, i + 6));
-    }
+    useEffect(() => {
```
Misieq01, line 45: I don't think we need useEffect and useState here as words come with props so it will be rerendered either way

shanimal08: oh i updated `eslint-plugin-react-hooks` and it threw an error because of that loop out in the open ☹️ 

i can revert and revert the update?

MCozhusheck (changes requested):

src/App/AppWrapper.tsx
Misieq01: I think its easier to uncomment code then look back through commit history but won't fight on it haha :D

src/containers/floating/StagedSecurity/sections/SeedPhrase/SeedPhrase.tsx
shanimal08: only changed it after the package bump https://github.com/tari-project/universe/pull/1229#discussion_r1908542970

 but yeah i wouldn't do that normally 😬 lmk if i should revert the update and these changes or just leave as is?

src/containers/phase/Setup/components/AppVersion.tsx
```diff
@@ -6,13 +6,20 @@ export default function AppVersion() {
     const [tariVersion, setTariVersion] = useState<string | null>(null);
 
     useEffect(() => {
```
Misieq01, line 8: I don't think `let fetched` make a difference here its defined in scope of useEffect so it won't be preserved between executions of this useEffect

src-tauri/src/main.rs
```diff
@@ -538,6 +536,23 @@ async fn setup_inner(
...
+                break;
+            }
+
+            interval.tick().await;
+            if let Ok(wallet) = commands::emit_tari_wallet_details(app_state).await {
+                drop(w_move_handle.emit("wallet_details", wallet));
+            }
+        }
+    });
+
```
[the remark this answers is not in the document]
shanimal08: neat, thanks! should i do the same for the `miner_metrics` emit loop just above?:D

src/containers/floating/StagedSecurity/sections/SeedPhrase/SeedPhrase.tsx
Misieq01: What about removing hooks and just moving this loop to function outside component like `const groupSeedWords = (words: string[]) => {}` and then even you can do `groupSeedWords(words).map((group, groupIndex) => ...`

src/hooks/mining/useMiningMetricsUpdater.ts
```diff
@@ -1,58 +1,40 @@
...
+
     const displayBlockHeight = useBlockchainVisualisationStore((s) => s.displayBlockHeight);
     const setDisplayBlockHeight = useBlockchainVisualisationStore((s) => s.setDisplayBlockHeight);
 
+    const deferredblock = useDeferredValue(currentBlockHeight);
+
     return useCallback(
         async (metrics: MinerMetrics) => {
             if (metrics) {
                 const isMining = metrics.cpu?.mining.is_mining || metrics.gpu?.mining.is_mining;
```
Misieq01, line 18: These checks for node_connection and setting animation are no longer needed or they were moved somewhere else ?

shanimal08: not needed since `useUiMiningStateMachine` checks against is_mining (gpu & cpu), so that actually caused a bit of weirdness with the animation states and i think `is_mining` for either is connected to being connected already on the BE? (could be wrong about that though)

Misieq01: I was assuming that if someone added it then you can have `is_mining = true` and then `metrics.base_node?.is_connected either true or false` and when user wasn't connected to node then we didn't wanted to show animation of mining. I'm not sure but that's why I asked

src/containers/floating/StagedSecurity/sections/SeedPhrase/SeedPhrase.tsx
shanimal08: nice that did the trick :D updated :D

src-tauri/src/main.rs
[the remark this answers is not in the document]
MCozhusheck: > neat, thanks! should i do the same for the `miner_metrics` emit loop just above?:D

yes previous could be written similar

MCozhusheck (approved):

src/hooks/mining/useMiningMetricsUpdater.ts
Misieq01: Yeah I was thinking the same but someone added code that works like if there could be case where `base_node.connected` can be `false` and `is_mining` `true` so that's why I asked if you know for sure that this is not a case hah

shanimal08: rats! oki awesome fixing, thanks so much!

src-tauri/Cargo.lock
```diff
@@ -7612,16 +7606,18 @@ dependencies = [
...
 dependencies = [
  "tauri",
  "tauri-plugin",
 ]
 
 [[package]]
 name = "tauri-plugin-sentry"
-version = "0.3.0"
-source = "git+https://github.com/tari-project/sentry-tauri?rev=7554b9a8738d6442d175aa2add55ecc11c52842d#7554b9a8738d6442d175aa2add55ecc11c52842d"
+version = "0.2.0"
```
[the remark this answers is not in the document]
shanimal08: nice catch :D i just reverted to what it was before the tauri x tari fork but this branch is so old that a bunch of things have newer releases 😅 updated, ty!

PR #1237, opened 2024-12-12
fix: app restarts after sleepmode

### LINKED ISSUE #877: [Windows] App restarts after waking up from sleep mode
App go again through setup after waking up from sleep. This cause unnecessary waiting but may also mess up config if something gets overwritten.

This was noticed on my windows laptop but still requires confirmation if it's a common problem

Misieq01:
### [ Summary ]
- Created `SystemStatus` utility which listen to PowerStatus of device, for example `ScreenLocked`, `Suspend`, `ScreenUnlocked`
- Adjusted updater to not proceed with autoupdate when device is in `Suspend` or `ScreenLocked` state
- Adjusted update dialog to show information about not updating in this state
- Added flow for stoping all processes when entering sleep mode or locking screen
- Added flow for resuming all processes when leaving sleep mode or locking screen

image
image

## Summary by CodeRabbit

- **New Features**
  - Expanded localization across multiple languages with new translations for common actions (e.g., “Close” and “Update”) and enhanced feedback messages when auto-updates cannot run due to system lock or sleep modes.
  - Introduced a new user interface modal that provides real-time progress updates during application process resumption, offering clearer guidance and improved feedback during update and resume operations.


MCozhusheck:
This issue still persists. After putting machine to sleep mode I still had setup view again, looks like the problem is with `useEffect` which some reason triggers `application_setup`.
Logs in the attachment
[logs.zip](https://github.com/user-attachments/files/18349086/logs.zip)

src-tauri/Cargo.toml
```diff
@@ -95,6 +95,7 @@ hex = "0.4.3"
 openssl = { version = "0.10", features = ["vendored"] }
 ring-compat = "0.8.0"
 der = "0.7.9"
+psp = { git = "https://github.com/Misieq01/psp.git", rev = "ddef0c6" }
```
brianp, line 98: If we're running a forked version of the plugin it'll need to be hosted under the `tari-projects` organisation, and we need an explanation of why we're using the forked version.

src-tauri/src/main.rs
```diff
@@ -1324,7 +1352,15 @@ fn main() {
         "Starting Tari Universe version: {}",
         app.package_info().version
     );
-    app.run(move |app_handle, event| match event {
+
+    let power_monitor = SystemStatus::current().start_listener();
+    app.run(move |app_handle, event| {
+        // We can only recive system events from the event loop so this needs to be here
+        let _unused = SystemStatus::current().recive_power_event(&power_monitor).inspect_err(|e| {
+            error!(target: LOG_TARGET, "Could not recive power event: {:?}", e)
```
brianp, line 1360:
```suggestion
            error!(target: LOG_TARGET, "Could not receive power event: {:?}", e)
```

src-tauri/src/utils/system_status.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub fn start_listener(&self) -> PowerMonitor {
+        let power_monitor = PowerMonitor::new();
+        let _unused = power_monitor.start_listening().inspect_err(|e| {
+            error!(target: LOG_TARGET, "Failed to start power monitor: {:?}", e);
+        });
+
+        power_monitor
+    }
+
+    pub fn recive_power_event(&self, power_monitor: &PowerMonitor) -> Result<(), Error> {
```
brianp, line 58:
```suggestion
    pub fn receive_power_event(&self, power_monitor: &PowerMonitor) -> Result<(), Error> {
```

src-tauri/src/utils/shutdown_utils.rs
```diff
@@ -124,3 +134,255 @@ pub async fn stop_all_processes(
 
     Ok(())
 }
+
+#[allow(clippy::too_many_lines)]
+pub async fn resume_all_processes(app_handle: tauri::AppHandle) -> Result<(), anyhow::Error> {
```
brianp, line 139: **Disclaimer: This is a question for discussion, not a change request**

The thing I get worried about here is that it is similar but different. 

Without comparing line by line, how does another developer know what makes this resume sequence different than our startup sequence? And if they're the same is it possible to re-use or create a common startup sequence? 

I know we have changes coming up for the sequencing and order of the node, mmproxy, and p2pool. I'll be worried we'll change it in one place and not the other. Maybe it's okay to have to know we need to make changes in both places but that returns to the question of what exactly makes them different.

src-tauri/src/utils/system_status.rs
```diff
@@ -0,0 +1,96 @@
...
+        info!(target: LOG_TARGET, "Reciving power event");
```
brianp, line 59:
```suggestion
        info!(target: LOG_TARGET, "Receiving power event");
```

brianp, line 59: I believe this is the log line that is spamming the logs. Maybe we need some kind of debounce because I'm seeing it 50x a second. Which may mean this function is simply being called/polled way to often. 

The fix here isn't just to remove the log line, but lets make sure we're not hamming the function to often and wasting unneeded cycles.

brianp (changes requested):
Logs are flooded:
image

src-tauri/src/utils/shutdown_utils.rs
Misieq01: I don't like it that we have similar code for setup and resuming and would love to have some kind builder/manager which would store all logic related to setting up the app and starting processes but I didn't want to rewrite the setup function on this stage as any bug created there can stop people from entering the app.
As for what is different, resuming function cuts out everything related to starting process from setup function and also change events emitted, so we could with some work unify at least this part I think

src-tauri/src/utils/system_status.rs
Misieq01: This function is triggered when event loop receives new event, so its mainly spam when you move the app window around and I think it should not be any performance issue as its only checks the queue. Also I don't know if there is even fix for that as we cannot skip events ( like trigger if every 10 events ) as we can miss sleep mode entering moment, and I'm not sure if adding timeout will make the difference as we still will trigger it 50 times

brianp:
Tested on windows 👍🏻 
Was notable that in windows Lock mode triggered the pause, but sleep mode didn't. This could be due to my energy settings though.

PR #1239, opened 2024-12-12
fix: updater artifacts V1 compatible

shanimal08:
Description
---
- fixes missing release artifacts after Tauri V2 merge:

needed to change `createUpdaterArtifacts` to [`v1Compatible`](https://tauri.app/plugin/updater/#building) for the zips

How Has This Been Tested?
---

-  beta build branch:

macOs: 

image

linux:
image

windows: 
image

---

- check artifacts in [the release workflow](https://github.com/tari-project/universe/actions/runs/12294412932)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
           - platform: 'windows-latest'
             args: ''
           - platform: 'macos-latest'
-            args: '--target universal-apple-darwin '
+            args: '--target universal-apple-darwin'
     runs-on: ${{ matrix.platform }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


mmrrnn (approved):

brianp:
I wish I had looked closer. This PR did not produce the artifacts as listed in those images. Those are the artifacts it's *looking* for. But only 3 single artifacts are produced.

shanimal08:
> I wish I had looked closer. This PR did not produce the artifacts as listed in those images. Those are the artifacts it's *looking* for. But only 3 single artifacts are produced.
> 
> image
> 
> 

omg got me too 🙈 i previously removed the bundle.targets value so it'd default to "all" but reverted because mac failed. it worked after retrying though (2nd last commit) 

if you add `app` to bundle.target in the conf it'll add those.

still need v1Compatible though (see builds for windows and linux)

PR #1243, opened 2024-12-12
feat: telemetry service

### LINKED ISSUE #1186: Add additional telemetry events
**Description**
Create a new service to send Telemetry events.
- The telemetry endpoint should be ENV configurable
- It should be usable from the backend and frontend (via backend is okay)
- It will send telemetry via HTTPS post requests
- It should respect the users existing "Opt-in for Telemetry" settings

The data structure will be in the following format:

```
event_name : str
event_value  : json
created_at : timestamp
user_id (optional) : str
app_id : str
version: str
```

The first set of telemetry should match the existing setup progression being sent to Sentry. 

With this new event type, we should track setup progress as it happens. Ideally with the captured data we can see what percentage of users successfully made it across the different percents of completion, for example:
image

MCozhusheck:
Description
---
Resolves #1186 
Add new telemetry service which receives events and immediately sends them.

example event sent:
```json
{
  "app_id": "1b%i31CBnblWxHYxSanG",
  "cpu_name": "11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz",
  "created_at": {
    "nanos_since_epoch": 439482034,
    "secs_since_epoch": 1734106140
  },
  "event_name": "checking-latest-version-mmproxy",
  "event_value": { "percentage": 10, "service": "mmproxy" },
  "gpu_name": "Intel(R) Iris(R) Xe Graphics [0x9a49]",
  "os": "Linux",
  "user_id": "v3,FZNcifkHKvDUJ9Luwc4xBe5HiGHV,0.8.24,VWE2AdhnnMOqNDd7fcQnqVVEBAeee68KBM2+8K7iznY=",
  "version": "0.8.24"
}
```

Motivation and Context
---
This should be useful for tracking how many users  are stuck during application setup. We saw some users being stuck at 30% with now feedback. This should help in debugging issues like that.

How Has This Been Tested?
---
Run application and watch for error message coming from telemetry service. If no error are present that means that events were sent correctly. Also check the telemetry dashboard to check if any messages are present on the server.

To verify frontend command I created test button to invoke like this:
```Typescript
const sendTelemetry = () =>
    invoke('send_data_telemetry_service', {
        eventName: 'settings_opened',
        data: { service: 'frontend-test', percentage: 100 },
    });
```
And found those event present in the kafka dashboard:
```js
{
  "app_id": "CX4h8%%G$xQ0hpnZSu@N",
  "cpu_name": "11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz",
  "created_at": {
    "nanos_since_epoch": 795656093,
    "secs_since_epoch": 1736248665
  },
  "event_name": "settings_opened",
  "event_value": { "percentage": 100, "service": "frontend-test" },
  "gpu_name": "Intel(R) Iris(R) Xe Graphics [0x9a49]",
  "os": "Linux",
  "user_id": "v3,FZNcifkHKvDUJ9Luwc4xBe5HiGHV,0.8.40,VWE2AdhnnMOqNDd7fcQnqVVEBAeee68KBM2+8K7iznY",
  "version": "0.8.40"
}
```

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


leet4tari (approved):
utACK - Would need the BETA env

src-tauri/src/main.rs
```diff
@@ -236,20 +240,60 @@ async fn setup_inner(
...
+    let app_version = app.package_info().version.clone();
+    state
+        .telemetry_service
+        .write()
+        .await
+        .init(app_version.to_string(), telemetry_id.clone())
+        .await?;
+    let telemetry_service = state.telemetry_service.clone();
+    let telemetry_service = telemetry_service.read().await;
+    let telemetry_service = telemetry_service.deref();
```
brianp, line 279: interesting. We can't just clone it, or borrow it. We have to deref it?

src-tauri/src/main.rs
```diff
@@ -265,6 +309,17 @@ async fn setup_inner(
         sleep(Duration::from_secs(1));
     }
 
+    drop(
```
brianp, line 312: Why do we have to drop after each call?

src-tauri/src/telemetry_service.rs
```diff
@@ -0,0 +1,234 @@
...
+    let cpu_name = match cpu_name.first() {
+        Some(cpu) => cpu.public_properties.name.clone(),
+        None => "Unknown".to_string(),
+    };
+    let gpu_name = hardware.get_gpu_devices().await?;
+    let gpu_name = match gpu_name.first() {
+        Some(gpu) => gpu.public_properties.name.clone(),
+        None => "Unknown".to_string(),
+    };
+    let os = PlatformUtils::detect_current_os();
```
brianp, line 191: Is it heavy to re-request this data constantly?

Most of this won't change. Could we get it once during initialization and cache it for each request?

brianp (commented):
I have added the telemetry address to CI.

src-tauri/src/telemetry_service.rs
MCozhusheck: Good point, moved it to initialization.

src-tauri/src/main.rs
MCozhusheck: Compiler complains if we ignore result of sending telemetry data. This workaround another possibility could be to write `let _ = ...`

src-tauri/src/main.rs
MCozhusheck: Yes it's unnecessary, we can just use `&` to get reference to rw lock guard.

src-tauri/src/main.rs
brianp: It would be my preference to use `let _ = ...` just because having the drop makes me wonder if there's some special case happening that really requires the drop, where `let _ = ...` is a common pattern.

brianp (approved):

PR #1263, opened 2024-12-13
fix: broken-seed-words-after-reset

### LINKED ISSUE #1203: 0.8.7: Reset Settings: When I reset settings on the app, I find that when the app restarts my wallet seed words are no longer visible
**Describe the bug**
As a user when I reset my settings. When the app restarts, I find that my wallet seed words are no longer visible. Prior to resetting my the settings, the wallet seed words were visible

**To Reproduce**
**Pre-conditions:**

- User did a clean install of the app on Windows (I deleted all files in App data, I deleted wallet from credentials manager, I removed downloaded file and uninstalled existing tari app before downloading and installing)
- User has mining app open
- User has wallet settings visible

**Steps:**
1. Go to 'tari universe'
2. Click on 'Settings' icon
3. Scroll down to 'General' Settings option
4. Scroll down to `Reset Settings`
5. Click on `Reset Settings`
6. Ensure `Reset wallet` toggle is disabled on modal
7. Click on `Yes` within the reset modal
8. App should restart and re-initialize
9. Once app starts
10. Click on `Settings` icon
11. Click on `Wallet`
12. Scroll to `Seed words`
13. Click on eye icon
14. Observe
15. Click on edit icon and observe input box

**Expected behavior**
After a reset user should be able to view their wallet seed words

**Screenshots**
Screenshot (440)
Screenshot (441)

 - Aspire A315-59
 - Windows 11

**Logs**
[logs.zip](https://github.com/user-attachments/files/18035550/logs.zip)

Misieq01:
### [ Summary ]
- Marked wallet directory for skipping deletion when `reset_wallet` is false
- Added extra check for omitting keychain items in deletion process when `reset_wallet` is false


brianp (changes requested):
This mostly looks good. Where it fails is in the MacOS implementation.

When a user first creates the wallet if they deny access to the keychain it creates a backup file in the root application directory `credentials_backup.bin`. This file becomes lost during the reset. So only in the case where credentials are not stored in the keychain the seed words still become lost. 
When stored in the keychain this PR works fine.

We can't simply move the location of the file though, as this would be a breaking change for anyone who already has the file located in the current location.

To correct this situation we would need to migrate the file to the new location in the wallet directory for existing users to ensure a smooth upgrade process.

brianp (changes requested):
Update doesn't work as expected. The `credentials_backup.bin` file is still removed.

This happens because the filename was added to a `folder_block_list`. This list is checked to see if the members of the list are directories. Which `credentials_backup.bin` is not.

Misieq01:
@brianp Sorry, my bad, got lost in ifs and had problems with testing it properly. Thought it would work and waited for someone to test it. I think now it should be working but my mac still do not cooperate with me so I can't test it

brianp (approved):

brianp:
Tested and looks good 👍🏻

Tas4tari:
Tested on Windows 11, app version 0.8.44 and this is now working

Anziskavds:
Tested successfully on MacBook Air M1 2020 13 inch on Sequoia 15.2 on v0.8.44.

PR #1283, opened 2024-12-18
fix: use tari's fork of tauri

stringhandler:
Tauri has a bug in it which causes it to panic. I've created a PR to fix it (https://github.com/tauri-apps/tauri/pull/12000) but it might take a while for them to merge it and release a new version. In the meantime, we can switch to using our fork.


brianp (approved):
LGTM

PR #1291, opened 2024-12-18
fix: retry window fetching

brianp:
Description
---
Seeing errors in Sentry where the main window can't be fetched. Retry for a second before panic'ing.

Motivation and Context
---
A brief delay between window creation and closing may help alleviate the error.

How Has This Been Tested?
---
Error isn't present locally so 🤞🏻


stringhandler (approved):
Ok for now, but should really try get rid of this code altogether

PR #1298, opened 2024-12-19
fix: p2p settings crash

PanchoBubble:
Screenshot 2024-12-19 at 07 52 21


brianp (approved):
Tested. Works for me.

PR #1302, opened 2024-12-20
feat: sys tray reduced updates

PanchoBubble:


brianp (approved):
Works as expected. TaskTray updates every 10s. Seeing much less polling in the logs.

PR #1305, opened 2024-12-23
fix(ui): removed X logo from claim gems and possible linux click fix

peps:
This PR removes the "X" logo from the `Claim Gems` button in the login modal. This is because we now support multi-auth logins. 

I also added `preventDefault` and `stopPropagation` to see if it fixes an issue where this button is not working on Linux. I don't have access to a Linux machine to test, so this is a shot in the dark fix. This is low risk though, since adding those event functions shouldn't effect anything on the working platforms. 

CleanShot 2024-12-23 at 17 10 37


brianp (approved):
Tested on Mac and the button still button'd.

Did not test on other OS's.

PR #1312, opened 2024-12-26
feat(ui): airdrop login and code sharing on setup screen

peps:
This adds a new element to the Setup Screen that allows users to log in and share their referral code while they wait for the App Setup process to complete. 

@PanchoBubble I moved the Airdrop login functionality to its own hook `useAirdropAuth` which exports `handleAuth` and `checkAuth`.

I also disabled the `isFirstLoad` check on the `AirdropPermission` component. I now have it hooked up so if the user disabled telemetry the new Share/Login UI will toggle. This makes it pretty obvious that telemetry is required in order to collect gems. 

Since this changes the code that handles the login functionality and auth checks, we should test the login functionality both on the setup screen and the sidebar.   

CleanShot 2025-01-20 at 16 03 26@2x

CleanShot 2025-01-20 at 16 03 51@2x

CleanShot 2024-12-27 at 13 31 47


src/containers/main/Airdrop/AirdropGiftTracker/hooks/useAirdropAuth.ts
```diff
@@ -0,0 +1,62 @@
...
+            }, 1000);
+            const timeout = setTimeout(
+                () => {
+                    clearInterval(interval);
+                    setAuthUuid('');
+                },
+                1000 * 60 * 5
+            );
+
+            return () => {
```
PanchoBubble, line 53: we could probably have this already in a useEffect here and have a `checkAuth` state, so when you do `handleAuth` it sets `setAuthUuid` and `setCheckAuth` to start the interval checking, so usage is contained within the hook itself

PanchoBubble (approved):

src/containers/main/Airdrop/AirdropGiftTracker/hooks/useAirdropAuth.ts
peps: Good call, just updated it

brianp:
@peps Sorry mate. I was about to merge this and now it's conflicted.

peps:
@brianp it should be ok now. This PR involves moving the Airdrop Login logic into a separate hook `useAirdropAuth` so any change to the inline logic there will be a conflict. 

I tested pretty thoroughly myself, but when testing make sure to test both...

1. The login functionality on the Setup screen
2. The login functionality in the Sidebar

They're both sharing the same `useAirdropAuth` hook to handle the login flow.

shanimal08 (approved):
LGTM! just the new conflicts (sorry!) 
i like this setup screen app version a lot better:D

btw, not sure when #1335 will go in, but it will likely cause even more conflicts, so not sure if you want to wait for that, or resolve here and i can resolve there after this is merged? 

cc @brianp

peps:
@brianp @shanimal08 Conflicts fixed

peps:
@PanchoBubble when you get a chance, mind taking a look at @brianp 's login session issue? 

I'm not sure what the cause would be, i'm pretty sure my changes shouldn't effect the sessions. Maybe the BE for fetching user data was down while Brian was testing?

peps:
@brianp FYI, there was a copy + design update for this component. So this now reflects the latest UI in Figma.

brianp:
I am beginning to think the logout/persistance problem stems from https://github.com/tari-project/universe/pull/1271

As I encountered it more today on different PR's.

peps:
> I am beginning to think the logout/persistance problem stems from #1271
> 
> As I encountered it more today on different PR's.

cc @PanchoBubble

peps:
@brianp if it's not an issue with this PR specifically, can we fix the session issue in a separate PR? It should be an issue currently on `main` as well.

PanchoBubble:
This one should fix several related with auth issues https://github.com/tari-project/universe/pull/1426 @peps @brianp

peps:
conflicts fixed

FYI: we'll run into conflicts whenever `LoggedOut.tsx` is updated since the logic is moved to a `useAirdropAuth` hook in this PR.

peps:
Closing PR

"we decided against merging your loading screen sign-up affordance because we altered the general flow so that you only have to signup before you download, plus it had no way to apply a referral code."

PR #1335, opened 2025-01-02
refactor: move setup to app::ready instead of javascript triggered

stringhandler:
Move setup to be triggered by rust instead of javascript. I believe this is a better pattern, but also there were two places where setup was being called and it was not clear whether this is correct or not


src/containers/floating/ExternalDependenciesDialog/ExternalDependenciesDialog.tsx
```diff
@@ -37,15 +37,17 @@ export const ExternalDependenciesDialog = () => {
         setShowExternalDependenciesDialog(false);
         if (isInitializing) return;
         setIsInitializing(true);
```
PanchoBubble, line 39: We should comment out these too
and we can comment lie 19 and 20 for `setView` and `setCriticalError`

src/containers/floating/ExternalDependenciesDialog/ExternalDependenciesDialog.tsx
```diff
@@ -37,15 +37,17 @@ export const ExternalDependenciesDialog = () => {
...
+        // Setup now done in rust
+        // invoke('setup_application')
+        // .catch((e) => {
+        // setCriticalError(`Failed to setup application: ${e}`);
+        // setView('mining');
+        // })
+        // .then(() => {
+        // setIsInitializing(false);
+        // });
+        setIsInitializing(false);
```
PanchoBubble, line 49: if we comment the lines above we dont need `setIsInitializing(false)` either

PanchoBubble (approved):

src/hooks/app/useSetUp.ts
```diff
@@ -69,10 +69,10 @@ export function useSetUp() {
         if (syncedAidropWithBackend && !isInitializingRef.current) {
             isInitializingRef.current = true;
             clearStorage();
-            invoke('setup_application').catch((e) => {
-                console.error(`Failed to setup application: ${e}`);
-                setCriticalError(`Failed to setup application: ${e}`);
-            });
+            // invoke('setup_application').catch((e) => {
```
mmrrnn, line 72: We have to wait with setting up until `set_airdrop_access_token` is called. It's important for setting up telemetry data

stringhandler: can I emit an event `setup_complete` instead?

mmrrnn: We need to pass the access token from the frontend to the backend before the setup_application process.

brianp:
@shanimal08 can we get your assistance managing some telemetry problems here?

mmrrnn:
I decided to take over this issue. We're discussing about the possibility to move some airdrop logic to rust and reduce ping pong calls. It's on hold for now since we have higher priorities :D

src-tauri/src/app_config.rs
```diff
@@ -232,7 +232,6 @@ pub(crate) struct AppConfig {
     created_at: Option<DateTime<Utc>>,
     mode: MiningMode,
     display_mode: DisplayMode,
-    auto_mining: bool,
```
brianp, line 235: I think this will break existing configs. The value in existing configs won't be ignored, it'll consider the config invalid and fail to parse it, and reset the entire config to defaults. 

This needs to be validated before merge.

src-tauri/src/main.rs
```diff
@@ -982,6 +987,24 @@ fn main() {
                 }
             };
 
+            let token_state_clone = app.state::<UniverseAppState>().airdrop_access_token.clone();
+            let memory_state_clone = app.state::<UniverseAppState>().in_memory_config.clone();
+            app.listen("airdrop_token", move |event| {
```
brianp, line 992: Will this continue to listen for the rest of the apps runtime? And could an airdrop token get changed at a time it's not supposed to?

src-tauri/src/app_config.rs
brianp: I think I'm wrong. It will only break if it was a value in `AppConfigFromFile`

src-tauri/src/main.rs
shanimal08: yes, and not sure about "not supposed to" but it can change if it expires (then refresh it triggered on FE) also if someone logs out and back in for whatever reason

PR #1336, opened 2025-01-02
fix: optimize transaction history

mmrrnn:
Whole history component was rerendered when user hover over balance to see decimals.

* added react-window so we don't render all items at once. Instead, we render additional elements only when user scrolls down
* split components to separate rerendering so we rerender only specific part of the tree that has changed
* added memoization for history component

We can go further and implement infinite scrolling with fetching elements partially when scrolling:
https://codesandbox.io/p/sandbox/r7rp40n0zm?file=%2Fsrc%2Findex.js%3A19%2C35


PanchoBubble (approved):
Nice virtualization was needed there, works great on mac

Misieq01 (approved):

mmrrnn:
Currently I'm working on even better approach of this(pagination, not just virtualization in React). We can either merge it and replace with better solution after that, or wait for combined improvement

PR #1339, opened 2025-01-02
fix: set target framerate to reduce gpu usage

stringhandler:
At a later stage we can use config to set this


shanimal08 (approved):

brianp (approved):
Tested on windows and mac. Saw no detriment to animations.

PR #1345, opened 2025-01-03
feat(ui): release notes modal

peps:
This adds a Release Notes modal that pops up after a new app version is detected.

CleanShot 2025-01-20 at 17 22 50@2x

Functionality requirements:

- The modal will trigger after the Setup screen on the main dashboard
- Only show the popup if the app version and fetched release notes version are a match.
- Only show the popup once per version.
- Only fetch the release notes on startup if the versions are not a match


src/containers/floating/Settings/sections/releaseNotes/useReleaseNotes.ts
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    useEffect(() => {
+        if (!triggerEffect) return;
+
+        const currentAppVersion = packageInfo.version;
+
+        // TODO: Need to save the fetched releaseNotesVersion to persistant storage
+        //       and compare it with the appVersion to determine if the release notes
+        //       dialog has already been shown to the user.
```
PanchoBubble, line 37:
```suggestion
```

PanchoBubble (approved):

src/containers/floating/Settings/sections/releaseNotes/useReleaseNotes.ts
peps: done

src/containers/phase/Setup/Setup.tsx
```diff
@@ -5,9 +5,12 @@ import InfoNav from './components/InfoNav/InfoNav';
 import { SetupWrapper } from '@app/containers/phase/Setup/Setup.styles';
 import grain from '/assets/img/grain.png';
 import AppVersion from './components/AppVersion';
+import { useReleaseNotes } from '@app/containers/floating/Settings/sections/releaseNotes/useReleaseNotes';
 
 export default function Setup() {
     useSetUp();
+    useReleaseNotes({ triggerEffect: true });
```
shanimal08, line 12: LGTM & tested locally with the admin trigger

just a bit concerned about this one in during setup and how it will react with auto update stuff? 😅  might be good for @mmrrnn to have a look through too

peps: @shanimal08 good point, I can move the trigger for the modal to happen after setup completes. That way it wont conflict with the updater.

peps:
@brianp @shanimal08 I've updated the modal to trigger in `MainView` instead of `Setup`, so it should no longer conflict with the update modal. 

I also updated the `Markdown` styles for the latest release notes

src/containers/phase/Setup/Setup.tsx
mmrrnn: Why not to invoke `check_for_updates` to ensure there is no update available?

mmrrnn: We can also add it to the `get_last_changelog_version` response

peps: @mmrrnn I moved it to the `MainView` so it shouldn't conflict with the Updater anymore. 

One weird requirement for this is that `CHANGELOG.md` might not be updated in time for a release. So this will pop only when the `latestVersion` in `CHANGELOG.md` matches the `appVersion` number, and when `latestVersion` does not equal `get_last_changelog_version`

mmrrnn: Yeah, that's fine :+1:

mmrrnn (approved):

peps:
Note: The layout had been changed a bit. 

- Simplified Header
- Got It confirmation button that closes the modal on click
- Scrollbar added

brianp:
Mentioned in DM, but just to add here we're going to revisit this PR and bring all the logic from the front end to the backend. @Misieq01 will look to address this before we merge it.

brianp:
Closing in favour of #1454

PR #1364, opened 2025-01-09
fix: remove persistent window location

### LINKED ISSUE #1352: Every single .39 instance started life in a much smaller window that I had to resize
**Describe the bug**
Every single .39 instance started life in a much smaller window that I had to resize.
The opposite also happened---Universe opening in a huge window on large monitors.

Misieq01:
## Should fix #1352

[ Summary ]
- Removed logic for saving and applying window size and position


brianp (approved):
I decided to actually remove the commented code just to keep things tidy. We can always access it from the history and it'll probably look pretty different as the splash screen will have changed etc.

Misieq01:
@brianp I think its more convenient  to bring back commented code, rather then look back through commit history but won't fight on it

brianp:
I don't disagree, but I have a feeling it's going to look very different from what we had. So in this case cleanup is nicer I think.

Misieq01:
I finished reviews so I will be jumping to it now, but i think the commented part will more or less will be used in similar way , as either way we will be needed to listen for closing event and save cords to config and during startup apply them to main window

brianp:
Maybe we should use https://v2.tauri.app/plugin/window-state/ ?

brianp:
hah. So not our freezing bug, but a different freezing bug 🤦🏻‍♂️ 

A later problem in any case 😆

PR #1371, opened 2025-01-10
feat: add squad to p2pool

stringhandler:
1. add squad to p2pool for telemetry
2. change p2pool to latest version
3. change minotari suite to latest version
4. change p2pool stats to use tokio::watch instead of polling (this means fewer requests to p2pool)

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -33,7 +33,7 @@ use crate::external_dependencies::{
 use crate::gpu_miner_adapter::{GpuMinerStatus, GpuNodeSource};
 use crate::hardware::hardware_status_monitor::{HardwareStatusMonitor, PublicDeviceProperties};
 use crate::internal_wallet::{InternalWallet, PaperWalletConfig};
-use crate::p2pool::models::{Connections, Stats};
+use crate::p2pool::models::{Connections, P2poolStats};
 use crate::progress_tracker::ProgressTracker;
 use crate::tor_adapter::TorConfig;
 use crate::utils::shutdown_utils::stop_all_processes;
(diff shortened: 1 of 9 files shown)


src-tauri/src/commands.rs
```diff
@@ -523,29 +523,13 @@ pub async fn get_monero_seed_words(
...
-        }
-        warn!(target: LOG_TARGET, "Already getting p2pool stats");
-        return Err("Already getting p2pool stats".to_string());
-    }
-    state.is_getting_p2pool_stats.store(true, Ordering::SeqCst);
-    let p2pool_stats = state.p2pool_manager.get_stats().await.unwrap_or_else(|e| {
-        warn!(target: LOG_TARGET, "Error getting p2pool stats: {}", e);
-        None
-    });
+    let p2pool_stats = state.p2pool_latest_status.borrow().clone();
```
brianp, line 528: I've seen some issues with the latest status caching. When you stop mining the last mining status will be cached, and continue to offer whatever stats were last given. Even if no longer connected. (This was fixed for the GPU miner before). 

Just mentioning it here in case it's something that should be looked at in regards to p2p stats. Such as p2p is no longer connected but the cached stats have `connected_since` and makes everyone think it's still connected.

https://github.com/tari-project/universe/pull/1304/files#diff-31c51000768ebc0fedcd78a7c5282936ecee9a65108f5deff570b4d98af9a12cR123-R129

brianp (commented):
One consideration around the caching.

src-tauri/src/commands.rs
stringhandler: That is a good point, although for p2pool it's not relevant since you can't turn off p2pool without restarting. Do you agree?

brianp: That's fine with me for now. For some reason I though p2p got shut off if you weren't mining at all.

PR #1374, opened 2025-01-13
feat: unify app windows

### LINKED ISSUE #1351: Universe sometimes loads two windows/instances on launch
**Describe the bug**
Both possum and greg had this experience on 0.8.38 and 0.8.39 on Windows.

After install, opening Tari Universe produces two app windows.

Misieq01:
## Potentially fixes #1351 

[ Summary ]
- Unify Splash screen and main windows into one
- Moved logic that determined if we should close splashscreen to backend
- Moved lottie animation to react

[ What's missing ]
- Splash screen animation not working after it was moved to react

[ Testing ]
- Just open the app, you should see splashscreen windows and it should change to setup view


src-tauri/src/main.rs
```diff
@@ -959,6 +992,29 @@ fn main() {
...
+                            error!(target: LOG_TARGET, "Could not set window settings: {:?}", e);
+                        }
+                    } else {
+                        error!(target: LOG_TARGET, "Could not get window position or size");
+                    }
+                } else {
+                    error!(target: LOG_TARGET, "Could not get main window");
+                }
+            }
+        }
```
brianp, line 1017: Why is all the window management stuff back in? 

We don't want to re-add this yet.

Misieq01: Will remove it

brianp: Ahhh it makes sense why you wanted to leave it commented instead of removed before if you thought we were bringing it back right away :D

Misieq01:
@brianp Do we want to show splashscreen for full duration of its animation and only then switch to main view or just for the time when app is gettting ready and switching in the middle of animation

brianp:
> Do we want to show splashscreen for full duration of its animation and only then switch to main view or just for the time when app is gettting ready and switch in the middle of animation

I saw your timer for 3 seconds. I think that's reasonable. To let play entirely, and if ready move on. 

I gave it a quick try because tbh I don't recall the animation, but I actually only saw a blank screen, and then the setup screen. But this is probably as you stated, the animation is not playing currently.

Misieq01:
> > Do we want to show splashscreen for full duration of its animation and only then switch to main view or just for the time when app is gettting ready and switch in the middle of animation
> 
> I saw your timer for 3 seconds. I think that's reasonable. To let play entirely, and if ready move on after done playing.
> 
> I gave it a quick try because tbh I don't recall the animation, but I actually only saw a blank screen, and then the setup screen. But this is probably as you stated, the animation is not playing currently.

Yeah, I've asked shanon for help with animation. I'm gonna convert to PR once is ready. I've tested it on all three systems and it changing to setup view just fine. Its small change but maybe it will be good idea that everyone give it quick try as if there is some edge case it will prevent people from seeing the app

shanimal08 (approved):
added the fixed lottie files + tested. works great! :D

brianp:
There was one additional change required in the release ci otherwise we would have seen build failures. I pushed that and will merge on green.

PR #1375, opened 2025-01-13
fix(gpu): make GPU unavailable if all devices are excluded

### LINKED ISSUE #1360: GPU Mining is not recognized ...
**Describe the bug**
Two issues. In both cases GPU mining is not recognized.

1. GPU mining functions anyway despite claiming it is not recognized
2. GPU mining will not start

MCozhusheck:
Description
---
Partially fixes #1360 
Makes entire GPU unavailable for gpu mining if all devices were excluded.

Motivation and Context
---
Right now if all devices are excluded the gpu miner will restart over and over failing each time. This PR makes early check for this case.

How Has This Been Tested?
---
Before mining - turn on GPU mining and turn off every GPU device then:
 - check if GPU mining is not working (check GPU power and if `xtrgpuminer` process is absent)

Also check case if GPU mining is turned on and at least 1 GPU device then check if GPU mining is working as expected. 

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (changes requested):

brianp:
It's very easy to get the buttons into un-syncable states now as the button still reacts to state changes on click in some form.

Preferably if it's the last device in the list, and you switch it off it will switch the GPU off entirely.

https://github.com/user-attachments/assets/2f5e4a65-4afa-4119-94c1-69fb4ab0c7ca

MCozhusheck:
What should be the behavior if we switch all devices off and `GPU miming` becomes disabled and then but turn it back enabled - should we switch all devices as enabled?

brianp:
> What should be the behavior if we switch all devices off and GPU miming becomes disabled and then turn it back enabled should we switch all devices as enabled?

Yes. I think that is perfectly acceptable behaviour. If everything is off, the toggle turns them all on. The toggle can turn them all off.

And if you turn them off 1 by 1, when you turn off the last one, the toggle for all turns off.

Misieq01:
Tested on windows with rtx 4060 and Intel Iris Xe, works like a charm

PR #1379, opened 2025-01-14
chore: go back to Tauri crate | bump outdated deps

shanimal08:
Description
---
- bump outdated npm deps (only items not on `wanted` version)
  - updated files where needed(like in the SeedPhrase file for eslint)
- use the original tauri crates again (instead of our fork)
- consolidated capability files
- fixed lints

Motivation and Context
---

wanted to bring some of the extra work done in #1229 into their won PRs, for less reviewing heaviness there + to go in sooner if that PR isn't merged any time soon

How Has This Been Tested?
---

locally


src-tauri/Cargo.lock
```diff
@@ -7431,8 +7437,9 @@ dependencies = [
 
 [[package]]
 name = "tauri"
-version = "2.1.1"
-source = "git+https://github.com/tari-project/tauri.git?rev=67a06c8a9bae94f412b8059dfa8b4d8dd8ea0a25#67a06c8a9bae94f412b8059dfa8b4d8dd8ea0a25"
+version = "2.2.2"
```
brianp, line 7440: This concerns me, because on CI it was updating to tauri 2.2.2 which caused builds to break. 

So CI is locked to 2.1.1. Which is messy, as it's not what we're all running etc.

shanimal08: CI looks like it ran fine since that's where the install happens - https://github.com/tari-project/universe/actions/runs/12768650543/job/35589473800?pr=1379

also this it just `tauri` not `tauri-cli` :D + on `main` the cargo.lock includes `tauri-plugin-cli` at `2.2.0` anyway 😅

shanimal08:
closing bc #1229 went in

PR #1396, opened 2025-01-17
feat: add new base node grpc call with new nextnet release

hansieodendaal:
Description
---
Added the new base node gRPC call `get_network_state` to `fn get_network_hash_rate_and_block_reward` that returns all required base node statuses in a single call.

**The new base node gRPC call `get_network_state` came with the new Tari base layer nextnet release, `v1.10.0-rc.0`**

`MinerMetrics` now has additional members

- `initial_sync_achieved` (renamed from `is_synced`, describing the actual status)
- `failed_checkpoints`
- `base_node_state` 

and also provides a method `MinerMetrics::is_synced()` that will return the is-synced status based on the `base_node_state`.

**Notes:** 
- New `MinerMetrics` must be wired into the code where applicable.

Motivation and Context
---
To implement the new more efficient base node gRPC method call

How Has This Been Tested?
---
CI

---
Code review

Breaking Changes
---

- [X] None
- [ ] Requires hard fork
- [ ] Other

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -41,6 +41,7 @@ use crate::wallet_adapter::{TransactionInfo, WalletBalance};
 use crate::wallet_manager::WalletManagerError;
 use crate::{node_adapter, UniverseAppState, APPLICATION_FOLDER_ID};
 
+use crate::node_adapter::BaseNodeState;
 use base64::prelude::*;
 use keyring::Entry;
 use log::{debug, error, info, warn};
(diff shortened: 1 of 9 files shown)


SWvheerden (approved):

src-tauri/src/commands.rs
```diff
@@ -401,17 +413,11 @@ pub async fn get_miner_metrics(
...
+        failed_checkpoints,
     } = node_status;
-    // let (sha_hash_rate, randomx_hash_rate, block_reward, block_height, block_time, is_synced) = state.node_manager
-    //     .get_network_hash_rate_and_block_reward().await
-    //     .unwrap_or_else(|e| {
-    //         if !matches!(e, NodeManagerError::NodeNotStarted) {
-    //             warn!(target: LOG_TARGET, "Error getting network hash rate and block reward: {}", e);
-    //         }
-    //         (0, 0, MicroMinotari(0), 0, 0, false)
-    //     });
```
brianp, line 414: Any reason to keep this? 

```suggestion
```

brianp (approved):
Looks really clean. Just waiting for CDN to be updated so I can test.

brianp:
CDN updated. Tested and works with the updated binaries.

src-tauri/src/commands.rs
hansieodendaal: That looks like a bad merge - I removed it

stringhandler (commented):
Will need to revert the rename of `initial_sync_achieved` to `is_synced` but will do that in a new pr

PR #1403, opened 2025-01-20
chore: move system tray implementation to the backend

mmrrnn:
Related: [#1397](https://github.com/tari-project/universe/issues/1397)

Description
--
Move system tray implementation to the backend. From user perspective it almost hasn't changed. About item opens app metadata only on macs(left it as it was for now)
* Format token decimals for est. earning.
* Shorten texts(tooltip has limit so text was cut on windows)
* Combine minimize/unminimize into one toggle button.

Linux:
[Screencast from 01-21-2025 05:29:36 PM.webm](https://github.com/user-attachments/assets/90d79995-e342-4ddf-883a-61ee6d72f988)

Windows:
Tooltip:
image
Menu:
image

Macbook:
Tested on mac, haven't recorded it :(


brianp (changes requested):
image

The toptip, of which will not accept photos at this time 😆 shows the correct number formats, where the menu bar does not.

src-tauri/src/systemtray_manager.rs
```diff
@@ -0,0 +1,239 @@
...
+                    }
+                } else {
+                    info!(target: LOG_TARGET, "Minimizing window");
+                    window.minimize().unwrap_or_else(|error| {
+                        error!(target: LOG_TARGET, "Failed to minimize window: {}", error);
+                    });
+                }
+            },
+            _ => {
+                println!("menu item {:?} not handled", event.id);
```
brianp, line 196: Would info or debug be better here? It's not *super* useful information unless we're looking for missed events.

src-tauri/src/systemtray_manager.rs
```diff
@@ -0,0 +1,239 @@
...
+                        error!(target: LOG_TARGET, "Failed to get main window");
+                        return;
+                    }
+                };
+
+                if window.is_minimized().unwrap_or(false) {
+                    info!(target: LOG_TARGET, "Unminimizing window");
+                    match PlatformUtils::detect_current_os() {
+                        CurrentOperatingSystem::Linux => {
+                            window.hide().unwrap_or_else(|error| error!(target: LOG_TARGET, "Failed hide window: {}", error));
```
brianp, line 174: We have to hide it before we unminimize it?

src-tauri/src/systemtray_manager.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        self.menu.replace(menu);
+        self.tray.replace(tray);
+
+        Ok(())
+    }
+
+    pub fn update_tray(&mut self, data: SystemTrayData) {
+        if let Some(tray) = &self.tray {
+            if let Err(e) = tray.set_tooltip(Some(self.get_tooltip_text(data.clone()))) {
```
brianp, line 208: Maybe `get_tooltip_text` should return an `Option<String>` ? 

Only if it makes the code simpler. It means `get_tooltip_text` could return None for linux. And if set_tooltip takes an option then everything should flow nicely.

src-tauri/src/utils/formatting_utils.rs
```diff
@@ -0,0 +1,58 @@
...
+    } else if hashrate < 1_000_000_000.0 {
+        format!("{:.2} MH/s", hashrate / 1_000_000.0)
+    } else if hashrate < 1_000_000_000_000.0 {
+        format!("{:.2} GH/s", hashrate / 1_000_000_000.0)
+    } else if hashrate < 1_000_000_000_000_000.0 {
+        format!("{:.2} TH/s", hashrate / 1_000_000_000_000.0)
+    } else {
+        format!("{:.2} PH/s", hashrate / 1_000_000_000_000_000.0)
+    }
+}
```
brianp, line 37: Do we not already have formatters? Or are those in js only?

mmrrnn: All are in js. We haven't needed them in rust before.

src-tauri/src/systemtray_manager.rs
mmrrnn: `window` works really bad on linux. I played with this a lot, it won't never unminimize using the mac/windows scenario

brianp:
Tested on MacOS and Windows.

Validated `main` grows in memory on the front end, where this branch grows, and drops, in repetition. So the memory is being garbage collected somewhere along the way.

PR #1410, opened 2025-01-21
fix: missing dependency dialog | add flag to state for missing deps to emit after FE ready

shanimal08:
Description
---

- add flag (`has_missing_dependencies`) to state for missing deps to emit after FE ready
- fix missing key in dialog and button x theme issues

Motivation and Context
---

https://github.com/tari-project/universe/issues/1421


src-tauri/src/main.rs
```diff
@@ -212,6 +212,7 @@ async fn setup_inner(
             .await;
 
         if is_missing {
+            *state.has_missing_dependencies.write().await = true;
```
brianp, line 215: 🤔

PR #1415, opened 2025-01-21
fix(ui): change wording on ludicrous warning modal cancel button

peps:
The Ludicrous warning modal cancel button used to say "Revert to Eco". That wording was misleading since the user wasn't really "reverting" anything. This change makes the button work for both Eco mode and Custom mode and it acts as a clear "Opt-out" for Ludicrous mode. 

I also changed the title of the modal from "Stay in Ludicrous Mode?" to "Switch to Ludicrous Mode?".

CleanShot 2025-01-22 at 11 11 35@2x

### DIFF
--- a/public/locales/en/components.json
+++ b/public/locales/en/components.json
@@ -13,6 +13,6 @@
     "description": "Ludicrous mode may make your system unresponsive. Please confirm you want to start mining in Ludicrous mode or Tari Universe will switch back to Eco mode in",
     "seconds": "seconds",
     "keepChanges": "Keep Changes",
-    "revertToEco": "Revert to ECO"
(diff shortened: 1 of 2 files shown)


shanimal08 (approved):

brianp:
I still think "Stay in Ludicrous Mode?" is weird. Because we aren't actually in Ludicrous mode yet. Maybe something like "Switch to Ludicrous Mode?" ?

peps:
> I still think "Stay in Ludicrous Mode?" is weird. Because we aren't actually in Ludicrous mode yet. Maybe something like "Switch to Ludicrous Mode?" ?

@brianp makes sense, the title is now updated.

PR #1417, opened 2025-01-21
feat(ui): updated universe logo in the sidebar

peps:
This PR updates the sidebar logo to the plash screen logo. Before it was using the plain text "Tari Universe". This change looks nicer and is better for branding. 

BEFORE: 

CleanShot 2025-01-21 at 12 28 31@2x

AFTER: 

CleanShot 2025-01-21 at 12 22 33@2x


shanimal08 (approved):
looks dope 💜

brianp:
Universe text is now again in new layout.

Although I was in complete support of this change. I will now grudgingly close the PR as it's no longer needed.

PR #1433, opened 2025-01-23
fix(ui): airdrop join copy update

peps:
Small copy update for the Airdrop login card in the sidebar

CleanShot 2025-01-23 at 09 41 35@2x


brianp:
#yolo

PR #1434, opened 2025-01-24
fix: mining animation sync [DNM]

shanimal08:
Description
---
- bring down `wallet_details` emit interval to 3s
- split out the metrics state setter, so we can update `block_height` & `block_time` separately
- move new block check logic to where the listeners are so we can use refs on the previous metrics payload
- 

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


shanimal08:
closing because this doesn't work either

PR #1437, opened 2025-01-24
feat: containerization

MCozhusheck:
Description
---
Adds `Dockefile` and script with allows to run container with Tari Universe as AppImage. This was done as proof of concept of creating lightweight VM/container to allow running TU on different machines with various OS and architecture.

The conclusion is that it would be impossible to have GPU mining available on these different machines - this could only work if host machine was also linux with nvidia drives, take a look at [rocker](https://github.com/osrf/rocker) which allows running container with X11 tunneling and accessing host GPU with nvida.

But in any other cases this is impossible. The only reasonable way to run TU with different CPU and GPU architecture would be some hosted cloud machine with remote desktop. 

Motivation and Context
---
Allow testing TU on different OS and architecture from single machine.

How Has This Been Tested?
---
Linux only:
Allow access to X11 server (at least for time of testing) with command `xhost +local:root ` later it can be later revoked with  `xhost -local:root `
Copy AppImage to root directory and rename all recurrences of `Tari Universe (Alpha)_0.8.45_amd64.AppImage` to your just copied file.
build docker image like `docker build -t taritest . ` 
run this image from script `./docker_script.sh `

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
Can we move this into some kind of containerization document? 

I'm hesitant to merge as is because than people will expect us to maintain it over time. Maybe we add a `/docs/containerization/` folder with a containerization.md and the dockerfile and script as examples.

leet4tari:
I don't think this as value. Only Linux user might use this. Is there a reason why the AppImage is embedded in the docker image vs just mounting the folder holding the AppImage at run time?

Linux users can run Windows platform without compiling their own images using something like https://github.com/dockur/windows

I would think that what Linux distros we are testing on or supported might be better info.

markovAutomorphism:
Hi I am trying to get this containerization to work on linux. I have made two small changes to docker_script.sh during my testing so far --env="TARI_NETWORK=mainnet" and --volume="$HOME/.local/share/com.tari.universe:/root/.local/share/com.tari.universe". 

I have tari_universe_1.0.3_amd64.AppImage in the same directory as the dockerfile. When I run the docker script after building the docker image, I see the error:

exec /app/entrypoint.sh: exec format error. What could fix this?

PR #1438, opened 2025-01-26
feat: implement events manager

mmrrnn:
Adjustment possible thanks to https://github.com/tari-project/tari/pull/6763

Description
---
Create a backend -> frontend middleware which controls data flow and events triggering. Example: It'll ensure wallet scanning is finished to detect win/lose when new block mined and emit update events in a controlled order and frequency.

Changes
---
At the end of `setup_inner`, we call `initialize_frontend_updates` which updates frontend with initial data(wallet address, hardware details etc.) and waits for changes in tokio channels. We call `EventsManger` to handle updates for specific update using `EventService` and call `EventEmitter` 

Improvements
---
* Reduce unwanted rerender
* Solve problem when importing wallet with extensive tx history - update balance when wallet finished scanning
* Robust win/lose block determination
* Unify new block height ui updates timing

Losing new block height:
[lose.webm](https://github.com/user-attachments/assets/10de3c75-9f4a-4b7f-bf96-4fe91c42e345)
Winning new block height:
[win.webm](https://github.com/user-attachments/assets/b840dedb-de8c-4197-bdd0-76ed9a107da1)
Initial wallet scanning:
scanning


brianp (changes requested):
Tested on Mac, I'm winning every block but being shown tower bust/lose animations every time. 

I thought the refactoring here would solve timing issues and was expecting this to work smoother.

Misieq01:
Tested `win/lose animations`, switching mining mode, replaying win from history, on **Windows** | **Linux** | **MacOS**

- **Linux** => Everything works
- MacOS => Everything works
- Windows => Switching modes works, replay works, but found extra case for animations when you win blocks to fast then they can be broken

src/App/App.tsx
```diff
@@ -25,6 +26,7 @@ export default function App() {
     const setError = useAppStateStore((s) => s.setError);
     const setIsWebglNotSupported = useUIStore((s) => s.setIsWebglNotSupported);
     const { t } = useTranslation('common', { useSuspense: false });
+    useTauriEventsListener();
```
Misieq01, line 29: What about putting it in AppWrapper.tsx with the rest of listeners

src/containers/floating/Settings/sections/connections/ConnectionStatus.tsx
```diff
@@ -7,7 +7,8 @@ import { useMiningMetricsStore } from '@app/store/useMiningMetricsStore.ts';
 
 export default function ConnectionStatus() {
     const { t } = useTranslation('settings');
-    const isConnectedToTariNetwork = useMiningMetricsStore((s) => s.base_node?.is_connected);
+    const isConnectedToTariNetwork = useMiningMetricsStore((s) => s.connected_peers?.length > 0);
```
Misieq01, line 10: Its not a big deal but wouldn't it be better to just set is_connected flag on the backend and send it to frontend instead doing some conditions on frontend ? In that case we could easily expand is_connected flag on backend if there would be a need in the future

src/containers/floating/StagedSecurity/sections/ProtectIntro/ProtectIntro.tsx
```diff
@@ -13,8 +13,10 @@ interface Props {
 export default function ProtectIntro({ onButtonClick, isLoading }: Props) {
     const { t } = useTranslation(['staged-security'], { useSuspense: false });
 
-    const balance = useWalletStore((state) => state.balance);
-    const formatted = formatNumber(balance || 0, FormatPreset.TXTM_COMPACT);
+    const calculated_balance = useWalletStore((state) => state.calculated_balance);
+    const formatted_balance = calculated_balance
+        ? formatNumber(calculated_balance || 0, FormatPreset.TXTM_COMPACT)
+        : '-'; // Wallet still scanning
```
Misieq01, line 19: Do we know if wallet is currently scanning ? If so maybe it will be good idea to put some loading indicator in the future here

src/containers/main/Dashboard/MiningView/components/BlockTime.tsx
```diff
@@ -6,13 +6,13 @@ import { useMiningMetricsStore } from '@app/store/useMiningMetricsStore.ts';
...
 function BlockTime() {
     const { t } = useTranslation('mining-view', { useSuspense: false });
-    const isCPUMining = useMiningMetricsStore((s) => s.cpu.mining.is_mining);
-    const isGPUMining = useMiningMetricsStore((s) => s.gpu.mining.is_mining);
-    const blockTime = useBlockchainVisualisationStore((s) => s.displayBlockTime);
-    const isConnectedToTari = useMiningMetricsStore((s) => s.base_node?.is_connected);
+    const isCPUMining = useMiningMetricsStore((s) => s.cpu_mining_status.is_mining);
+    const isGPUMining = useMiningMetricsStore((s) => s.gpu_mining_status.is_mining);
+    const displayBlockTime = useBlockchainVisualisationStore((s) => s.displayBlockTime);
+    const isConnectedToTari = useMiningMetricsStore((s) => s.connected_peers?.length > 0);
```
Misieq01, line 12: See comment in ConnectionStatus.tsx

src/store/useMiningMetricsStore.ts
```diff
@@ -1,42 +1,105 @@
...
+        set({ gpu_mining_status });
+    },
+    setCpuMiningStatus: (cpu_mining_status) => {
+        set({ cpu_mining_status });
+    },
+    handleConnectedPeersUpdate: (connected_peers) => {
+        set({ connected_peers });
+
+        const { miningInitiated } = useMiningStore.getState();
+        const wasNodeConnected = getState().connected_peers?.length > 0;
```
Misieq01, line 67: See comment in ConnectionStatus.tsx

src-tauri/src/events_emitter.rs
```diff
@@ -0,0 +1,166 @@
...
+    commands::CpuMinerStatus,
+    hardware::hardware_status_monitor::PublicDeviceProperties,
+    wallet_adapter::{TransactionInfo, WalletBalance},
+    BaseNodeStatus, GpuMinerStatus,
+};
+
+const LOG_TARGET: &str = "tari::universe::events_emitter";
+
+#[derive(Debug, Serialize, Clone)]
+pub enum FrontendEventType {
```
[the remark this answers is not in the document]
Misieq01: I think just EventType would be enough or EmitEventType

src-tauri/src/events_emitter.rs
```diff
@@ -0,0 +1,166 @@
+// Copyright 2024. The Tari Project
```
Misieq01, line 1: Could we include event from events.rs also here ?

src-tauri/src/events_emitter.rs
```diff
@@ -0,0 +1,166 @@
...
+//
```
Misieq01, line 2: Won't fight for it but could we group event handling releated files into one folder ?

src-tauri/src/main.rs
```diff
@@ -791,12 +903,12 @@ async fn setup_inner(
 #[derive(Clone)]
 struct UniverseAppState {
     stop_start_mutex: Arc<Mutex<()>>,
-    base_node_latest_status: Arc<watch::Receiver<BaseNodeStatus>>,
-    wallet_latest_balance: Arc<watch::Receiver<Option<WalletBalance>>>,
+    node_status_watch_rx: Arc<watch::Receiver<BaseNodeStatus>>,
+    #[allow(dead_code)]
+    wallet_state_watch_rx: Arc<watch::Receiver<Option<WalletState>>>,
```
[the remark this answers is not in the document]
mmrrnn: So this field is actually being used indirectly through the `EventsManager`. I'll remove it for now

src/containers/main/SideBar/components/Wallet/Wallet.tsx
```diff
@@ -66,7 +65,8 @@ export default function Wallet() {
                             text={t('paper-wallet-tooltip-message')}
                         />
                     )}
-                    {balance ? (
+                    {/* TODO: User might have spent his rewards */}
```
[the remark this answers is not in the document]
mmrrnn: I've notice we hide "Show reward history" button when use has zero balances. He could've mined 100 blocks and send his funds to other wallet(via minotari console wallet or somehow in future). In this case user won't be able to see this button

src/containers/floating/Settings/sections/connections/ConnectionStatus.tsx
mmrrnn: I've set this flag on the frontend during the `ConnectedPeersUpdate` event. I prefer not to modify the base node status type returned unaltered from the wallet grpc server. More refactoring stuff will be done in a separate PR, as Brian asked for minimal changes for this.

src-tauri/src/events_emitter.rs
mmrrnn: I left it as it is because it requires more changes to be done there and I don't want to make this pr even bigger. More refactoring stuff will be done in a separate PR, as Brian asked for minimal changes for this.

src-tauri/src/events_emitter.rs
mmrrnn: I left it as it is because it requires more changes to be done there and I don't want to make this pr even bigger. More refactoring stuff will be done in a separate PR, as Brian asked for minimal changes for this.

PR #1439, opened 2025-01-27
fix: localstorage persisted tokens

PanchoBubble:
* Prevents multiple calls to set tokens on startup
* Waits for mm proxy setup before setting tokens from localstorage
* Removes old flare set
* Adds flare on login if installRewards is set
* Adds event listener on referral install rewards


src/hooks/app/useSetUp.ts
```diff
@@ -20,12 +21,17 @@ export function useSetUp() {
...
         async function initWithToken() {
             const beConfig = await fetchBackendInMemoryConfig();
+            await getExistingTokens();
             if (beConfig?.airdropUrl) {
                 await handleRefreshAirdropTokens(beConfig.airdropUrl);
             }
         }
-        void initWithToken();
-    }, []);
+        if (mmProxyVersion) {
```
PanchoBubble, line 29: This was causing issues before, it was trying to set the tokens and BE was failing for missing mmProxy setup

src/store/useAirdropStore.ts
```diff
@@ -222,28 +221,52 @@ export const setAirdropTokens = async (airdropTokens?: AirdropTokens) => {
...
+export const fetchBackendInMemoryConfig = async () => {
     let backendInMemoryConfig: BackendInMemoryConfig | undefined = undefined;
 
     try {
         backendInMemoryConfig = await invoke('get_app_in_memory_config', {});
         const airdropTokens = (await invoke('get_airdrop_tokens')) || {};
-        const newState = {
-            ...currentState,
+        const newState: AirdropState = {
+            seenPermissions: false,
```
BalazsSevecsek, line 268: why does it remain false in case of the airdropTokens being present?

src/store/useAirdropStore.ts
```diff
@@ -222,28 +221,52 @@ export const setAirdropTokens = async (airdropTokens?: AirdropTokens) => {
...
-    const currentState = useAirdropStore.getState();
-
-    // Checks for old persisted tokens
+export const getExistingTokens = async () => {
     const existingTokensStore = localStorage.getItem('airdrop-store');
     let existingTokens: AirdropTokens | undefined = undefined;
     if (existingTokensStore) {
         try {
-            existingTokens = (JSON.parse(existingTokensStore).state as AirdropState).airdropTokens;
+            const parsedStore = JSON.parse(existingTokensStore);
```
BalazsSevecsek, line 229: shouldn't we check as a precondition if the fetchBackendInMemoryConfig had already brought up a valid token which was put into the store already?

src/store/useAirdropStore.ts
PanchoBubble: these are not being used will remove

src/store/useAirdropStore.ts
PanchoBubble: fetchBackendInMemoryConfig gets called first, there shouldn't be a persisted token in localstorage if there is one coming from BE. This is needed only once, can be removed after a couple versions to make sure sessions are persisted

src/store/useAirdropStore.ts
```diff
@@ -252,27 +262,14 @@ export const fetchBackendInMemoryConfig = async () => {
...
-                console.error('get_app_in_memory_config error:', e);
-            }
         }
 
         useAirdropStore.setState(newState);
     } catch (e) {
         console.error('get_app_in_memory_config error:', e);
     }
+    if (!backendInMemoryConfig?.airdropUrl) {
+        console.error('Error geting BE in memory config');
```
Krakaw, line 272:
```suggestion
        console.error('Error getting BE in memory config');
```

Krakaw (approved):
From a code perspective it LGTM

PR #1454, opened 2025-01-29
feat: move release notes logic to rust

Misieq01:
### [ Summary ]
- Moved all logic related to fetching and triggering ReleaseNotes modal to Rust => `release_notes.rs`
- Modal display is now triggered at the end of setup with event being emitted once all rules are met
- Added retires to fetching of release_notes
- Added system for caching release notes to stop fetching them each time we open page in settings
- Updating `last_release_notes_version_shown` now trigger by closing release notes dialog
- Cleanup frontend logic

------------------

### [ Functionalities descriptions ]

- #### `should_show_release_notes`
Determines if we should show release_notes by checking two cases:
   - Is current `release_notes_version` higher then `last_release_notes_version_shown`
   - Is app on version equal or higher then current `release_notes_version`
If both cases are true then we returns `true` and later emit event for showing dialog

- #### `should_force_update`
  Checks if current app version is higher then `last_release_notes_version_shown`
  It ensures that we won't used cached versions if exist and instead fetch straight from the server

- #### Caching
Scenarios:
  - If we don't have cache => then always end up downloading
  - If we have cache with the same app version that we are currently on (e.g. app is on 0.8.50 and cache is on 0.8.50 ) => then use cache and also every time TIME_BETWEEN_FETCHES passes ( currently set to 1 hour ) we fetch e_tags to ensure 
that cache matches the server and fetch if not

------------------

### [ Examples of when we show dialog ]
There are three variables that are compared:
- Current app version
- Current release notes version
- Last shown release notes version

[ Example 1] **Show dialog** 
- Current app version => 0.8.50
- Current release notes version => 0.8.50
- Last shown release notes version => 0.8.46

[ Example 2] **Show dialog** 
- Current app version => 0.8.50
- Current release notes version => 0.8.48
- Last shown release notes version => 0.8.46

[ Example 3] **Do not show dialog**
- Current app version => 0.8.46
- Current release notes version => 0.8.50
- Last shown release notes version => 0.8.46

[ Example 4] **Do not show dialog** 
- Current app version => 0.8.50
- Current release notes version => 0.8.50
- Last shown release notes version => 0.8.50

[ Example 5] **Do not show dialog** 
- Current app version => 0.8.50
- Current release notes version => 0.8.46
- Last shown release notes version => 0.8.46
------------------

### [ Testing  ]
For testing I was playing around with changing versions of `last_release_notes_version_shown`  in `app_config.json`
and version of the app in dev environment


src-tauri/src/commands.rs
```diff
@@ -1791,3 +1793,51 @@ pub async fn proceed_with_update(
...
+        .map_err(|e| e.to_string())?;
+
+    if timer.elapsed() > MAX_ACCEPTABLE_COMMAND_TIME {
+        warn!(target: LOG_TARGET, "get_release_notes took too long: {:?}", timer.elapsed());
+    }
+    Ok(release_notes.content)
+}
+
+#[tauri::command]
+pub async fn update_last_shown_release_notes_version(
```
brianp, line 1821: Yeah, i just don't see why this function needs to exist. See other comments.

src-tauri/src/main.rs
```diff
@@ -758,6 +761,16 @@ async fn setup_inner(
...
 
+    let should_show_release_notes = ReleaseNotes::current()
+        .should_show_release_notes(state.clone(), app.clone())
+        .await?;
+
+    if should_show_release_notes {
+        app.emit("show_release_notes", ()).inspect_err(
+            |e| error!(target: LOG_TARGET, "Could not emit event 'setup_message': {:?}", e),
+        )?;
+    }
```
brianp, line 772: See this makes sense to me. The BE telling the FE what to do.

brianp (changes requested):
I could be wrong about how this functions, but I think we could still alleviate knowledge of this management from the front end more.

src/store/appStateStore.ts
```diff
@@ -99,6 +106,29 @@ export const useAppStateStore = create<AppState>()((set, getState) => ({
...
+            console.error('Error fetching release notes', error);
+        }
+    },
+    updateLastShownReleaseNotesVersion: async () => {
+        try {
+            await invoke('update_last_shown_release_notes_version');
+        } catch (error) {
+            console.error('Error updating last shown release notes version', error);
+        }
+    },
```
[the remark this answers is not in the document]
Misieq01: The idea behind is that when we update `last_shown_release_notes_version` we won't show release notes for current version again and initially I've been updating it from backend but this in the end look clearer for me and prevents case when something crashes, `last_shown_release_notes_version`  is updated in meantime and user never sees release notes dialog for this version when restarts the app ( this is about case when we update from backend ). As for this function its triggers when user close release notes dialog so he we know he saw it for sure. I can revert it back to backend if you want but I like it more this way

src-tauri/src/commands.rs
```diff
@@ -1791,3 +1793,51 @@ pub async fn proceed_with_update(
     }
     Ok(())
 }
+
+#[tauri::command]
+pub async fn get_release_notes(
```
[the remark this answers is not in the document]
Misieq01: I mean, I've focused to much on moving all logic related to fetching and triggering to backend that I forgot that we can change way of requesting release_notes from fronted haha. Good point. What about sending release notes with event emit at the end of setup where we send event for triggering dialog modal ? Second option is to fetch it in a loop like every hour and send to frontend but I think its overkill and first option is enough

src-tauri/src/release_notes.rs
```diff
@@ -0,0 +1,312 @@
...
+
+use crate::{
+    events::ReleaseNotesHandlerEvent, updates_manager::UpdatesManager, UniverseAppState,
+    APPLICATION_FOLDER_ID,
+};
+
+const LOG_TARGET: &str = "tari::universe::release_notes";
+const CHANGELOG_URL: &str = "https://cdn.jsdelivr.net/gh/tari-project/universe@main/CHANGELOG.md";
+const RELEASE_NOTES_FILE_NAME: &str = "CHANGELOG.json";
+const TIME_BETWEEN_FETCHES: Duration = Duration::from_secs(1 * 1 * 1); // 1 hour
```
MCozhusheck: I would prefer something like `FETCH_INTERVAL`

MCozhusheck (approved):

mmrrnn (approved):
Successfully tested by changing tauri.conf version to 0.8.46

image

PR #1459, opened 2025-01-31
refactor: motion animations performance fixes

shanimal08:
Description
---
- replace `framer-motion` with `motion` (latest version, motion is now independent)
- use the lighter-wight [`'motion/react-m'`](https://motion.dev/docs/react-reduce-bundle-size#reduce-size) import where we can
- changed the `LazyMotion` feature to `domAnimation` instead of `domMax` and removed the layout animations
- replaced direct `style` animations that used state variables, with `useMotionValue`(s) to prevent re-renders, which also came with a few changes and improvements
  - in [`BlockHeightAccent.tsx`](https://github.com/tari-project/universe/pull/1459/files#diff-4d192051ef0a95bd69da9c152724a2099abb1c0a2eea328f4a2215568d899000): changed the scaling values from updating `fontSize` and `width`, to just use`width` and `scale` directly as motion values
  - in [`ButtonOrbitAnimation.tsx`](https://github.com/tari-project/universe/pull/1459/files#diff-15e9f16e473eb22538a7c75f276e2c337754f6bc95adc19e347950eb8744ae3a): simplified the orbit animations and used `useTime` and `useTransform`, also removed adding more blocks over time
- also memoized some components in the setup flow that had a lot of re-renders 

Motivation and Context
---

locally, using [`react-devtools`](https://react.dev/learn/react-developer-tools) and checking the profiling against `main`

before: 
[profiling-data.BEFORE.SetupAndMiningButton03-02-2025.15-30-22.json](https://github.com/user-attachments/files/18643977/profiling-data.BEFORE.SetupAndMiningButton03-02-2025.15-30-22.json)

after:
[profiling-data.AFTER.SetupAndMiningButton.03-02-2025.15-52-06.json](https://github.com/user-attachments/files/18643987/profiling-data.AFTER.SetupAndMiningButton.03-02-2025.15-52-06.json)

| before | after |
| --- | --- |
|||

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src/containers/main/Dashboard/MiningView/components/BlockHeightAccent.styles.ts
```diff
@@ -24,9 +24,11 @@ export const AccentText = styled(m.div)`
     color: ${({ theme }) => (theme.mode === 'dark' ? 'rgba(255, 255, 255, 0.095)' : 'rgba(255, 255, 255, 0.4)')};
     user-select: none;
     height: min-content;
+    rotate: -90deg;
+    //font-size: 16px;
```
MCozhusheck, line 28: Remove commented line

MCozhusheck (approved):
Found no memory leak.
Tried switching visual mode on and off again but found no bug.
Also notice small memory usage from 85 MB to 79MB while mining.

Also should we now on use `useMotionValue` or `useTramsform` with every component ?

src/containers/main/Dashboard/MiningView/components/BlockHeightAccent.tsx
```diff
@@ -8,58 +8,52 @@ export function BlockHeightAccent() {
...
 
-    useLayoutEffect(() => {
+    useEffect(() => {
         function handleResize() {
-            setWindowHeight(window.innerHeight - 80);
-            setWindowWidth(window.innerWidth);
+            windowHeight.set(window.innerHeight);
+            windowWidth.set(window.innerWidth);
         }
         window.addEventListener('resize', handleResize);
```
mmrrnn, line 23: Don't we need a debounce here?

mmrrnn:
LGTM

src/containers/main/Dashboard/MiningView/components/BlockHeightAccent.tsx
shanimal08: i tried adding a debounce and it stopped the scaling :( 
not sure if motion's `.set`/`useMotionValue` handles it internally or something. i tried digging through the docs a bit more and saw there are checks for it it's already animating but couldn't really see any more details 😅

PR #1466, opened 2025-02-03
fix(ui): updated app version design on setup screen

peps:
This PR updates the design of the app version on the setup screen. I positioned it to the right since it didn't look balanced in the center. I'm also only rendering the app version pill if the airdrop opt in box is not showing, before they were overlapping each-other. 

CleanShot 2025-02-03 at 11 46 19@2x

CleanShot 2025-02-03 at 11 45 20@2x


shanimal08 (approved):

Tas4tari:
Tested successfully on Windows on app version 0.9.6
Screenshot (103)

Issue #1395, opened 2025-01-17
0.8.44: MAC: When I download the app and run through installation, I find that whilst block syncing it reverted to the Universe lander and stayed there


Tas4tari:
**Describe the bug**
When I download the app and run through installation, I find that whilst block syncing it reverted to the Universe lander and stayed there. I did not run the data import with the data.mdb

**To Reproduce**
1. Download the dmg file for 0.8.44
2. Save to Applications
3. Open Tari Universe app
4. Once syncing of headers and blocks start
5. Observe the app
6. After some time the app reverts to the landing page

**Expected behavior**
Headers and blocks should sync successfully

**Screenshots**
<img width="1492" alt="Image" src="https://github.com/user-attachments/assets/0bb3c5b4-ea6c-4b3a-abd4-5b930d23b084" />

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11, 24H2

**Additional context**
Also additional context here: https://nomatter.tari.com/tari/pl/9jfgcmqr6jnadxtq6b67rwxsoc

[com.tari.universe.zip](https://github.com/user-attachments/files/18451221/com.tari.universe.zip)

Anziskavds:
This is also happening on a MacBook Air M1 2020 13 inch on Sequoia 15.2 on v0.8.44.

brianp:
Feels related, if not the same as https://github.com/tari-project/universe/issues/1424#issue-2804107316. Closing the other as duplicate as it's slightly more vague.

brianp:
Should be fixed on #1474

PR #1474, opened 2025-02-05
fix: trigger the frontend_ready event

mmrrnn:
#1395 

Description
---
Instead of relying on the Tauri app_ready event, we set up a listener and notify the backend that setup is complete.

Motivation and Context
---
Sometimes, Tauri triggers `app_ready` event before we setup corresponding listener on the frontend. This led to an issue when we haven't closed splashscreen at all.

How Has This Been Tested?
---
Use MacOS fresh install, observe whether splashscreen is closed.


src-tauri/src/main.rs
```diff
@@ -1214,31 +1209,6 @@ fn main() {
                 }
             }
         })
-        .on_page_load(|webview, _ | {
```
brianp, line 1217: Bummer. I thought `on_page_load` was great. How come this fails to work as expected?

mmrrnn:
A fresh installation has an 80% probability of success on my Intel Mac. However, this shouldn't be the only case.

PR #1488, opened 2025-02-06
fix(ui): setup screen performance optimization

peps:
This PR optimizes the Setup screen animations and markup. The goal is to reduce the memory impact of the Setup screen by replacing `motion/react` with native CSS animation ( without much loss in polish / quality ) 

Before: 400MB to 1000MB 

Now: 370MB to 530MB

The memory seems to be a lot more stable after these updates, most of the time stays in the 400MB range. 

- Replace `motion/react` animations with CSS animation where possible
- Rebuild steps navigation to use plain CSS, no longer uses `LinearProgress` component
- Text animation on the steps now animates by word instead of by character
- Change background "TARI UNIVERSE" text elements to use base tags instead of multiple `styled-components`
- Various Small animation tweaks and layout changes to add polish

CleanShot 2025-02-06 at 12 02 10@2x

Note: i'm still using `motion/react` just for simple fade out exit `AnimatePresence` animations, if we remove those there will be even more performance gain... Exit animations are tricky with CSS and would require a more drastic restructure... maybe in the future.


peps:
@brianp once/if this gets merged, ill bring it into the `parallax` branch, it should give more memory headroom for the effect.

shanimal08 (approved):
muuuch better ⭐ and the test/word animations are honestly better like this imo

brianp (approved):
Mate this is great. I think that info nav rebuild is making a world of difference once we move to the main screen 👏🏻

PR #1493, opened 2025-02-07
fix: block wind sound and broken animation

MCozhusheck:
Description
---
Fix bug introduced from merging #372 
Adds CSP for inline styling to allow rendering tower animation.
Also changes loading audio file to bundle it with FE assets from `public/assets` instead of bundling it from BE and exposing for FE  to fetch it.

Unfortunately this sound effect still doesn't work on Linux. This is caused by webgtk which relies on [GSteamer](https://gstreamer.freedesktop.org/documentation/index.html?gi-language=c) to stream media files. There are known issues with this and requires more research.

Motivation and Context
---
Fix P0 bug and simplify handling of sound effect.

How Has This Been Tested?
---
Few things to test:
 - sound effect when winning block mine
 - sound effect when replay won block
 - disabling and enabling audio from general settings

All above should be done on release mode since previously this issues didn't occur on dev build.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


MCozhusheck:
Tested on windows and ubuntu:
 - everything works on windows
 - on ubuntu there is on sound on release build

Misieq01:
Tested on MacOS and it works fine

MCozhusheck:
Just noticed the splash screen is no longer displayed. Until it's fixed I am changing this PR status to drat.

PR #1499, opened 2025-02-10
fix: cpu_miner deadlock

mmrrnn:
Description
---
1. Using separate scoped blocks for operations that require locks
2. Getting all configuration values upfront before starting operations
3. Minimizing the duration of write locks
4. Dropping locks explicitly when no longer needed

This should prevent deadlocks by ensuring that locks are held for the minimum time necessary and are properly released before other operations that might need them are started.

Testing
---
I can't reproduce this bug locally. App gets stuck at 100% setup screen for some users

## Summary by CodeRabbit

- **Refactor**
  - Enhanced error handling and retry logic for reading and writing shared state variables during mining startup.
  - Improved lock management within the CPU miner methods to ensure better structure and safety.
  - Adjusted the setup flow to mark completion without waiting for mining to finish, ensuring a more responsive user experience.
  - Introduced new utility functions for retrying lock acquisitions to enhance application stability.


brianp (approved):
> This should prevent deadlocks by ensuring that locks are held for the minimum time necessary and are properly released before other operations that might need them are started.

I think it's much tidier than before, and the scope & borrow duration changes should really help. It still becomes a bit of a probability game though. 

Could some of these reads still benefit from `try_read` with graceful handling such as a retry. At either the read, or parent function level.

MCozhusheck (approved):

Issue #1514, opened 2025-02-12
Feature: Add tiered win sounds


brianp:
**Describe the bug**
Intended behavior:
 - The notification sound should always play when a win is detected to build suspense.
 - The Level 1, 2, and 3 sounds should correspond to their respective animation win levels and play a moment later, in sync with the animations.
 - The complete block win sound feature should be hidden behind a feature flag, including the on/off switch

Sounds file to be shared via DM

MCozhusheck:
If we replay block win sound then we shouldn't play notification before the level 1,2 or 3 win sound right ?

brianp:
Closing as stale. Pool mining has changed the way this would work now.

PR #1532, opened 2025-02-17
feat: add block win audio base on tiers

MCozhusheck:
Description
---
Implements #1514 
Add multiple audio sources to the app that plays when user wins block or replays one from his wallet history. There is also notification sound that precedes before the main block win sound plays.

Motivation and Context
---
Give users audio feedback when new block is mined.

How Has This Been Tested?
---
build new `glApp.js` from this branch https://github.com/shanimal08/TARI_MINER_APP_VISUAL/pull/9
replace the `glApp.js` in `public/assets/glApp.js`

test sound when:
 - actually mined block (It is easier to do on esmeralda)
 - replay some win animation in wallet history

---
same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


MCozhusheck:
invalid branch
this is the [correct PR](https://github.com/tari-project/universe/pull/1533)

PR #1533, opened 2025-02-17
feat: add block win audio base on tiers

MCozhusheck:
Description
---
Implements #1514 
Add multiple audio sources to the app that plays when user wins block or replays one from his wallet history. There is also notification sound that precedes before the main block win sound plays.

Motivation and Context
---
Give users audio feedback when new block is mined.

How Has This Been Tested?
---
build new `glApp.js` from this branch https://github.com/shanimal08/TARI_MINER_APP_VISUAL/pull/9
replace the `glApp.js` in `public/assets/glApp.js`

test sound when:
 - actually mined block (It is easier to do on esmeralda)
 - replay some win animation in wallet history

---
same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp:
I am confused about the testing steps. If a different `gpApp.js` is needed, can't it be included in this branch?

brianp:
While testing i noted there are 4 sounds
- Notification
- Tier 1-3

On block win we get a notification, and tier sound.
On replay, we only get tier sound.

This is a feature, and I verified it is by design, as a replay isn't notification-worthy. Notification sounds only occur during a real block win.

We also confirmed the tiers for win noises:
```
    const humanValue = earnings / 1_000_000;
    if (humanValue < 100) {
        return 'success';
    }
    if (humanValue <= 1000) {
        return 'success2';
    }
    return 'success3';
```

brianp:
@metalaureate and I chatted regarding the win tiers and have decided to change the tiers to align with current p2pool win rewards:

@metalaureate:
> 10th,3.61
25th,7.64
50th,25.94
90th,265.52
95th,803.16
99th,2328.81
99.9th,13732.2

@brianp 
> 90th percentile would achieve tier 2
99th+ tier 3

@metalaureate 
> so maybe
10, 50, 1000

> Basic Animation: 1 – 25.94 (1st to 50th percentile)
Cool Animation: 25.94 – 803.16 (50th to 95th percentile)
Epic Animation: 803.16+ (95th to 99.9th percentile

@brianp 
> can I round
26 - 804

@metalaureate 
> ofc

brianp:
PR failed acceptance testing. The audio and animations are not closely enough aligned.

@shanimal08 I'll probably need you to come back around to this. Although I'm deeming it lower priority than the tx send/recv stuff.

Issue #1546, opened 2025-02-19
Update dialog is never displayed when `auto_update` disabled


mmrrnn:
**Describe the bug**
When `auto_update` disabled, update dialog is not displayed at all.

**To Reproduce**
1. Edit `app_config.json` file to disable `auto_update` flag(either manually editing the file or using Tari Universe settings)
2. Use older version of the app
3. Update dialog is not shown even though there is an update available 

**Expected behavior**
When `auto_update` disabled, update dialog should be displayed and ask whether user wants to update the app to the latest version.

**Desktop (please complete the following information):**
 - Ubuntu 24.04.1 LTS
 - Windows 10

![Image](https://github.com/user-attachments/assets/3f9fdb25-1994-4d8e-9a65-3c1f269a4ea4)

brianp:
Closed with #1547

PR #1547, opened 2025-02-19
fix: missing update dialog when auto update disabled

mmrrnn:
#1546

Description
---
When `auto_update` is disabled, the update dialog did not appear during the initial app launch. This issue arose because the `AutoUpdateDialog` was setting up listeners for the `updates_event `after the event had already been emitted from the backend.

**Solution**: I resolved this issue by moving the `initialize_periodic_updates` function to the `frontend_ready` command. This change ensures that the frontend is already prepared to handle the event appropriately.

Testing
---
Run the app when newer version is available and `auto_update` is disabled. You should see the update dialog:
image

## Summary by CodeRabbit

This release refines how the app manages updates for a smoother experience. Key changes include:

- **Refactor**
  - Enhanced the update process with improved error feedback to better inform users if issues arise.
  
- **Chores**
  - Removed the automatic background update checks at startup, streamlining the launch process.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -208,6 +208,14 @@ pub async fn frontend_ready(app: tauri::AppHandle) {
             .emit("app_ready", setup_complete_value)
             .expect("Could not emit event 'app_ready'");
 
+        if let Err(e) = state
+            .updates_manager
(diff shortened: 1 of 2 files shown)


MCozhusheck (approved):
Tested and it worked well

PR #1549, opened 2025-02-20
fix: shutdown deadlock

### LINKED ISSUE #1297: 0.8.29 on app close, shutdown spinner of death
**Describe the bug**
Mac 0.8.29, closing the app shows the shutdown screen that hangs mid-spinner animation and cursor turns into beachball. App becomes unresponsive and must be force closed.

MCozhusheck:
Description
---
Fixes #1297 
Use Tokio's TaskTracker to have an easy way of tracking running tokio's tasks. This also greatly simplifies awaiting shutdown of all running tasks by simply calling `TASKS_TRACKER.wait().await;`.
Also this adds [tokio-console](https://github.com/tokio-rs/console) to helpdebug any runtime tasks failures.

Motivation and Context
---
Some users experience deadlock while shutting down TU. This was caused by deadlock when in parallel we were `stop_all_processes()` and also every process was stopping which was caused by `app_shutdown` signal being triggered.

This PR changes it so we only rely on `app_shutdown` to stop every process.

How Has This Been Tested?
---
Run TU with rustflag to enable tokio-console `RUSTFLAGS="--cfg tokio_unstable" cargo tauri dev` and launch [tokio-console](https://github.com/tokio-rs/console) in different terminal
Trigger shutdown in different scenarios like:
 - mining
 - app initialization
 - reset setting
 - import seeds words

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Chores**
  - Updated package version to `0.9.823` across various configuration files.
  - Updated backend dependencies for improved stability.
- **Refactor**
  - Centralized asynchronous task management to boost overall responsiveness.
  - Enhanced shutdown handling and telemetry collection for a smoother operating experience.
  - Simplified process management by removing several methods across multiple components.
- **Bug Fixes**
  - Improved error handling in the download utility for better clarity in reporting issues.


mmrrnn (approved):
Great work, it looks fantastic :rocket:

NOTE: Not thoroughly tested on Ubuntu

src-tauri/src/utils/shutdown_utils.rs
```diff
@@ -20,107 +20,17 @@
...
-        tor_manager.stop().await.map_err(|e| e.to_string())?;
-    }
-
-    if should_shutdown && !state.shutdown.is_triggered() {
-        state.shutdown.clone().trigger();
-    }
-
-    Ok(())
+    state.shutdown.clone().trigger();
+    TASKS_TRACKER.close();
```
stringhandler, line 34: I don't like this naming. I suggest we wrap the static in a struct

src-tauri/src/telemetry_service.rs
```diff
@@ -115,33 +115,36 @@ impl TelemetryService {
...
-                                        system_info.clone(),
-                                        app_id.clone(),
-                                    ))
-                                },
-                                3,
-                                2,
-                                "send_telemetry_data",
-                            )
-                            .await);
+            loop {
```
brianp, line 118: Why did we need to introduce the `loop`?

MCozhusheck: Sorry I introduced this loop by mistake. Removing it right away.

brianp (changes requested):
I don't normally get stuck on shutdown, but this branch freezes while trying to shutdown ~~p2pool~~ every single time. 

Crossed out p2pool. It's not always the same process, but is always the same result. It has not successfully shut down.

MCozhusheck:
This PR was reverted as it caused on macOS to kill TU after waking up from sleep mode.

There are still some issues with killing processes. When app tries to kill and resume processes in short period of time we end up with `minotari_node` failing to spawn process watcher. Simple solutions as adding retries ends up in deadlock.

Issue #1510, opened 2025-02-12
NextNet 0.9.801: Windows 11: Import wallet: When I click on Import wallet, the app closes and does not re-launch on it's own


Tas4tari:
**Describe the bug**
As a user I find when I go to import a new wallet within the app, when I click on Import wallet, the app closes but does no relaunch on its own. After restarting the app, I can confirm that the wallet import was succesful

**To Reproduce
Steps to reproduce the behavior:**

1. Go to Settings
2. Go to Wallet
3. Click on the edit icon
4. Remove current seed words
5. Paste in a different set of seed words
6. Click on the check mark
7. Observe

**Expected behavior**
Wallet should be imported successfully, and app should relaunch automatically

**Desktop (please complete the following information):**
- Aspire A315-59
- Windows 11 24H2

Tas4tari:
Tested on Windows 11 and app version 0.9.805 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.812 and this has been resolved

Issue #1479, opened 2025-02-06
v0.9.6  “Use dynamic fail data”


MarkusMeyer1:
If I activate “Use dynamic fail data”, it shutdown universe instead of restarting it and I need to open universe manually

mac m2

mmrrnn:
Related [#1338](https://github.com/tari-project/universe/issues/1338)

mmrrnn:
I haven't found any problem with restarting the app after settings change(tested on intel's mac, m2 max mac, win 10, win11, linux).

I can't reproduce it anyhow. I'm moving this to on hold and wait until we notice something that could help me

Tas4tari:
Tested on Windows 11, app version 0.9.812 and did not see this issue occur

Issue #1463, opened 2025-01-31
0.9.1: Windows 11: Switch Tor off: App does not restart on it's own


Tas4tari:
**Describe the bug**
As a user I find that when I switch Tor off and click on Restart Now, the app does not re-launch on it's own

**To Reproduce**
1. Click on `Settings`
2. Go to and click on `Experimental Settings`
4. Scroll down to 'Tor'
5. Disable the `Tor` toggle and click on `Save`
6. Click on `Restart Now`
7. Observe

**Expected behavior**
When user clicks on `Restart `Now` app should re-launch on it's own

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11, 24H2

brianp:
is this not a duplicate of https://github.com/tari-project/universe/issues/819 ?

Tas4tari:
No, with Tor, the app does launch on it's own, however it never passes 48% during the inuitialization phase

Tas4tari:
Tested on Windows 11, app version 0.9.812 and this has been resolved

Issue #1338, opened 2025-01-02
0.8.38: MacBook Pro I5 & Windows 11: Reset Settings: App did not restart automatically and logs did not clear


Tas4tari:
**Describe the bug**
As a user I find when I clicked on `Reset Settings`, the app closed but did not auto restart. When I checkedd the logs, I confirmed they had not cleared

**Additional Information**
On Windows 11, my logs do clear however my app does not re-launch. I have to manually start the app

**To Reproduce**
**Pre-conditions:**
- User has mining app open

**Steps:**
1. Go to 'tari universe'
2. Click on 'Settings' icon
3. Scroll down to 'General' Settings option
4. Scroll down to `Reset Settings`
5. Click on `Reset Settings`
6. Ensure `Reset wallet` toggle is either disabled or enabled on modal
7. Click on `Yes` within the reset modal
8. Observe
9. On your laptop, go to: Users/[user name]/Library/Logs/com.tari.universe
10. Open the logs and view `Date modified` which shows you the date and time

**Expected behavior**
App should auto restart and logs should be cleaared

**Screenshots**
<img width="1508" alt="Screenshot 2025-01-02 at 6 44 24 PM" src="https://github.com/user-attachments/assets/16722b28-7dbc-478f-83d3-1f13c0974714" />

**Desktop (please complete the following information):**
 - MacBook Pro  Dual-Core Intel Core i5
 - Ventura 13.7.1

**Additional context**

Tas4tari:
Tested on Windows 11, app version 0.8.40 and when I confirmed restart, the app closed but did not automatically restart

Tas4tari:
Tested on MacBook Pro i5 on app version 0.8.40 and this is still occurring
<img width="1032" alt="Screenshot 2025-01-08 at 1 16 41 PM" src="https://github.com/user-attachments/assets/b6e6d9c0-3fa1-4aff-a8e3-cf719445e586" />
<img width="1032" alt="Screenshot 2025-01-08 at 1 16 18 PM" src="https://github.com/user-attachments/assets/045a1a12-870e-4a4a-838a-ad239f950dc3" />

Tas4tari:
Tested in MacBook Pro i5, app version 0.8.41 and can confirm that the app does restart automatically, however the log files did not clear out

Tas4tari:
Tested in MacBook Pro i5, app version 0.8.42 and the app did not restart on it's own and the logs did not clear
[...]

PR #1551, opened 2025-02-21
fix: prevent app exit on restart

mmrrnn:
Fix for:
#1279
#1330
#1510
#1463
#1510
#1479
#1338

Description
---
This pull request addresses a race condition encountered in the `App::restart()` method of the Tauri framework, which prevents the application from restarting properly after some actions(update, settings change).
The issue arises when `RunEvent::ExitRequested` is triggered immediately before `process::restart()` is called, leading to the application exiting before the new process can fully spawn.

Occurs on both Windows and macOS, with uncertainty regarding its presence on Ubuntu. This is particularly problematic due to the `tauri_plugin_single_instance`

Testing
---
The problem is difficult to reproduce purposefully, which requires constant observation in the future.

---
Based on: https://github.com/tauri-apps/tauri/issues/11392

## Summary by CodeRabbit

- **Refactor**
  - Enhanced error reporting for file downloads to provide more consistent warning messages.
  - Improved window focus behavior on launch and refined exit handling, ensuring a smoother restart process when needed.
- **Chores**
  - Removed unnecessary debug output for cleaner application logging.
- **New Features**
  - Introduced logging adjustments to focus on critical log levels for better monitoring.
  - Added restart logic to manage application restarts more effectively.


src-tauri/src/main.rs
```diff
@@ -1335,21 +1345,28 @@ fn main() {
                     .inspect_err(|e| error!(target: LOG_TARGET, "Could not setup app: {:?}", e));
             });
         }
-        tauri::RunEvent::ExitRequested { api: _, .. } => {
-            info!(target: LOG_TARGET, "App shutdown request caught");
+        tauri::RunEvent::ExitRequested { api: _, code, .. } => {
+            info!(target: LOG_TARGET, "App shutdown request caught with code: {:#?}", code);
+            if code.unwrap() == RESTART_EXIT_CODE {
```
MCozhusheck, line 1350: linter probably doesn't like this `unwrap()`. Replace it with `if let Some(code) = code { ... }`

src-tauri/src/main.rs
```diff
@@ -1335,21 +1345,28 @@ fn main() {
...
+                is_restart_requested.store(true, Ordering::SeqCst);
+            }
             let _unused = block_on(stop_all_processes(app_handle.clone(), true));
             info!(target: LOG_TARGET, "App shutdown complete");
         }
         tauri::RunEvent::Exit => {
+            if is_restart_requested_clone.load(Ordering::SeqCst) {
+                app_handle.cleanup_before_exit();
+                let env = app_handle.env();
+                tauri::process::restart(&env); // this will call exit(0) so we'll not return to the event loop
```
MCozhusheck, line 1361: Doesn't this mess with `stop_all_processes(...)` few lines below ? Honestly I have no idea how this affect tokio runtime and if this is a problem during clean up.

MCozhusheck (changes requested):
If I change event to something like this:
```rust
        tauri::RunEvent::Exit => {
            app_handle.cleanup_before_exit();
            let env = app_handle.env();
            tauri::process::restart(&env); // this will call exit(0) so we'll not return to the event loop
            info!(target: LOG_TARGET, "App shutdown caught");
            let _unused = block_on(stop_all_processes(app_handle.clone(), true));
            info!(target: LOG_TARGET, "Tari Universe v{} shut down successfully", app_handle.package_info().version);
        }
```
        I get a warning that everything after `tauri::process::restart(&env);` is unreachable. Copy  `let _unused = block_on(stop_all_processes(app_handle.clone(), true));` and put there too so it is always called.

src-tauri/src/main.rs
```diff
@@ -1069,10 +1069,15 @@ fn main() {
...
-                );
+            match app.get_webview_window("main") {
+                Some(w) => {
+                    let _unused = w.show().map_err(|err| error!(target: LOG_TARGET, "Couldn't show the main window {:?}", err));
+                    let _unused = w.set_focus();
+                },
+                None => {
+                    error!(target: LOG_TARGET, "Could not find main window");
+                }
+            };
```
brianp, line 1080: What's the benefit of doing this focus?

brianp: And what was the single-instance event for previously?

PR #1565, opened 2025-02-25
fix: await killed processes on unix systems

MCozhusheck:
Description
---
Call `waitpid` to get confirmation that `SIGTERM` signal did indeed terminate process.

Motivation and Context
---
Previously, having left out minotari_node process caused next TU launch to drop database - this happened because we didn't await previous process to be closed and instantly spawning next instance which failed since it could get lock on the database. This caused minotari node to mark DB as corrupted and drop it forcing user to sync every block again.

How Has This Been Tested?
---
Launch app and close it with hanging processes (on unix use `kill -9 pid` or on windows use task manager to kill a process or with dev build `cargo tauri dev` just refresh project by saving file)

Launch app again and check logs for errors that looks like this:
```
2025-02-24 16:05:59.102121316 [c::cs::lmdb_db::lmdb_db] [,] ERROR Could not acquire exclusive write lock on database lock file: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" } // base_layer/core/src/chain_storage/lmdb_db/lmdb_db.rs:1852
2025-02-24 16:05:59.102616030 [minotari::base_node::app] [,] ERROR Exiting with code (114): ExitError { exit_code: DatabaseError, details: Some("Cannot acquire exclusive file lock, another instance of the application is already running") } // applications/minotari_node/src/main.rs:96
```
Also verify if it's resyncing whole blockchain (but there could be different causes for this not solved by this PR)

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced process termination handling to ensure processes close reliably, contributing to improved system stability and smoother operations.

### DIFF
--- a/src-tauri/src/process_killer.rs
+++ b/src-tauri/src/process_killer.rs
@@ -21,6 +21,8 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use anyhow::Result;
+use nix::sys::wait::{waitpid, WaitStatus};
+use tokio::task;
 
 pub async fn kill_process(pid: i32) -> Result<(), anyhow::Error> {
     #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


MCozhusheck:
closing since there were major changes to the core logic of handling spawned processes.

Issue #1595, opened 2025-03-03
Application Stalls on shutdown


brianp:
**Describe the bug**
When closing the application via `cmd + q` or using the `X` on the window, the application goes to the "Shutting down" screen and freezes. The app never quits and eventually being force quit.

This does not exist on the 0.9.811 build, and is related to commit `333aa88a` when the block tower animation fixes were merged.

**To Reproduce**
1. Start the app
2. Allow mining to start
3. Close the app

**Expected behavior**
The app should close

**Desktop (please complete the following information):**
 - OS & Version: Mac OS 15.1.1

PR #1596, opened 2025-03-03
fix(ui): app stuck on shut down

shanimal08:
Description
---

in #1457 i wanted the `isAppReady` & `isShutting` values for styling changes, so just called the hooks again, but this caused them to be mounted twice. the shut down hook includes the adds a `preventDefault` on `closeRequested`, _then_ calls our `exit_application` command so things froze up with it happening twice (including the state var having already been set to true)

- just removed the second references to the hooks in `CurrentAppSection` (the original place they were used) and passed down the values as props

Motivation and Context
---

- resolves https://github.com/tari-project/universe/issues/1595

How Has This Been Tested?
---

locally:

https://github.com/user-attachments/assets/bf1d0193-ad0b-4716-b9f6-c1b60a5b10ad

---
- same steps as ticket but it actually shuts down 

## Summary by CodeRabbit

- **Refactor**
	- Updated the application's state handling so that operational statuses (e.g., readiness and shutdown signals) are now provided by parent components instead of being determined internally. This change streamlines control flow, ensuring components receive accurate status updates.

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -17,9 +17,13 @@ import ThemeProvider from '../theme/ThemeProvider.tsx';
 import { useIsAppReady } from '@app/hooks/app/isAppReady.ts';
 import Splashscreen from '@app/containers/phase/Splashscreen/Splashscreen.tsx';
 
-const CurrentAppSection = memo(function CurrentAppSection() {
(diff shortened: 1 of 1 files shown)


brianp (approved):
Tested on MacOS 👍🏻

PR #1617, opened 2025-03-06
fix: wrong installation folder not appearing

### LINKED ISSUE #1609: Warning for wrong install location doesn't appear
**Describe the bug**
Previously, if the application on MacOS isn't installed in the `Applications` directory we had a warning that popped up on the setup screen and prevented progress. The warning stated the app has to be installed in the correct location to run. And forced the user to quit.

Currently the setup screen stalls on 0% but no error message is displayed to the user.

**To Reproduce**
1. Move the app to any directory except Applications
2. Run the app
3. Get stuck on 0%

**Expected behavior**
A modal appears informing the user of the installation location need, and forces them to close the app.

Misieq01:
### [ Summary ]
- Dialog was not appearing because frontend wasn't ready to receive event about it when it was send
- Added extra logic to wait for frontend ready inside setup function

### [ Thoughts ]
- Maybe we could just fire setup function inside `frontend_ready`  ?

## Summary by CodeRabbit

- **New Features**
  - Introduced a startup synchronization mechanism that ensures the application waits for the interface to be fully ready before proceeding, enhancing initialization reliability and delivering a smoother, more consistent user experience.
  - Added a new utility module for managing frontend readiness communication.
  - Implemented a new struct for broadcasting and managing the readiness state of the frontend.
  - Added methods for signaling and waiting on frontend readiness.


brianp (commented):
Something about this feels weird. It feels like we're fighting what should be a known sequence of events. 

Or if this style is the best way to ensure any thing we send to the front end is ready then we should apply it to all places we see this type of issue.

Misieq01:
@brianp I think this whole flow could use some revision on separate task.

As for applying it to all places, to this moment everything that should be executed only after frontend is ready to receive events we have put in `frontend_ready` method but for this case we need to block `setup_inner` as we don't want to proceed with setup if application is in wrong folder.
So the solution to it is either:
- doing something like I've already done here on PR
- Moving setup method execution from **tauri Ready event** to `frontend_ready` method

I've chosen the first one currently as its seems less prone to break something and I've prioritized fixing the issue here

src-tauri/src/utils/app_flow_utils.rs
```diff
@@ -0,0 +1,60 @@
...
+            receiver: Mutex::new(receiver),
+        }
+    }
+
+    pub fn set_ready(&self) {
+        self.sender.send(true).expect("Failed to send ready signal");
+    }
+
+    pub async fn wait_for_ready(&self) -> Result<(), tokio::sync::watch::error::RecvError> {
+        let mut receiver = self.receiver.lock().await;
```
mmrrnn, line 52: Hmm it's not a bad idea. I don't like it's single use due to the lock though. We won't be able to reuse it anywhere else

Misieq01: Yeah, but receiver must be mutable to use `wait_for`, but I can think about solution to make it reusable

Misieq01: Played around and we can use it multiple times in different places without problem

brianp:
An installation error now pops up as expected.

Weirdly, the updater can still run. But no other process or startup happens. So it think this is an okay edge case.

PR #1623, opened 2025-03-06
chore: add user data from payment-id & debugging logs

mmrrnn:
Description
---
* Extracted user data from `payment_id` to the ` TransactionInfo`
* Enabled command logs so we can easily debug the current issues
* Optionally pushed payment id as an arg


src-tauri/src/wallet_adapter.rs
```diff
@@ -125,7 +126,7 @@ impl WalletAdapter {
                 direction: tx.direction,
                 fee: tx.fee,
                 timestamp: tx.timestamp,
-                payment_id: tx.payment_id.to_hex(),
+                payment_id: PaymentId::from_bytes(&tx.payment_id).user_data_as_string(),
```
SWvheerden, line 129: Side note, I believe this is useless as is here. 
We probably need to change it on the base layer to be the user string. This is probably from what I can see going to be the encrypted bytes on the transaction

mmrrnn: @SWvheerden It works fine, I've extracted the `user_data` for inbound/outbound transactions. One small issue is that empty value is parsed to the "None" string. For other cases it works as expected

PR #1629, opened 2025-03-07
feat: add remote node proof of concept

stringhandler:
Proof of concept PR for adding a remote node implementation.

Requirements:
- adding a remote node address to your app config: `https://grpc.esmeralda.tari.com:443`
- development version of p2pool

Existing issues:
- MMProxy may need updating as it is marked as unhealthy


brianp:
> Looks like this needs a re-base? The CI was merged this morning.

This goes into a feature branch, not main. So the feature branch is what needs to be updated.

PR #1638, opened 2025-03-07
Feature txs qr code for receive

mmrrnn:
Description
---
Added QR code functionality to the Receive tab to facilitate easier wallet address sharing and transaction initiation. The QR can be scanned by Aurora Wallet.

Note: Not sure about the styling so I just used the max width

## Dependencies
- Added `qrcode.react` latest version  `"^4.2.0"`

## Screenshots
Screenshot from 2025-03-07 23-33-04
Screenshot from 2025-03-07 23-32-53

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
         "i18next-http-backend": "^3.0.2",
         "linkify-react": "^4.2.0",
         "motion": "^12.4.7",
+        "qrcode.react": "^4.2.0",
         "react": "^18.3.1",
         "react-dom": "^18.2.0",
         "react-hook-form": "^7.54.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
         "i18next-http-backend": "^3.0.2",
         "linkify-react": "^4.2.0",
         "motion": "^12.4.7",
+        "qrcode.react": "^4.2.0",
```
shanimal08, line 32: oop `react-qr-code` is already installed from the PaperWallet stuff, not sure which is better between the two, but can we just use one or the other? :D 🙏🏼

shanimal08 (approved):
thank you 💜

PR #1675, opened 2025-03-13
fix(ui): Made the Airdrop Link Right-click Copyable to Clipboard

audieleon:
Description

A few people are saying they can't click the link to the airdrop. Since this is so central to rewards and engagement, I thought it would be a good idea to provide a secondary way to get the link - a simple right click. Added a indicator that the link was copied that disappears after a couple seconds.

Motivation and Context

Found a few people in the discord unable to get the browser to open when the link to airdrops, and wanted to get them in the community and mining for gems as soon as possible.  

How Has This Been Tested?

I tested it locally on my mac. The fix should be independent of platform though. 

Build and start the app as a new user. Right-click on the "Join Airdrop" button. You should momentarily see "Copied" in the button, and then you can take the link from the clipboard and open it in 

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Enhancements to the claim action now let you copy a referral URL by right-clicking on the button.
  - The button provides dynamic visual feedback, indicating when the URL is being copied and when the process is complete.
  - New instructional messages for users regarding the copying action have been added to improve user experience.

- **Style**
  - The ClaimButton now includes conditional styles for error states, featuring a red border and a shake animation to indicate issues during the copying process.

### DIFF
--- a/public/locales/en/airdrop.json
+++ b/public/locales/en/airdrop.json
@@ -39,6 +39,7 @@
   "loggedInTitle": "Airdrop Rewards",
   "loggedOutTitle": "Mine & Invite Friends to Earn Gems",
   "copying": "Copying...",
+  "copied": "Copied!",
   "copyFailed": "Copy Failed",
   "rightClickToCopy": "Right-click to copy link",
   "joinAirdropButtonAriaLabel": "Join Airdrop and earn gems. Right-click to copy link"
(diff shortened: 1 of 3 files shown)


stringhandler:
Amazing! Thanks

shanimal08 (approved):
thanks so much, tested locally & LGTM!

just wanted to note two things (neither are merge blockers) 

- this only works if you click first, _then_ right click to copy. i don't think that's a problem though, since the normal click works for most users :) 
if we do want it to work just the copying off the bat - we can  use `setLinkOpened(true)` after a successful copy
- the shake on error is neat! just the error colour could use a theme colour

those can be taken care of that after this is merged though :D

brianp (approved):
Thanks for your contribution; we really appreciate it!

PR #1714, opened 2025-03-25
fix: airdrop refresh retry

shanimal08:
Description
---
- fixed the retry count reset logic (tokens never got to the reset point before the next retry would happen)
- slightly clearer log message

Motivation and Context
---

- logging out and in 3+ times would kill the auth flow since the retry counts got maxxed out with the 404 url/CORS error and never got to reset

How Has This Been Tested?
---

- by spamming log out and in locally (was reproducible locally before too)

https://github.com/user-attachments/assets/195ead0c-37c5-41b9-abb1-bdc4b3d007ca

## Summary by CodeRabbit

- **Style**
  - Introduced minor formatting improvements for enhanced code readability.
  
- **Bug Fixes**
  - Updated error messages to offer clearer, more descriptive feedback during token authentication and refresh.
  - Refined the retry logic for token refresh operations to ensure more consistent and reliable behavior.


brianp (approved):
The logic seems awkward. But it works.

Issue #1594, opened 2025-03-03
`0.9.811` Tower animation x block sync bug


shanimal08:
**Describe the bug**
Tower animation dies if there are lots of new block height evts coming in (e.g in the case of block heights not being synced)

**To Reproduce**
hard to reproduce, but if you start up and you're "on a fork" but the block height is catching up, observe the tower animation while mining 

**Screenshots**

![Image](https://github.com/user-attachments/assets/7bf58b8e-faac-4388-bce1-bf8131a41bae)

brianp:
<img width="693" alt="Image" src="https://github.com/user-attachments/assets/dac6f2b3-0da5-407a-afa9-d92e2260bfee" />

PR #1735, opened 2025-03-27
fix: animation block sync

MCozhusheck:
Description
---
Fixes: #1594 
Add debounce for `handleNewBlock` to add only latest block within 200ms.

Motivation and Context
---
Sometimes app syncs during animation is active which sends lots of `New Block Height` events which causes animation to crash.

By introducing debounce we can take only take latest value which is the only relevant.

How Has This Been Tested?
---
In dev build click `Add New Dummy Blocks` in admin panel while mining. Previously this would cause animation to crash but now it should work as expected.

---
same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive button that allows users to add dummy blocks, enhancing the blockchain visualization experience.

- **Refactor**
  - Optimized block processing by implementing a debouncing mechanism to ensure smooth and efficient updates during blockchain interactions.
  
- **Chores**
  - Expanded the list of ignored dependencies in the configuration to include `prettier-eslint` for CI purposes.

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -11,7 +11,7 @@ const config: KnipConfig = {
         duplicates: 'error',
     },
     ignoreBinaries: ['commitlint'],
-    ignoreDependencies: ['babel-plugin-styled-components'], // needed by plugins
+    ignoreDependencies: ['babel-plugin-styled-components', 'prettier-eslint'], // needed by plugins and prettier for CI
     ignoreExportsUsedInFile: true,
 };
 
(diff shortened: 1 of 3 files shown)


shanimal08 (approved):
nice, tested and LGTM! even spam clicked the admin button!

question though, shouldn't this PR point to `main` not the feature branch?

PR #1749, opened 2025-03-31
fix: better tor and mmproxy health monitoring and frontend event processing fix

stringhandler:
Intended to be rolled as a hotfix

* Added tor control port monitor, cherry picked from main branch.
* Added better mmproxy health check for `wait_ready`, previously this would just wait for 20 seconds.
* Tried to code some front end fixes for setup getting stuck and not processing events

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "tari-universe",
     "private": true,
-    "version": "0.9.830",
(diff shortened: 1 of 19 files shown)


brianp (approved):
Thanks for doing the core api updates!

PR #1765, opened 2025-04-03
fix: node_manage improvements

mmrrnn:
nodes management improvements & logging improvements & back to tx history when send tx & fix send tx & improve usage of rwlocks to prevent from deadlocks

Telemetry works correctly - data can be adjusted

### DIFF
--- a/src-tauri/src/node_manager.rs
+++ b/src-tauri/src/node_manager.rs
@@ -155,11 +155,11 @@ impl NodeManager {
         tor_control_port: Option<u16>,
         remote_grpc_address: Option<String>,
     ) -> Result<(), NodeManagerError> {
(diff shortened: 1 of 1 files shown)


src-tauri/src/local_node_adapter.rs
```diff
@@ -607,7 +610,12 @@ impl StatusMonitor for MinotariNodeStatusMonitor {
             Ok(res) => match res {
                 Ok(status) => {
                     let _res = self.status_broadcast.send(status.clone());
-                    if status.num_connections == 0 {
+                    // Remote Node always returns 0 connections
```
brianp, line 613: 🤔 any idea why?

brianp (approved):

PR #1775, opened 2025-04-07
feat(quick node): adapter refactor

mmrrnn:
Create a common trait for NodeAdapter, dynamically get current adapter, move some logic to the service


src-tauri/src/node/local_node_adapter.rs
```diff
@@ -0,0 +1,327 @@
...
+                let _unused = fs::remove_dir_all(peer_db_dir).inspect_err(|e| {
+                    warn!(target: LOG_TARGET, "Failed to remove peer db: {:?}", e);
+                });
+            }
+            info!(target: LOG_TARGET, "Node Migration v1 complete");
+            migration_info.version = 1;
+        }
+        migration_info.save(&migration_file)?;
+
+        // if is_first_start {
```
brianp, line 176: 🤔 ?

brianp (approved):
Why comment out the first run bit?

src-tauri/src/node/local_node_adapter.rs
mmrrnn: https://github.com/tari-project/universe/blob/main/src-tauri/src/node_adapter.rs#L143

I have no idea, we merge it from main

PR #1781, opened 2025-04-07
feat: rebuild ws handling

BalazsSevecsek:
Description
---

Rebuilds ws handling an puts many of the functionalities of ws into the rust side. Goal of the PR is to for us to get consistent data even if the application is minimized to the tray.

Motivation and Context
---

How Has This Been Tested?
---

Locally, it still needs the external server changes to be put into production first before it can be tested against the real environment.

---

Once the external changes hit production, it can be tested against DEV.

Breaking Changes
---

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

It needs to be tested against the old version as well. Everything must work the same (gems, referral, mining time bonus, proofs).

## Summary by CodeRabbit

- **New Features**
  - Introduced native WebSocket support using Tauri, enabling real-time communication for airdrop features.
  - Added commands to manually connect and disconnect the WebSocket from the app interface.
  - Implemented periodic mining status updates sent over the WebSocket.

- **Refactor**
  - Replaced the previous socket.io-client-based WebSocket system with a Tauri-native solution across the application.
  - Updated React hooks and event handlers to use the new WebSocket integration.

- **Chores**
  - Added and removed dependencies to support the new WebSocket implementation.
  - Cleaned up unused code and removed legacy socket.io-client logic.

- **Bug Fixes**
  - Improved event handling and message parsing for more reliable WebSocket communication.


brianp:
I can appreciate the work involved in moving the WS from the front end to the back end here. After a bit of technical review, I wonder if this is our opportunity to ditch websockets altogether. We weren't huge fans of the initial WS implementation and if we're looking to remove it, then let's actually remove it.

We would have less dependencies and more simplicity just to be making HTTPS requests against our data server. It's simple to fire and forget, lighterweight, and less prone to failure than trying to maintain WS connections. It's less lifting for the servers as well. Receiving simple requests is basic. Maintaining WS connections is error-prone. 

I think its worth the discussion to abandon the WS idea.

@stringhandler @BalazsSevecsek @PanchoBubble @metalaureate

PR #1754, opened 2025-04-01
feat(ui): reconnect redesign

MCozhusheck:
Description
---
Implements: #1703 

Screenshot from 2025-04-03 11-57-41
Screenshot from 2025-04-03 11-58-14
Screenshot from 2025-04-03 11-58-46

Motivation and Context
---

Make it more readable to the user if he gets disconnected and give more option to handle such situation.

How Has This Been Tested?
---
There is new button in Admin UI panel to switch between views with text `Change connection status`.
Screenshot from 2025-04-03 12-02-41

What process can a PR reviewer use to test or verify this change?
---
Use newly created button in dev mode.

Breaking Changes
---

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Introduced a reconnection flow with user interface for handling network disconnections, including retry logic and cooldown periods.
  - Added visual feedback and actions for both standard and severe disconnection scenarios, with options to reconnect, restart the application, or seek help via Telegram.
  - Enhanced localization with new translations for reconnection messages in multiple languages.

- **Improvements**
  - Dashboard now displays a dedicated disconnection overlay when network issues occur.
  - State management updated to track connection status and reconnection progress.
  - Event system extended to handle connection status updates and notify the frontend.
  - Added countdown and retry logic to manage reconnection attempts with backoff.
  - Connection status integrated into modal visibility logic to improve user experience.
  - Feature flag fetching logic refined for orphan chain UI handling.
  - Adjusted orphan chain UI feature flag logic to use a "disabled" flag instead of "enabled".

- **Style**
  - Added new styled components to support the reconnection UI.
  - Updated text color in floating components to respond to theme changes.

- **Chores**
  - Removed unused list component.

PR #1782, opened 2025-04-07
fix: handle error from resuming all processes

MCozhusheck:
Description
---
Add error handling if resuming all processes fails. This only hides endlessly spinning spinner, there should be better strategy if this issue persist for longer - maybe we should display same modal if orphan chain is detected like in #1754?

Motivation and Context
---
During testing #1754 I noticed that we only cover happy path in re-connection scenario and any failures result in hanging up loader.

How Has This Been Tested?
---
I have turned off wi-fi and put my machine in sleep mode and waked up moment later to trigger `resuming_all_processes`.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Refactor**
  - Streamlined how operations resume by separating core resumption logic from error management.

- **Bug Fixes**
  - Enhanced error handling to ensure issues during the resume process are properly logged and users are notified of any failures.
  - Improved event emission for successful and failed resumption of processes, providing detailed feedback.


MCozhusheck:
After quick load this is no longer relevant

PR #1785, opened 2025-04-08
feat: [ quick load ] add config per network

Misieq01:


mmrrnn:
Why not `https://grpc.{}.tari.com:443", Network::Network::get_current_or_user_setting_or_default().as_key_str()`?

brianp:
We should probably update mmproxy too https://github.com/tari-project/universe/blob/feature/quick-load/src-tauri/src/mm_proxy_adapter.rs#L165

brianp:
> is there the same pattern for the mmproxy address ? "[https://grpc-p2pool.nextnet.tari.com:443](https://grpc-p2pool.nextnet.tari.com/)" => "[https://grpc-p2pool.esmeralda.tari.com:443](https://grpc-p2pool.esmeralda.tari.com/)"

Yes that should work.

> Also we could provide local p2pool address once we switch to local node yes ?

Yes.

There's a bug assumption here, that the local base node will take longer to sync than p2pool will. So p2pool can switch to local when the base node is ready.

Misieq01:
@brianp https://grpc-p2pool.esmeralda.tari.com:443 does not work

mmrrnn:
We don't have a p2pool on esmeralda @brianp @Misieq01

PR #1796, opened 2025-04-08
feat(quick-load): show background node sync status

mmrrnn:
* Add local node syncing status in settings when synced in the background
Screenshot from 2025-04-09 09-15-56

* use note types as Remote and Local only

* fix displaying Block sync progress(local header was incorrect)


brianp:
Can we duplicate this information on the General screen for something like this:

image

PR #1844, opened 2025-04-14
fix(ui): better send form field error message ui

peps:
I've improved the error message UI for the Send form fields. The errors now animate and are part of the Input components. This provides more context to the user and looks much better visually. 

CleanShot 2025-04-14 at 18 09 43@2x

CleanShot 2025-04-14 at 18 15 06@2x


shanimal08 (approved):
just the lints but LGTM :D

PR #1865, opened 2025-04-16
fix(quick load): double check wallet scanned before grpc server

mmrrnn:
* wallet returns scanned_height as 0 when immediately scanned, before grpc service started. Double check this value when connected


brianp:
Should this also be a wallet fix in core? Should it not return 0 if scanning is complete before grpc is ready?

PR #1867, opened 2025-04-17
fix(quick load): history list - fix re-renders and styling | fix settings close issue

shanimal08:
Description
---
- memoized history list & wallet sidebar components
- don't update wallet scanning state if data from payload is the same
- adjusted history `ListItem` styling:
  - fixed alignment(for uncentred `NEW` chip)
  - fixed tx info styling 
- adjust settings modal styling so there's nothing under the heading bar & you don't accidentally enable/disable a setting underneath

Motivation and Context
---
- prevent re-renders
- fix settings close click issue
- ui feedback

How Has This Been Tested?
---

locally:

https://github.com/user-attachments/assets/480874aa-37d0-4c39-a210-dcc1d9b8102a

image

image


src/containers/navigation/components/Wallet/WalletBalanceMarkup.tsx
```diff
@@ -15,7 +15,7 @@ import { useTariBalance } from '@app/hooks/wallet/useTariBalance.ts';
 import { toggleHideWalletBalance } from '@app/store/actions/uiStoreActions.ts';
 import { useUIStore } from '@app/store';
 
-export default function WalletBalanceMarkup() {
+const WalletBalanceMarkup = memo(function WalletBalanceMarkup() {
```
brianp, line 18: If this value is memoized. Won't it never update again?

shanimal08: its children change so that counts as a prop update, and it was updating while i tested, but lemme double check :D

shanimal08: https://github.com/user-attachments/assets/e9f49635-ce56-4048-86c7-f17d0323a7c2

yep updating normally :D

brianp: Probably because JS doesn't know what memoize means

PR #1908, opened 2025-04-23
feat: phase restart

Misieq01:
### [ Summary ]
- Added method for deleting local files for each binary adapter
- Added mechanism for executing soft ( without file deletion ) and hard ( with file deletion ) restarts for each phase
- Implemented phase builder for customizing phase behaviour
- Added setup_adapter for keeping setup logic in one place instead writing it in each phase

### [ TODO ]
- Need to close all phase related processes on each restart not only on hard restart


Misieq01:
Most of the utilities for this PR were implemented on other ones and we moved to restarting specific parts of phase instead of restarting whole phase

Issue #1895, opened 2025-04-21
[Quick Load] - Bug: Transition between remote and local node will show the fork UX briefly while local node is finalizing


brianp:
## Description 

After the local node finishes a complete sync, it will sometimes display the Fork UX. Viewing the logs shows sometimes the node may be re-orging some blocks or finalzing the sync in some other way.

Logs and video to be added when re-captured

## Expectations

Users who have finally successfully synced their local node should not be shown a fork message, risking them to restart the process or making them believe they've synced the wrong chain. It's a very disheartening message to receive after all that time spent.

Ensure the local node is _fully prepared_ to be changed too.

## Solutions

If we don't have any other flag to go by than the `sync: true` and matching tip heights, then maybe we need to meet 3 criteria:

Only change to local node from remote when:
- sync is true
- tip heights match
- 30 seconds have passed where the first two criteria are true

PR #1913, opened 2025-04-23
fix: retry sync if local node is still 50 blocks behind remote node

MCozhusheck:
Description
---
Fixes: #1895 

Retries local node sync if it's at least 50 blocks behind remote node.

Motivation and Context
---
Syncing process doesn't updates tip height while syncing so if it takes a long time then right after syncing it could sync again.

How Has This Been Tested?
---
The best way is to have local node (preferably on nextnet) that is few hundred blocks behind the [textexplorer](https://textexplore-nextnet.tari.com/).
Run TU and check logs for sync progress and potential retries and fork chain warnings.

---
Same as above

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Improvements**
  - Enhanced node sync logic to more accurately track and report the local node's block height after syncing.
  - Improved reliability when switching node types by ensuring the local node is sufficiently up-to-date before switching.
  - Added clearer informational and warning messages during the sync and switching process.

- **Style**
  - Minor code cleanup for improved maintainability (no impact on user-facing features).

### DIFF
--- a/src-tauri/src/node/node_adapter.rs
+++ b/src-tauri/src/node/node_adapter.rs
@@ -178,38 +178,12 @@ impl NodeAdapterService {
             let tip_res = tip.into_inner();
             let sync_progress = sync_progress.into_inner();
             if tip_res.initial_sync_achieved {
-                if sync_progress.local_height >= sync_progress.tip_height {
-                    info!(target: LOG_TARGET, "Initial sync achieved and local height is equal or greater than tip height");
-                    let tip_height = match tip_res.metadata {
-                        Some(metadata) => metadata.best_block_height,
-                        None => 0,
-                    };
-                    return Ok(tip_height);
-                } else {
-                    info!(target: LOG_TARGET, "Initial sync achieved but local height is lower than tip height");
-                    // Report to sentry that we have initial sync achieved but local height is lower than tip height
-                    let error_msg =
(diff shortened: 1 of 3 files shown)


src-tauri/src/node/node_adapter.rs
```diff
@@ -174,35 +174,16 @@ impl NodeAdapterService {
                 .get_sync_progress(Empty {})
                 .await
                 .map_err(|e| NodeStatusMonitorError::UnknownError(e.into()))?;
+
             let tip_res = tip.into_inner();
             let sync_progress = sync_progress.into_inner();
             if tip_res.initial_sync_achieved {
-                if sync_progress.local_height >= sync_progress.tip_height {
```
MCozhusheck, line 180: Removed this whole `if {...} else {...}` statement since `sync_progress.local_height` and `sync_progress.tip_height` are always equal if node is not syncing anymore  (`tip_res.initial_sync_achieved` is true`) according to [get_sync_progress grpc call](https://github.com/tari-project/tari/blob/development/applications/minotari_node/src/grpc/base_node_grpc_server.rs#L2259)

MCozhusheck:
Putting this in draft since testing this takes a long time.

PR #1914, opened 2025-04-23
chore: check best_block_height before finishing node sync

mmrrnn:
Description
---
Save highest tip we sync to, compare it with the synced node metadata

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization logic to ensure more accurate detection of initial sync completion.


mmrrnn:
Closing, since it shouldn't make a difference when https://github.com/tari-project/universe/pull/1913 is merge

PR #1949, opened 2025-04-28
fix: node updates when setting up

mmrrnn:
Description
---
* Count down est. remaining startup time regardless of the speed of sync.
* Immediately apply node related values when node type changed(prevent from problematic nullable values)
* Fix changing node type during setup
* Prevent from some Setup failures
* Hide connected peers/miners connection to the network until setup is finished(it's not updated during the setup anyway)

## Summary by CodeRabbit

- **New Features**
  - Network and peers settings are now hidden while the app is still setting up, improving the setup experience.
  - Countdown timer for node synchronization now provides more accurate estimates and a clearer indication when estimates are unavailable.

- **Bug Fixes**
  - Node type updates are now consistently propagated across the app, ensuring UI and state remain in sync.

- **Improvements**
  - Synchronization progress countdown is more reliable and easier to understand.
  - Enhanced event handling for node type changes, leading to a more responsive UI during configuration changes.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -1784,12 +1784,13 @@ pub async fn set_node_type(
     ConfigCore::update_field_requires_restart(
         ConfigCoreContent::set_node_type,
         node_type.clone(),
-        vec![SetupPhase::Wallet, SetupPhase::Node, SetupPhase::Unknown],
+        vec![SetupPhase::Node, SetupPhase::Wallet, SetupPhase::Unknown],
     )
     .await
     .map_err(InvokeError::from_anyhow)?;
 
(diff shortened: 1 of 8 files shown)


brianp:
> Fix changing node type during setup 

This is interesting. It's fine for now. But if it were a public feature we'd want to allow this. Or to stop the current setup to switch.

mmrrnn:
> > Fix changing node type during setup 
> 
> This is interesting. It's fine for now. But if it were a public feature we'd want to allow this. Or to stop the current setup to switch.

It's allowed, just needed to handle some edge cases

brianp:
lol i think i read these two lines at the same time.

> Fix changing node type during setup
> Prevent from some Setup failures

and it became

> Prevent changing nodee type during setup

Issue #1995, opened 2025-05-01
Tor is has a static version to download


brianp:
**Describe the bug**
The binary version in the binary file is just the binary "we want to run" not the binary we actually fetch, because that would make wayyyyy too much sense

Fix [this file](https://github.com/tari-project/universe/blob/main/src-tauri/src/binaries/adapter_tor.rs) so it's less lame.

Issue #1573, opened 2025-02-26
Linux has high number of Tor restarts


stringhandler:
## Description

According to telemetry, Linux has a high number of tor restarts. Linux users in Telegram and Discord also report seeing tor restarts in the logs frequently.

## Screenshot from telemetry data

![Image](https://github.com/user-attachments/assets/cc5ba8f3-e0d8-4645-a1e0-a6e369940a36)

## Investigation

We need to figure out why TOR is restarting so often on Linux systems. If the issue is not reproducible on any of our development systems, we need to implement some additional logging to try and assist in future debugging. See what kind of information we can collect and send to Sentry or Telemetry as the restarts happen.

Misieq01:
The issue with constant tor restart was present on my machine.
After running tor manually from terminal:

`cd .cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/13.5.7/tor/`

`./tor --allow-missing-torrc --ignore-missing-torrc --clientonly 1 --socksport 44135 --controlport 127.0.0.1:38943 --clientuseipv6 1 --DataDirectory /home/bartek/.local/share/com.tari.universe.alpha/tor-data --Log notice file /home/bartek/.local/share/com.tari.universe.alpha/logs/tor.log`

Got following: `error while loading shared libraries: libevent-2.1.so.7: cannot open shared object file: No such file or directory`

Installed this package with `sudo apt install libevent-2.1-7`

Now when I open the tari app, tor is no longer restarting constantly

Can someone also fix this issue with following steps ?

ln021:
@Misieq01 
I'm getting the following error message:
```
$ cd ~/.cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/13.5.7/tor
$ ./tor --allow-missing-torrc --ignore-missing-torrc --clientonly 1 --socksport 44135 --controlport 127.0.0.1:38943 --clientuseipv6 1 --DataDirectory ~/.local/share/com.tari.universe.alpha/tor-data --Log notice file ~/.local/share/com.tari.universe.alpha/logs/tor.log
./tor: symbol lookup error: ./tor: undefined symbol: evutil_secure_rng_add_bytes
```
I'm running Arch Linux and `libevent 2.1.12-4` was already installed.

Misieq01:
Could you try updating libevent version to 2.1-7 ?

ln021:
> Could you try updating libevent version to 2.1-7 ?

[...]

PR #1997, opened 2025-05-02
Fix tor startup failures on Linux

bivsk:
Description
---
I am trying to implement a fix to the tor startup failures on Linux systems. However, my knowledge of rust is far from my understanding of Linux. This is only a crude implementation of a fix, demonstrating that it works. 

The issue is caused by an ABI break between the bundled `tor` and common distro builds of  `libevent`. See: https://github.com/tari-project/universe/issues/1573#issuecomment-2741814167

The goal of this PR is to _force_ the use of the bundled `libevent-2.1.so.7` when TU is being run on Linux.

What I _want_ to implement: 
 - Do nothing if we aren't on Linux.
 - Don't hardcode the path to the tor binary/libevent.so

Any help, pointers, etc, would be greatly appreciated.  I'm not very experienced with systems languages.

Lastly, I just noticed https://github.com/tari-project/universe/issues/1995 
I do not fully understand the context of the issue, but maybe fixing that will make my PR here unneeded. 

How Has This Been Tested?
---
Running `cargo tauri dev` results in a successful tor startup.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/tor_adapter.rs
+++ b/src-tauri/src/tor_adapter.rs
@@ -22,6 +22,7 @@
 
 use std::path::PathBuf;
 use std::time::Duration;
+use std::collections::HashMap;
 
 use anyhow::{anyhow, Error};
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


brianp:
@mmrrnn Please see if the recent tor upgrade resolves this issue, or if we should get this PR up to date and merge it for a fix.

bivsk:
Glad to hear! Happy to help any way I can.

I looked over the implementation, and I have a feeling the environment should be set regardless of the existence of a system-wide libevent.

I'll definitely help test to find out.

mmrrnn:
Why is that? It shouldn’t be the case for most Linux users.

Issue #1981, opened 2025-05-01
0.9.910: Mining Proof: When I search my App ID on Metabase I find that the Mining Proof query does not display any results


Tas4tari:
**Describe the bug**
As a user when I search up my App ID on Metabase, I find that there are no results displayed. I am connected to Airdrop.
`oBVxw0wHw296Ltoe5i3Y`
I also did a clean install of the build
https://mb.emoji.id/public/question/cf8e335e-d254-4d43-90cf-88bef6b09743?app_id=oBVxw0wHw296Ltoe5i3Y

I also checked:
https://rwa.y.at/admin/resources/MiningProof?sortBy=createdAt&direction=desc&page=1&filters.appId=oBVxw0wHw296Ltoe5i3Y

Note:
- After Resetting configuration including wallet, I checked again and still was unable to see any results
`d&24gA#EbBThkccajVQH`

**Screenshots**
![Image](https://github.com/user-attachments/assets/26134f2a-4676-4deb-b28f-7d5288549243)
![Image](https://github.com/user-attachments/assets/3ea6c07e-097d-43d0-8d3b-1eaf3b117618)

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11

Tas4tari:
Tested on Windows 11, app version 0.9.913 on a clean install:
MetaBase Mining proof check still produces no results
https://mb.emoji.id/public/question/cf8e335e-d254-4d43-90cf-88bef6b09743?app_id=HnlBFAl%5EN5JnoSGDzpK4
Also checked:
https://rwa.y.at/admin/resources/MiningProof?sortBy=createdAt&direction=desc&page=1&filters.appId=HnlBFAl%5EN5JnoSGDzpK4


App ID: `HnlBFAl^N5JnoSGDzpK4`

Tas4tari:
I did an update (not a clean install) of v0.9.914 and I found that after an update, airdrop was still logged n but I'm not receiving the mining proof on Metabase:
https://mb.emoji.id/public/question/cf8e335e-d254-4d43-90cf-88bef6b09743?app_id=VO9nONyI!LNwVS1i6H34
VO9nONyI!LNwVS1i6H34
However it is showing on 
https://rwa.y.at/admin/resources/MiningProof?page=1&filters.appId=VO9nONyI%21LNwVS1i6H34



![Image](https://github.com/user-attachments/assets/c365396b-8305-4a04-83fb-3b4db478595a)
[...]

PR #2000, opened 2025-05-02
fix: app_id is now obtained from config-core

PanchoBubble:
This commit removes the `app_id` field from the `WebsocketManager` and `WebsocketEventsManager` structs.

The `app_id` is no longer passed as a parameter during the instantiation of these managers. Instead, it is retrieved directly from the `ConfigCore` within the `connect_to_url` function in `WebsocketManager` and within the `start` function in `WebsocketEventsManager` when needed.

This change simplifies the codebase by reducing the number of parameters passed around and ensures that the managers always use the most up-to-date `app_id` from the configuration. And, more importantly, returns the real annon_id

## Summary by CodeRabbit

- **Refactor**
	- Simplified how anonymous IDs are managed for websocket connections by retrieving them internally when needed, rather than storing and passing them throughout the code. No changes to user-facing functionality.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -1077,18 +1077,14 @@ fn main() {
         node_manager.clone(),
     );
 
-    let app_id = app_config_raw.anon_id().to_string();
-
     let websocket_manager = Arc::new(RwLock::new(WebsocketManager::new(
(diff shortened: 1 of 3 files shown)


PanchoBubble:
https://github.com/tari-project/universe/issues/1981

PR #2016, opened 2025-05-05
feat(ui): update autoplay logic

shanimal08:
Description
---
- add config item for auto-playing video once

## Summary by CodeRabbit

- **New Features**
  - The video player in the Banner section now automatically expands on first view if the warmup has not been seen.
  - The app tracks whether the warmup has been seen, enabling a more personalized user experience.

- **Bug Fixes**
  - Improved synchronization between UI state and backend when updating the warmup seen status.

- **Chores**
  - Added support for storing and updating the warmup seen status in app configuration and backend commands.

### DIFF
--- a/src-tauri/src/configs/config_ui.rs
+++ b/src-tauri/src/configs/config_ui.rs
@@ -53,6 +53,7 @@ pub struct ConfigUIContent {
     sharing_enabled: bool,
     visual_mode: bool,
     show_experimental_settings: bool,
+    warmup_seen: Option<bool>,
 }
 
 impl Default for ConfigUIContent {
@@ -69,6 +70,7 @@ impl Default for ConfigUIContent {
             sharing_enabled: true,
             visual_mode: true,
             show_experimental_settings: false,
+            warmup_seen: None,
         }
     }
 }
@@ -161,6 +163,7 @@ impl ConfigImpl for ConfigUI {
                 sharing_enabled: old_config.sharing_enabled(),
                 visual_mode: old_config.visual_mode(),
                 show_experimental_settings: old_config.show_experimental_settings(),
+                warmup_seen: None,
             };
             let _unused = Self::_save_config(self.content.clone());
         } else {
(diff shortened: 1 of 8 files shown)


src/store/actions/appConfigStoreActions.ts
```diff
@@ -295,3 +295,12 @@ export const setNodeType = async (nodeType: NodeType) => {
         updateNodeTypeForNodeStore(nodeType);
     });
 };
+
+export const setWarmupSeens = (warmupSeen: boolean) => {
+    useConfigUIStore.setState({ warmup_seen: warmupSeen });
+    invoke('set_warmup_seen', { warmupSeen }).catch((e) => {
+        console.error('Could not set seen', e);
+        setError('Could not change seen');
+        useConfigUIStore.setState({ visual_mode: !warmupSeen });
```
mmrrnn, line 304: Shouldn't be a `warmup_seen` here? :imp:

shanimal08: lol :') thank you!!

src-tauri/src/configs/config_ui.rs
```diff
@@ -53,6 +53,7 @@ pub struct ConfigUIContent {
     sharing_enabled: bool,
     visual_mode: bool,
     show_experimental_settings: bool,
+    warmup_seen: Option<bool>,
```
mmrrnn, line 56: We should probably use bool?

src-tauri/src/configs/config_ui.rs
```diff
@@ -69,6 +70,7 @@ impl Default for ConfigUIContent {
             sharing_enabled: true,
             visual_mode: true,
             show_experimental_settings: false,
+            warmup_seen: None,
```
mmrrnn, line 73: and false by default

src-tauri/src/configs/config_ui.rs
shanimal08: i didn't wanna set it because of the old config part but yeah will just set to bool &  false ty:D

PR #2038, opened 2025-05-06
fix: use dynamic network for websocket manager

mmrrnn:
Description
---
network wasn't dynamic


mmrrnn:
Duplicate, already merged here: https://github.com/tari-project/universe/pull/2037

PR #2086, opened 2025-05-09
fix: clean peer db folder on each node run

mmrrnn:


brianp:
I'm going to approve and merge this. But it looks like were doing the same thing in at least two different ways. The node_adaper already has this code but commented out. We have the first run code, migration code, and the node_manager has functions to also do these things.

We need to clean it all up, and settle on a single way to achieve this.

brianp:
This PR was reverted. As mentioned it happens in multiple places and causes race conditions. When other branched couldn't find the directory the setup failed to run. [3c0a04c738d0cdb3f692a50114d14f116dc77c10](https://github.com/tari-project/universe/commit/3c0a04c738d0cdb3f692a50114d14f116dc77c10)

PR #2095, opened 2025-05-12
feat(ui): mini block explorer

peps:
WIP


brianp:
@peps branch won't start

image


Lots of other warnings too:

image

shanimal08:
@brianp did you npm ci? :P

i'm testing locally and it looks goood, didn't see any of those warnings in console? 
the only annoying thing @peps is when i drag and release to the right (so back to latest blocks) it selects and highlights everything on the ruler 

image
image

peps:
@brianp it looks like that error is coming from `Swap.tsx` which isn't part of this PR. I did have a conflict with `package-lock` so I regenerated it... not sure if that caused an issue, its working for me. 

@shanimal08 thanks for testing and the feedback! I have a `user-select: none` on there and it doesn't seem to be happening for me... I cant seem to replicate the issue on Mac no matter how hard I try to drag and select things...

src/containers/main/Dashboard/components/BlockExplorerMini/useBlocks.ts
```diff
@@ -0,0 +1,64 @@
+import { useQuery } from '@tanstack/react-query';
+
+export const BLOCKS_KEY = ['blocks'];
+const address = 'https://textexplore.tari.com';
```
brianp, line 4: It's this url would need to be swapped out on a per network basis.

For mainnet this is the right address. For nextnet and esme it needs to be:

Emse: https://textexplore-esmeralda.tari.com/
Nextnet: https://textexplore-nextnet.tari.com/

brianp (approved):
Tested on Mac and worked well

PR #2105, opened 2025-05-13
feat(exchange-miner): set tari address

MCozhusheck:
Description
---
Allow users to set tari address in wallet settings to allow mining for different address.

How Has This Been Tested?
---
Open wallet settings and change it's address to different valid one.


src-tauri/src/commands.rs
```diff
@@ -562,11 +563,7 @@ pub async fn get_paper_wallet_details(
 }
 
 #[tauri::command]
-pub async fn get_seed_words(
-    _window: tauri::Window,
-    _state: tauri::State<'_, UniverseAppState>,
-    app: tauri::AppHandle,
-) -> Result<Vec<String>, String> {
+pub async fn get_seed_words(app: tauri::AppHandle) -> Result<Vec<String>, String> {
```
brianp, line 566: huh. I always thought `commands` need this specific parameters.

MCozhusheck: Commands are pretty smart and can deduce what parameters it needs. Take a look at [this](https://github.com/tari-project/universe/blob/main/src-tauri/src/commands.rs#L760) command that also only accepts app handle.

PR #2107, opened 2025-05-14
fix(ui): x spaces notification overlapping with the mainnet banner

peps:
The `X` spaces announcement banner was overlapping with the `Mainnet Live` banner. I just moved it down.

Note: When the MainNet banner is removed we'll need to manually adjust this notification again so it appears in the top right. 

Issue:

CleanShot 2025-05-14 at 15 05 50@2x

Fixed:

CleanShot 2025-05-14 at 15 01 31@2x

## Summary by CodeRabbit

- **Style**
  - Adjusted the position of the banner, moving it further down and closer to the right edge of the screen.


shanimal08 (approved):
thank you! :')

PR #2117, opened 2025-05-15
feat(exchange-miner): wait for exchange mining address during setup

MCozhusheck:
Description
---
Await for user input in exchange mining modal during modal and omit initializing wallet process if `EXCHANGE_ID` is defined.


src-tauri/src/setup/setup_manager.rs
```diff
@@ -581,7 +596,11 @@ impl SetupManager {
         self.setup_core_phase(app_handle.clone()).await;
         self.setup_hardware_phase(app_handle.clone()).await;
         self.setup_node_phase(app_handle.clone()).await;
-        self.setup_wallet_phase(app_handle.clone()).await;
+        let state = app_handle.state::<UniverseAppState>();
+        let in_memory_config = state.in_memory_config.clone();
+        if in_memory_config.read().await.exchange_id != DEFAULT_EXCHANGE_ID {
```
Misieq01, line 601: This option does not handle other places where we use `self.setup_wallet_phase`  second proposition with just doing the check in `self.setup_wallet_phase` and returning if we are on exchange miner would be better because it would cover all cases where we start wallet_phase ( switch to local node, restarting phases )

Issue #2119, opened 2025-05-15
Investigate Linux x AppImage white screen


shanimal08:
**Describe the bug**

Pretty sure [this ticket](https://github.com/tari-project/universe/issues/2097) is related (which includes a great amount of detail) but the TL;DR is some Linux users get a "white screen of death" when using appimage

- [helpful tweet context](https://x.com/darkfigod/status/1922808487345635392) which states: 

> use bash command to fix it:

> WEBKIT_DISABLE_DMABUF_RENDERER=1 ./TariUniverse.AppImage

> replace with the tari image name you used on your pc

### Action

see if we can add this env to our appimage builds via CI `WEBKIT_DISABLE_DMABUF_RENDERER`

MCozhusheck:
I found [this](https://www.reddit.com/r/tauri/comments/16tzsi8/tauri_desktop_app_not_rendering_but_web_does/) interesting post from Tauri maintainer:

> webkitgtk 2.42 is broken for all(?) users of propitary nvidia drivers because of its new renderer. You can disable it with an env var:
> ```
> 
> WEBKIT_DISABLE_DMABUF_RENDERER=1 npm run tauri dev
> 
> ```
> If that still doesn't work, try `WEBKIT_DISABLE_COMPOSITING_MODE=1` instead.
> 
> P.S. Ubuntu just disabled the new renderer in their webkitgtk distribution (for nvidia users only) so hopefully other distro follow suit. Last thing i read from the webkitgtk devs is that they don't plan to disable/revert it on their side.

brianp:
Should be fixed on #2125

PR #2125, opened 2025-05-15
fix(linux): add WEBKIT_DISABLE_DMABUF_RENDERER=1 on linux

Krakaw:
Description
---

This disables DMABUF_RENDERER on linux, found this snippet here: https://github.com/tauri-apps/tauri/issues/9394#issuecomment-2423273171

Motivation and Context
---

White screen on linux machines seems to be fixed by disabling DMABUF_RENDERER. 
#2097 #2119

How Has This Been Tested?
---

Ran this on linux, no white screen issue. Also ran it on OSX no changes, as expected.

---

Run this on Linux, don't get a white screen.

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

## Summary by CodeRabbit

- **Chores**
  - Improved environment configuration for Linux users in specific desktop session scenarios to enhance compatibility. No changes to visible features or exported functionality.


src-tauri/src/main.rs
```diff
@@ -969,6 +969,18 @@ struct FEPayload {
...
+    #[cfg(target_os = "linux")]
+    {
+        if std::path::Path::new("/dev/dri").exists()
+            && std::env::var("WAYLAND_DISPLAY").is_err()
+            && std::env::var("XDG_SESSION_TYPE").unwrap_or_default() == "x11"
+        {
+            // SAFETY: There's potential for race conditions in a multi-threaded context.
+            unsafe {
+                std::env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
+            }
```
MCozhusheck, line 981: Could we use [Once](https://doc.rust-lang.org/std/sync/struct.Once.html) to ensure we set this env only once?
```rust
static SET_WEBKIT_DISABLE_DMABUF_RENDERER: Once = Once::new();
SET_WEBKIT_DISABLE_DMABUF_RENDERER.call_once(|| {
    std::env::set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1");
});
```

MCozhusheck: But I am not sure if we need to be concerned about thread-safety in `main` function since it's run only once and all async actions are blocking, that's why there is `block_on` called on spawned tasks.

MCozhusheck: according to docs: `This function is safe to call in a single-threaded program.`
Ignore the code suggestion - it's safe to call in `main` function.

MCozhusheck (approved):

PR #2159, opened 2025-05-21
fix: handle panic caused by acquiring system bus on linux systems

MCozhusheck:
Description
---
There was common bug report in sentry:
```
called `Result::unwrap()` on an `Err` value: MethodError(OwnedErrorName(ErrorName(Str(Owned("org.freedesktop.DBus.Error.TimedOut")))), Some("Failed to activate service 'org.freedesktop.login1': timed out (service_start_timeout=25000ms)"), Msg { type: Error, sender: UniqueName(Str(Borrowed("org.freedesktop.DBus"))), reply-serial: 9, body: Signature("s") })
```

This was caused by external library used for listening for power status changes. The inner implementation used unhandled unwraps that would crash whole application.

```rust
fn get_lock_monitor(
  system_bus: zbus::blocking::Connection,
  manager_proxy: ManagerProxyBlocking<'_>,
) -> Result<(
  UnlockIterator<'_>,
  zbus::blocking::PropertyIterator<'_, bool>,
)> {
  let session_proxy = SessionProxyBlocking::new(&system_bus).unwrap();
  let session_id: String = session_proxy.id().unwrap();
  let session_obj_path: OwnedObjectPath = manager_proxy.get_session(&session_id).unwrap();

  let login_session_proxy = SessionProxyBlocking::builder(&system_bus)
    .path(session_obj_path)
    .unwrap()
    .build()
    .unwrap();
  let unlock = login_session_proxy.receive_unlock().unwrap();
  let locked_hint = login_session_proxy.receive_locked_hint_changed();

  Ok((unlock, locked_hint))
}
```

I suspect this was only possible while running TU in container or other virtualization machine that doesn't have D-Bus.

How Has This Been Tested?
---
I did not verify if this fix app crash since I don't have environment to reproduce it but I does not brake already working build.

### DIFF
--- a/src-tauri/src/utils/system_status.rs
+++ b/src-tauri/src/utils/system_status.rs
@@ -20,7 +20,7 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::sync::LazyLock;
+use std::{panic, sync::LazyLock};
 
 use anyhow::Error;
 use log::{error, info};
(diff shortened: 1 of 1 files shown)


MCozhusheck:
On a second thought It's not that safe to unwind panic or expect. It's not always safe and in this case we should rather show list of external dependencies for linux similar how it's already done for windows.

Issue #2156, opened 2025-05-20
Review manual Sentry data sends


brianp:
## Description

We're overloading Sentry with issues and not all of them are relevant now, and some don't have enough information to debug.

Review all manual sends to sentry and either remove them, or add additional info for debugging. 

Setup phases are important at the moment and should not be removed. Most other locations it is probably safe to remove.

MCozhusheck:
Some of these errors seems impossible to handle for example:
```
Failed to initialize gtk backend!: BoolError { message: "Failed to initialize GTK", filename: "/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gtk-0.18.2/src/rt.rs", function: "gtk::rt::init", line: 141 }
```
GTK is crucial for rendering on linux systems.  Maybe we could just check for presence of GTK and throw more readable error and don't report them to the sentry.

PR #2162, opened 2025-05-21
fix: speedtest panic on timeout

MCozhusheck:
Description
---
Addresses #2156 
Speedtest crate have a bunch of `expect(...)` that crashes app on timeout like this one:
```
failed to get response: reqwest::Error { kind: Request, url: "https://speed.cloudflare.com/__up", source: TimedOut }
```
In sentry it's highlighted that it's caused by:
```
cfspeedtest::speedtest::test_upload (speedtest.rs:238)
```
Which is:
```rust
pub fn test_upload(client: &Client, payload_size_bytes: usize, output_format: OutputFormat) -> f64 {
    let url = &format!("{BASE_URL}/{UPLOAD_URL}");
    let payload: Vec<u8> = vec![1; payload_size_bytes];
    let req_builder = client.post(url).body(payload);
    let (status_code, mbits, duration) = {
        let start = Instant::now();
        let response = req_builder.send().expect("failed to get response"); // This is line 238 that causes so many sentry reports
        let status_code = response.status();
        let duration = start.elapsed();
        let mbits = (payload_size_bytes as f64 * 8.0 / 1_000_000.0) / duration.as_secs_f64();
        (status_code, mbits, duration)
    };
    if output_format == OutputFormat::StdOut {
        print_current_speed(mbits, duration, status_code, payload_size_bytes);
    }
    mbits
}

```
Now these functions are run in separate tokio tasks that on panic returns errors instead of crashing whole app.

Testing
---
To verify it I added `panic!()` in to only download test to mimic timeout
```rust
        match spawn_blocking(|| {
            test_download(
                &reqwest::blocking::Client::new(),
                NETWORK_DOWNLOAD_SPEED_PAYLOAD_TEST,
                OutputFormat::None,
            );
            panic!("Download speed test failed");
        })
```
In logs I saw it just returned default value of 0:
```
09:34:03 INFO  [DEBUG NETWORK]Download speed: 0.00 Mbps
09:34:03 INFO  [DEBUG NETWORK]Upload speed: 22.44 Mbps
09:34:03 INFO  [DEBUG NETWORK]Latency: 87.50 ms
```
 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of network speed and latency tests by adding automatic retries on failure.
  - Enhanced error logging for network tests with more consistent and targeted log output.


MCozhusheck:
After reading that [unwinding is considered unsafe](https://smallcultfollowing.com/babysteps/blog/2024/05/02/unwind-considered-harmful/) it should probably be rewritten to handle panics by spawning a new task/thread and catch error from the task/thread handler.

src-tauri/src/utils/network_status.rs
```diff
@@ -107,35 +106,50 @@ impl NetworkStatus {
...
+            match spawn_blocking(|| {
+                test_upload(
+                    &reqwest::blocking::Client::new(),
+                    NETWORK_UPLOAD_SPEED_PAYLOAD_TEST,
+                    OutputFormat::None,
+                )
+            })
+            .await
+            {
+                Ok(speed) => upload_speed = speed,
```
brianp, line 141: This doesn't `break` and will run all 3 times.

brianp, line 141: This doesn't `break` and will run all 3 times.

src-tauri/src/utils/network_status.rs
```diff
@@ -107,35 +106,50 @@ impl NetworkStatus {
...
+                Err(e) => {
+                    error!(target: LOG_TARGET, "Failed to perform upload speed test: {:?}", e)
+                }
+            };
+        }
+
+        for _ in 0..3 {
+            match spawn_blocking(|| test_latency(&reqwest::blocking::Client::new())).await {
+                Ok(lat) => latency = lat,
```
brianp, line 150: Same here. It doesn't `break` and will run all 3 times.

brianp (changes requested):
I'm unsure if we need to test 3 times. I'm okay if we actually just try once, and if it fails it fails.

src-tauri/src/utils/network_status.rs
MCozhusheck: I removed retries altogether.

PR #2210, opened 2025-05-28
fix: integer overflow in pool status data

### LINKED ISSUE #2207: Pool mining fails with u32 overflow error and endless "Pool is warming up..." message
**Describe the bug**
Pool mining functionality is broken due to a data type overflow error. The console shows an error when fetching pool status, and the UI displays "Pool is warming up..." indefinitely, preventing actual mining from starting.

**To Reproduce**
Run Tari Universe from console
Ensure node and wallet are fully synced
Stay on main screen and attempt to start pool mining
Observe console error and UI behavior

**Expected behavior**
Pool mining should initialize properly after node and wallet synchronization is complete. The "Pool is warming up..." message should disappear once the pool connection is established, and mining should begin normally.

**Screenshots**
Console error: `03:35:49 ERROR Error fetching pool status: invalid value: integer `13739509030`, expected u32 at line 1 column 84`

Image

**Desktop:**
Ubuntu 24.04, Tari Universe 1.1.1

**Additional context**

- Node and wallet are fully synced
- Internet connection is stable
- The error appears to be related to a u32 integer overflow where a value 13739509030 exceeds the maximum u32 value (4,294,967,295)
- This suggests the pool status response contains a value that's too large for the expected data type
- The issue prevents pool mining entirely as the pool status fetch fails

This looks like a data type mismatch issue where the pool is returning a value larger than what the client expects to handle in a u32 field.

Misieq01:

### DIFF
--- a/src-tauri/src/pool_status_watcher.rs
+++ b/src-tauri/src/pool_status_watcher.rs
@@ -27,7 +27,7 @@ use serde::{Deserialize, Serialize};
 const LOG_TARGET: &str = "tari::universe::pool_status_watcher";
 #[derive(Clone, Debug, Serialize)]
 pub(crate) struct PoolStatus {
-    pub accepted_shares: u32,
+    pub accepted_shares: u64,
     pub unpaid: u64,
     pub balance: u64,
     pub min_payout: u64,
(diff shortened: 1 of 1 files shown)


brianp (approved):
LGTM

PR #2235, opened 2025-05-30
fix: small memleak fixes

fluffypony:
Description
---

Fixes small memory leaks in Tari Universe identified through comprehensive code analysis. Addresses global event listener leak in WebSocket management, implements data pruning for wallet store arrays, and improves process cleanup in Rust backend.

Motivation and Context
---

Memory leak analysis revealed several areas where resources were not being properly cleaned up:

1. Global WebSocket event listener in `socket.ts` was never unlistened, causing accumulation with repeated calls
2. Transaction arrays in wallet store could grow unbounded without any pruning mechanism
3. Process cleanup in `Drop` implementation could potentially deadlock in certain runtime contexts

These issues could lead to gradual memory accumulation over time, especially during long-running sessions.

How Has This Been Tested?
---

- Manual testing of WebSocket connection/disconnection cycles to verify proper cleanup
- Wallet store testing with large transaction datasets to confirm pruning functionality
- Process lifecycle testing to ensure proper cleanup under various shutdown scenarios
- Memory monitoring during extended application runtime

---

1. **WebSocket Leak Test**: 
   - Call `initialiseSocket()` and `removeSocket()` multiple times
   - Verify only one listener exists and is properly cleaned up
2. **Wallet Store Pruning Test**:
   - Add large numbers of transactions to trigger pruning limits
   - Verify arrays are trimmed to configured maximums
3. **Process Cleanup Test**:
   - Start/stop mining processes multiple times
   - Monitor system processes to ensure no zombies remain

Breaking Changes
---

- [x] None

### DIFF
--- a/src-tauri/src/process_adapter.rs
+++ b/src-tauri/src/process_adapter.rs
@@ -254,13 +254,40 @@ impl ProcessInstanceTrait for ProcessInstance {
 
 impl Drop for ProcessInstance {
     fn drop(&mut self) {
+        // Always trigger shutdown first
         self.shutdown.trigger();
+        
         if let Some(handle) = self.handle.take() {
-            Handle::current().block_on(async move {
-                if let Err(e) = handle.await {
-                    warn!(target: LOG_TARGET, "Error in Process Adapter: {}", e);
+            // Check if we're in a tokio runtime context
+            if let Ok(current_handle) = Handle::try_current() {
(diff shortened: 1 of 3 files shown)


mmrrnn:
Regarding transactions: We use pagination for fetching completed txs. We could just refetch with offset

PR #2272, opened 2025-06-02
fix(ui): bump QR code error correction

shanimal08:
Description
---
bump QR code error correction (highest on receive QR with the logo)


brianp:
My phone still returns "no usable data" while scanning.

shanimal08:
> My phone still returns "no usable data" while scanning.

welp, that's the highest sooo 😬 on the receive QR?

PR #2274, opened 2025-06-02
fix: pid file contains invalid pid

MCozhusheck:
Description
---
There were multiple sentry reports where PID files contained invalid value - 4 null controls character (U+0000). How this happens is unclear - it could be some corrupted hardware or improperly written buffer.
As fallback method this PR adds option to find process from the list of all running processes and acquire precious instance's PID this way.

How Has This Been Tested?
---
Close running TU abruptly with `kill -9 tu_process_id` or run command `xkill` to leave child processes running. Clicking close button will cleanup all running processes which is not  what we want in this case.
Change PID file of one running process to some invalid integer - it could be either `minotari_node` or `xmrig`.
Run TU again and check for logs:
```
warn!(target: LOG_TARGET, "pid file is not a valid integer: {}. Attempting to kill process by name", pid);
```
Then verify same PID file if it's content is valid integer.


MCozhusheck:
> We write these PID files right?
> 
> What's the chance the problem is something in our writes? Maybe parsing data from machines in other languages and not reading it proper?

Created this [PR](https://github.com/tari-project/universe/pull/2310) to use fail proof method.

PR #2277, opened 2025-06-03
fix(ui): sync page coin video stream from cloudflare

peps:
This loads the coin video as an adaptive bitrate stream from cloudflare ( HLS.js ) instead of being baked into the app. Loading the video from the app was causing some issues, it also increased bundle sizes. 

This shows a static image while the streaming video is loading then seamlessly ( pretty close ) switches from static image to video when it's ready. 

Downsides:

- If the user has no internet connection they will only see the static image ( but I don't think they can sync in this case anyways ? )
- If their connection is slow, the quality of the video will automatically reduce to adapt to the slow connection, good and bad thing. 

Upsides:

- Better performance when loading the video
- Less overall app bundle size

Note:

I had to slightly change the background color of the app on Dark Mode to match the video. The difference is so slight it shouldn't make a difference. 

CleanShot 2025-06-03 at 13 26 18@2x

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -54,10 +54,10 @@
                     "https://ethereum-sepolia-rpc.publicnode.com",
                     "tauri:",
                     "tauri://localhost:*",
-                    "ipc:",
-                    "http://ipc.localhost",
                     "http://localhost:*",
                     "http://127.0.0.1:*",
+                    "ipc:",
+                    "http://ipc.localhost",
                     "data:",
                     "blob:",
                     "application/octet-stream",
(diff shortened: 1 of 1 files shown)


src/theme/palettes/dark.ts
```diff
@@ -62,7 +62,7 @@ const darkPalette: ThemePalette = {
             default: c.grey[900],
             paper: c.grey[700],
             accent: 'rgba(255,255,255,0.06)',
-            main: c.grey[800],
+            main: '#242424',
```
shanimal08, line 65: shouldn't we rather change a value in the `grey` colours and use it here? 😬 just one that we haven't used elsewhere maybe?

PR #2281, opened 2025-06-03
feat(ui): add telemetry toggle to address form

shanimal08:
Description
---
- add telemetry opt in to exchange miner address input


brianp (approved):
Tested on MacOS

PR #2294, opened 2025-06-03
fix: inform user if he has no permission to download file

MCozhusheck:
Description
---
There is a common sentry error that users lack permissions to download binaries. This is most probably caused by not running as administrator on windows machines.

How Has This Been Tested?
---
Add this to `binaries_manager.rs` at line 480 to simulate lack of permissions:
```rust
        return Err(anyhow!(io::Error::new(
            io::ErrorKind::PermissionDenied,
            "Permission denied"
        )));
```

### DIFF
--- a/src-tauri/src/utils/error_utils.rs
+++ b/src-tauri/src/utils/error_utils.rs
@@ -1,9 +1,9 @@
 use std::io;
 
-pub fn find_io_error_kind(error: &anyhow::Error) -> Option<io::ErrorKind> {
+pub fn find_io_error(error: &anyhow::Error) -> Option<&io::Error> {
     for cause in error.chain() {
         if let Some(io_error) = cause.downcast_ref::<io::Error>() {
-            return Some(io_error.kind());
+            return Some(io_error.to_owned());
         }
     }
     None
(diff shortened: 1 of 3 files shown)


src-tauri/src/utils/error_utils.rs
```diff
@@ -0,0 +1,32 @@
...
+use std::io;
+
+pub fn find_io_error(error: &anyhow::Error) -> Option<&io::Error> {
+    for cause in error.chain() {
+        if let Some(io_error) = cause.downcast_ref::<io::Error>() {
+            return Some(io_error.to_owned());
+        }
+    }
+    None
+}
```
brianp, line 32: I don't love going through the anyhow chain but i suppose if we're using anywho then this is probably the best way to manage it. Normally doing any kind of stack unwind is quite costly and underperformant so it caught my eye. 

A couple suggestions:

1. Instead of a `for` look we could call an iterative finder that might be slightly more performant (marginal gains) like `find_map`. It might also tidy up the function a bit. 

2. If we're making a function to check for error types maybe we could make it generic so we could check for any kind of type

```
pub fn find_error_in_chain<T: std::error::Error + 'static>(error: &(dyn Error + 'static)) -> Option<&T> {
    error.chain().find_map(|cause| cause.downcast_ref::<T>())
}
```

I'm not sure if this will actually work, or if the compiler will be happy with it but worth a try.

brianp (commented):
Just a bit of feedback.

PR #2301, opened 2025-06-04
feat: build swaps-router on iframe

PanchoBubble:


src/App/App.tsx
```diff
@@ -80,15 +69,15 @@ export default function App() {
 
     return (
         <ThemeProvider>
-            <WagmiWrapper>
+            <QueryClientProvider client={queryClient}>
```
PanchoBubble, line 72: keeping the query client for two reasons:
* A the freaking wallet connect even though its on an iframe which has a query client still cries if this is not present
* B we should def use it on our requests :D

PR #2335, opened 2025-06-06
fix: iframe adjustments

PanchoBubble:


brianp (approved):
Looks JS-ey that's for sure.

Two thumbs up.

PR #2437, opened 2025-06-23
feat(ui): new start button with integrated mode select

peps:
This is a ground up rebuild of the "Start Mining" button. 

- New design for Start / Stop / Loading button styles
- Mining Mode dropdown integrated into the button
- Button background style changes based on the selected Mining mode

CleanShot 2025-06-24 at 11 20 46@2x

CleanShot 2025-06-23 at 17 16 25@2x

CleanShot 2025-06-23 at 17 17 18@2x

CleanShot 2025-06-23 at 17 17 47@2x

CleanShot 2025-06-23 at 17 18 19@2x


shanimal08 (approved):
very clean! 😻 LGTM, tested and it's working perfectly :D 

---

that said i have 2 gripes 😅, which are not related to these code changes but more the ux - 

1. can the base `Start Mining` disabled state be more clearly disabled? i had CPU mining off and was still syncing, so i couldn't start mining, but the button still _felt_ like i should have been able to 

2. the dropdown in the CTA only comes up when you've started mining, which is a bit frustrating if you just want to change modes/mess with the custom settings while not mining. maybe we should add the base dropdown to settings or something?

shanimal08 (approved):
💜

PR #2447, opened 2025-06-26
feat: remove flow for downloading releases

Misieq01:
### [ Summary ]
- Removed step for downloading releases list and all related logic
- binares_versions.json now requires providing hash alongside version

Fixes: #1995


src-tauri/binaries-versions/binaries_versions_nextnet.json
```diff
@@ -1,11 +1,12 @@
...
     "binaries": {
-        "xmrig": "=6.22.0",
-        "mmproxy": "=4.6.2-rc.0",
-        "minotari_node": "=4.6.2-rc.0",
-        "wallet": "=4.6.2-rc.0",
-        "sha-p2pool": "=1.0.0",
-        "glytex": "=0.2.26",
-        "tor": "=14.5.1"
+        "xmrig": "6.22.0",
+        "mmproxy": "4.6.2-rc.0 | 4bce3ea",
```
brianp, line 4: Also thinking this through. We lose the ability to pre-build releases.

Right now we utilize the chance to know an upcoming version before it's been released. TU can build with the version in the binary list. 

This change takes away our ability to do that because we must wait for the tagged version to exist in the other repo before we can add the git hash here. This is a talking point. How do we feel about that? @shanimal08 @mmrrnn

src-tauri/binaries-versions/binaries_versions_mainnet.json
```diff
@@ -1,11 +1,12 @@
...
-        "glytex": "=0.2.26",
-        "tor": "=14.5.1"
+        "xmrig": "6.22.0",
+        "mmproxy": "4.6.2 | 476927b",
+        "minotari_node": "4.6.2 | 476927b",
+        "wallet": "4.6.2 | 476927b",
+        "sha-p2pool": "1.0.3 | cff9241",
+        "glytex": "0.2.26 | 78e833f",
+        "tor": "14.5.1",
+        "bridge": "0.1.7"
```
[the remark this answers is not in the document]
Misieq01: I'm fine either way, it was just quicker to do in a string

brianp:
I blocked local access to the Tari cdn but setup failed and spammed logs while trying to download from GH.

Node progress went from 100% to 0% and got stuck in this state even though logs showed the node was running.
image

image

XMRIG failed to download from `https://github.com/xmrig/xmrig/releases/download/v6.22.0/xmrig-6.22.0-macos-arm64.tar.gz.sha256`

Logs spammed infinitely until I shut the app down. Logs attached.
[com.tari.universe.alpha.zip](https://github.com/user-attachments/files/20924395/com.tari.universe.alpha.zip)

src-tauri/binaries-versions/binaries_versions_nextnet.json
Misieq01: Are we using that flow at all ?

PR #2449, opened 2025-06-26
feat: base node status for tapplet signer

mmrrnn:
Works along with https://github.com/tari-project/wxtm-bridge-frontend/pull/71

Description
---
For observing base node status when wrapping XTM 

* Exposed for Tapplet use
* Update with new BaseNodeStatus type properties

Example response:
```ts
block_height: 46009
block_reward: 13399875729
block_time: 1750933785
is_synced: true
monero_randomx_network_hashrate: 13249
num_connections: 26
readiness_status: 0
sha_network_hashrate: 24614943
tari_randomx_network_hashrate: 0
```
----
I want to save it in the `debug` property in the bridge method:

```ts
    const { paymentId } = await createTransaction.mutateAsync({
      to: ethAddress,
      from: tariAccount.address,
      tokenAmount: parsedAmount,
      debug: { ...baseNodeStatusBefore }
    })
```

```ts
    const { success } = await confirmTokenSent.mutateAsync(paymentId, {
      debug: { ...baseNodeStatusAfter },
    })
 ```


mmrrnn:
I want to display in it Admin Panel to manually check node status

Issue #2214, opened 2025-05-28
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...


toxicfox2491:
so a pool Im mining in just started mining tari with monero, so I wanted to check it out but the app image is blank.
ran it through CLI and this is the output :

`(tari_universe:6399): Gtk-WARNING **: 09:07:37.196: Theme parsing error: gtk.css:6754:15: Junk at end of value for padding
Initializing logging according to "/home/may/.local/share/com.tari.universe/logs/universe/configs/log4rs_config_universe.yml"
09:07:37 INFO  RunEvent Ready
09:07:38 INFO  Pre Setup
09:07:38 INFO  Starting telemetry manager
09:07:38 INFO  [config_core] [load_config] loaded config content
09:07:38 INFO  websocket manager app handle set
09:07:38 INFO  [config_mining] [load_config] loaded config content
09:07:38 ERROR Error getting node identity: transport error
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
** (tari_universe:6399): WARNING **: 09:08:03.020: atk-bridge: get_device_events_reply: unknown signature`

Im on arch linux with AMD hardware
if you want more info please tell me, if Im being a dumbass and theres a really easy way to fix this also please tell me

blambin:
I also had the same problem on the Archlinux Wayland KDE6 desktop. amd cpu

toxicfox2491:
> To get the app to display: `LD_PRELOAD=/usr/lib64/libwayland-client.so.0 ./tari_universe_1.2.9_amd64.AppImage`
> 
> But no idea how to fix the transport error, it can't really connect to the network.
> 
> ```
> 22:20:22 INFO  RunEvent Ready
> 22:20:22 INFO  [config_core] [load_config] loaded config content
> 22:20:22 INFO  Pre Setup
> 22:20:22 INFO  Starting telemetry manager
> 22:20:22 INFO  websocket manager app handle set
> 22:20:22 INFO  [config_mining] [load_config] loaded config content
> 22:20:22 ERROR Error getting node identity: transport error
> 22:20:23 INFO  [config_wallet] [load_config] loaded config content
> 22:20:23 INFO  Wallet config does not exist or is corrupt. Creating new wallet
> 22:20:23 ERROR Error loading internal wallet: Credentials didn't exist, and this shouldn't happen
> 22:20:23 INFO  [config_ui] [load_config] loaded config content
[...]

PR #2455, opened 2025-06-30
fix(tor): use custom torrc file to avoid system level conflicts

zquestz:
Description
---
Add flag to use custom torrc file when launching tor. This is not a new flag for Tari, just a new flag being passed to tor when it starts up.

Motivation and Context
---
In Arch Linux, when Tor is installed, it is setup for the tor user. The existing torrc file is not suitable to be used for Tari and leads to tor not launching at all when using the AppImage. This also allows for the  Tari project to customize tor more easily, as now you can adjust the custom torrc file.

For more context see:
https://github.com/tari-project/universe/issues/2214

How Has This Been Tested?
---
Just tested locally, with the new flag tor starts as expected.

---
Install a system level tor that has an incompatible torrc, like in Arch Linux. Or just setup a completely invalid /etc/tor/torrc file.

Then try to launch the AppImage (or the binary, I assume both are affected). 

Notice that before the change, tor fails to startup, and the app is not usable. After the change, tor starts up as it can happily use a non-existent config file, just not an incompatible one!

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

Nothing should break.

### DIFF
--- a/src-tauri/src/tor_adapter.rs
+++ b/src-tauri/src/tor_adapter.rs
@@ -220,6 +220,7 @@ impl ProcessAdapter for TorAdapter {
 
         let working_dir_string = convert_to_string(working_dir)?;
         let log_dir_string = convert_to_string(log_dir.join("tor.log"))?;
+        let torrc_string = convert_to_string(data_dir.join("torrc"))?;
         let mut lyrebird_path = binary_version_path.clone();
         lyrebird_path.pop();
         lyrebird_path.push("pluggable_transports");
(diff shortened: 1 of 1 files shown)


brianp:
Hey! Thanks for the PR. I'll have a look over and test soon.

zquestz:
Had to make a slight adjustment, I thought rust would take care of the arg quoting for the log line, but apparently it doesn't, and it required quoting as well.

zquestz:
Just for context, it appears the libevent fix will also be needed. For now I have patched my version on the AUR to LD_PRELOAD the correct libevent.

This just addresses the existing configuration file issue, for people that use tor on their machines already and have an incompatible configuration.

zquestz:
Any progress here? This is blocking fixes for the AUR. I really want to make this work for people with tor installed on Arch Linux. =)

PR #2464, opened 2025-07-01
fix(ui): remove blur from cubes to fix windows issue

peps:
Im removing the blur effect from the SVG to fix rendering on windows. The blur made it look cool, but this has better performance and will fix the windows issue. 

CleanShot 2025-07-01 at 22 21 18@2x


shanimal08 (approved):
lgtm but untested

brianp (approved):
Tested on windows and working fine

PR #2467, opened 2025-07-02
fix: always use tor-bundled `libevent-2.1.so.7` on linux

bivsk:
Description
---
Regardless of whether or not a user has installed `libevent` from their package manager, the version of `libevent-2.1.so.7` bundled with tor should be used on linux systems.  Even though both the bundled and system copy are listed as the same version, they do not provide the same symbols.  This ABI mismatch is the cause of tor failing to start, not the lack of it being installed.

Here, the linker is able to find the system-copy of `libevent-2.1.so.7`, but tor still fails to launch due to the missing symbol:
```
# libevent is installed
$ sudo find / -name libevent-2.1.so.7
/usr/lib64/libevent-2.1.so.7

$ cd ~/.cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/14.5.1/tor

# linker using system copy
$ ldd ./tor
...
libevent-2.1.so.7 => /lib64/libevent-2.1.so.7 (0x00007f184eb91000)
...


$ ./tor
./tor: symbol lookup error: ./tor: undefined symbol: evutil_secure_rng_add_bytes
```

If we instead force the usage of the bundled `libevent-2.1.so.7`, tor starts as expected:
```
# linker using bundled copy
$ LD_PRELOAD=~/.cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/14.5.1/tor/libevent-2.1.so.7 ldd ./tor
...
/home/user/.cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/14.5.1/tor/libevent-2.1.so.7 (0x00007f8588200000)
...

$ LD_PRELOAD=~/.cache/com.tari.universe.alpha/binaries/tor-binaries/esmeralda/14.5.1/tor/libevent-2.1.so.7 ./tor
Jul 01 23:58:38.758 [notice] Tor 0.4.8.16 (git-64ccafd8115ecdec) running on Linux with Libevent 2.1.12-stable, OpenSSL 3.2.4, Zlib 1.3.1.zlib-ng, Liblzma N/A, Libzstd N/A and Glibc 2.41 as libc.
...
```

We need to avoid mixing system and tor-bundled copies of the tor binary itself and the libevent shared object file. 

Motivation and Context
---
https://github.com/tari-project/universe/pull/1997#issuecomment-2956292620
https://github.com/tari-project/universe/pull/1997#issuecomment-3025076918

How Has This Been Tested?
---
I've tested briefly in a Fedora VM. I've been unable to test on bare-metal, so if others are able, I'd appreciate the time/help.
```
23:32:50 WARN  Using LD_PRELOAD for libevent-2.1.so.7.
...
23:33:05 INFO  Waiting for Tor bootstrap: 100%
```
---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/tor_adapter.rs
+++ b/src-tauri/src/tor_adapter.rs
@@ -22,8 +22,6 @@
 
 use std::collections::HashMap;
 use std::path::PathBuf;
-#[cfg(target_os = "linux")]
-use std::process::Command;
 use std::time::Duration;
 
 use anyhow::{anyhow, Error};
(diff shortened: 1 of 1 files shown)


zquestz:
This looks reasonable to me. Thx for the fix @bivsk

brianp:
Most excellent contribution. I'll try to review it a bit later.

brianp (approved):

PR #2478, opened 2025-07-03
fix: filtering bridge txs

### LINKED ISSUE #2475: Outbound bridge TX's are listed in the Rewards filter
**Describe the bug**
When using the transaction history filter and selecting rewards. It shows outbound bridge transactions. Rewards should only be inbound.

**To Reproduce**
1. Have previously made bridge transactions
2. Select the TX history filter "Rewards"
3. See outbound bridge transactions

**Expected behavior**
I should only see p2pool payouts and coinbases.

**Screenshots**

Image

 - OS & Version: MacOS 15.4.1

mmrrnn:
Closes #2475

Don't always append bridge transactions, do it only when `transactions` or `all activity` filter selected


brianp (approved):
Filtering is working as expected. Bridge TX no longer shows in rewards. Only in Transactions and All Activity.

PR #2503, opened 2025-07-09
fix: prevent multiple setup restarts within second

Misieq01:


brianp (approved):
I always thought the restart mechanisms took into account currently running calls etc.

PR #2514, opened 2025-07-10
feat: http wallet support

brianp:
Description
---
Add support for http wallet scanning from the local base node.

Motivation and Context
---
Mo faster scanning

How Has This Been Tested?
---
Locally

---
run the app

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/binaries-versions/binaries_versions_testnets.json
+++ b/src-tauri/binaries-versions/binaries_versions_testnets.json
@@ -1,9 +1,9 @@
 {
     "binaries": {
         "xmrig": "6.24.0",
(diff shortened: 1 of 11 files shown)


brianp:
The drop in replacement of the node and wallet work fine and anecdotally perform quicker wallet scanning locally once the node is synced. 

Remote node functionality works when scanning the node from `rpc.esmeralda.tari.com`. GRPC functionality from the node such as identity fetching still needs to come from `grpc.esmeralda.tari.com` as per our previous remote note configuration. 

Remote node functionality needs a tuneup as it also attempts to use remote p2pool which isn't something that will be available.

PR #2549, opened 2025-07-16
chore: switch node to remote until local

Misieq01:


brianp:
Value is over written from config 🤔 

Could we do a config migration, and make it the default for everyone?

PR #2573, opened 2025-07-18
fix: node type migration

Misieq01:


src-tauri/src/configs/config_core.rs
```diff
@@ -103,6 +104,7 @@ impl Default for ConfigCoreContent {
             .unwrap_or(ABTestSelector::GroupA);
 
         Self {
+            migration_id: 1,
```
brianp, line 107: `config_wallet.json` uses `version`. We should keep it consistent.

src-tauri/src/configs/config_core.rs
```diff
@@ -143,10 +145,9 @@ impl ConfigCore {
         let mut config = Self::current().write().await;
         config.load_app_handle(app_handle.clone()).await;
 
-        if config.content.node_type.ne(&NodeType::RemoteUntilLocal) {
-            config
-                ._get_content_mut()
-                .set_node_type(NodeType::RemoteUntilLocal);
+        if config.content.migration_id.eq(&0) {
```
brianp, line 148: Existing configs won't have this field, so it won't equal 0, it won't exist, or it would be default 1.

brianp (changes requested):

brianp:
Tested.

New wallet was fine going to remote. Existing wallet stayed on Local. Was not migrated to RemoteToLocal.

PR #2615, opened 2025-07-28
feat(i18n): Add portuguese (Brazil) translation

Su3h7aM:
This pull request introduces the initial translation of the user interface to Brazilian Portuguese. As this is a first pass, some elements may require further review and refinement.

Key Points:
 * Initial Version: This is a literal translation. I still intend to refine and adapt some phrases for better understanding by a native speaker.
 * UI Layout Issues: Due to the verbose nature of the Portuguese language, some UI elements may appear broken or misaligned. Words and phrases in Portuguese often occupy more space than their English counterparts, leading to text overflow or wrapping issues in fixed-size containers.
 * Directory Naming Convention: I have a question regarding the best practice for naming language directories with regional variations. I've used pt-BR for Brazilian Portuguese, but I'm wondering if there's a preferred convention for handling different dialects.
 * Fallback Strategy: For handling language variations, I suggest a fallback mechanism. For example, for Brazilian Portuguese (pt-BR), the application could first look for translations in the pt-BR directory. If a key is not found, it could then fall back to a base Portuguese (pt) directory, and finally to English (en) as the default. This would allow for more efficient management of translations, as the pt-BR directory would only need to contain strings that differ from the base Portuguese translation. However, I lack the expertise in European Portuguese to implement this fully at this time.
 
I'm open to feedback and suggestions on how to best approach these points. Thank you for your time and consideration.


brianp:
Hey ! Thanks for the contribution! We'll get it added in asap.

PR #2668, opened 2025-08-04
feat(ui): security reminder warning button

peps:
Instead of auto-triggering the staged security modal multiple times, let's add this button to the wallet in the sidebar. 

The button will no longer be visible once all the steps are completed.

With this new button in the UI, the modal should only auto-pop once. If a user Skips or closes it, it should not auto trigger again.


brianp (changes requested):
Exchange miners don't have seeds to store. In this case there's only 1 step to complete:
image


In the event the pin code is already set, this bar doesn't reflect that there's nothing to be done, and if clicked, offers both steps to be completed. Which it should not.
image

peps:
@brianp i've removed the button from the exchanges. So now it should now only show when mining on Tari.

PR #2670, opened 2025-08-06
feat: fallback for pool mining

Misieq01:
### [ Summary ]
- Implemented fallback to solo mining for **GPU** and **CPU** miners via `handle_unhealthy` method
- Change `handle_unhealthy` method to return either **CONTINUE** | **STOP** to determine if health checks should continue
- Fixed shutdown for status updates of gpu and cpu miners in do_health_check methods
- Fallback is triggered after 30 minutes of healthcheck fails
- Cleanup usage of app_handle in setup_manager

### [ Testing for Devs ]
- Go to `gpu_miner_sha_adapter` => `handle_unhealthy` =>`  if duration_since_last_healthy_status.as_secs().gt(&(60 * 30))`
- Replace **30** with **1** to trigger fallback after one minute


src-tauri/binaries-versions/binaries_versions_mainnet.json
```diff
@@ -5,7 +5,7 @@
                 "minotari_node": "4.10.1-pre.1 | ad5aac9",
                 "wallet": "4.10.1-pre.1 | ad5aac9",
                 "sha-p2pool": "1.0.3 | cff9241",
-                "glytex": "0.2.31 | 4e3148b",
+                "glytex": "0.2.26 | 78e833f",
```
brianp, line 8: In the future keep your filthy linux hacks to yourself 😆

brianp (commented):
I am confused about where the `start` of the alternative config settings happens.

Here's my best guess:
- Service starts with Pool
- Unhealthy for 30min 
- Service stops instead of continues
- Service updated configuration
- SetupPhase continues to retry setup? Starting with the new configuration?

brianp:
Also looks like the stats watched still isn't shutting down

image

src-tauri/src/process_watcher.rs
```diff
@@ -331,10 +339,19 @@ async fn do_health_check<TStatusMonitor: StatusMonitor, TProcessInstance: Proces
             *uptime = Instant::now();
             stats.num_restarts += 1;
             stats.current_uptime = uptime.elapsed();
-            match status_monitor3.handle_unhealthy().await {
-                Ok(_) => {}
+            match status_monitor3
```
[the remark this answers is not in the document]
Misieq01: Not sure its very old code, left the clench of it for future, we could prepare graph for process watcher / adapters flow like with setup_manager and maybe cleanup it a bit

Misieq01:
> Also looks like the stats watched still isn't shutting down
> 
> image

Are you mining when you see it ? Or did you stopped mining

brianp:
> Are you mining when you see it ? Or did you stopped mining

The error is present on startup, while GPU Pool mining is _attempting_ to connect and failing. But it goes into this infinite rapid repeat after switching successfully to GPU Solo mining.

src-tauri/src/xmrig_adapter.rs
```diff
@@ -229,6 +231,31 @@ pub struct XmrigStatusMonitor {
 
 #[async_trait]
 impl StatusMonitor for XmrigStatusMonitor {
+    async fn handle_unhealthy(
+        &self,
+        duration_since_last_healthy_status: Duration,
+    ) -> Result<HandleUnhealthyResult, anyhow::Error> {
+        // Fallback to solo mining if the miner has been unhealthy for more than 30 minutes
+        info!(target: LOG_TARGET, "Handling unhealthy status for Xmrig | Duration since last healthy status: {:?}", duration_since_last_healthy_status.as_secs());
+        if duration_since_last_healthy_status.as_secs().gt(&(60 * 1)) {
```
brianp, line 240: I think this is probably a test timeout length? :D

PR #2695, opened 2025-08-11
fix: clear tx cache when pending tx

mmrrnn:
Description
---
Clear converted tx cache when new pending tx


brianp (approved):
Smooth moves. 

Working as expecting. Outbound pending TX show up right away, and incoming TX's show up in timely order.

PR #2753, opened 2025-08-20
feat(ui): select dropdown keyboard navigation

peps:
This update adds keyboard navigation support to the `Select` and `ModeDropdown` select boxes.

This should improve usability and accessibility within the app when using these select dropdowns.

1. Arrow Key navigation

Once open, you can navigate through the select boxes using the up and down arrow keys. When you reach the end the selector jumps to the top. 

2. Letter Key navigation

Once open, you can type the first letter of the option you want to select, it should jump to the closest match.

CleanShot 2025-08-19 at 20 38 10@2x

CleanShot 2025-08-19 at 20 38 45@2x


peps:
> @peps it's got some conflicts now

ok fixed

PR #2766, opened 2025-08-25
feat: beta build delivery

mmrrnn:
It's finished,
On rwa backend side we just need to fix asset when publishing(null)


brianp:
Great work but at this point we have no need for enhanced build delivery.

PR #2777, opened 2025-08-27
fix: clamp cpu cores to use between 1 and max available cores

Misieq01:
### [ Summary ]
- Prevent passing to xmrig threads arg values below 1 and bigger then available threads


brianp (approved):

Misieq01:
> I'm curious what brought this on? 😄

Saw in one of error logs that it tried to run xmrig with 0 threads, forward you the error

PR #2783, opened 2025-08-28
chore(ci): update ci linting to run more often, but with less perms

leet4tari:
Description
update ci linting to run more often, but with less perms


shanimal08 (approved):
ty 💜

PR #2795, opened 2025-09-01
feat: pools status fetching improvements

Misieq01:
### [ Summary ]
- Separated pool logic form mining flow so we are able to fetch status even when not mining
- When mining we are fetching every 60 seconds | When we were mining and stopped we will continue to fetch stats for hour every 300 seconds
- Supports stats fetching when not mining and trying to switch pool / change wallet address / import seed words 
- Added caching to every downloaded stat during session
- Fronted now receives all available pools stats ( opportunity in future to display summary of all stats instead of just current selected pool )


src-tauri/src/mining/gpu/mod.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2024. The Tari Project
```
brianp, line 1: Does this empty file serve a purpose. It's not even re-exporting anything :D

src-tauri/src/tasks_tracker.rs
```diff
@@ -66,23 +66,23 @@ impl TaskTrackerUtil {
...
 impl TasksTrackers {
     fn new() -> Self {
         Self {
-            wallet_phase: TaskTrackerUtil::new("Wallet phase"),
-            node_phase: TaskTrackerUtil::new("Node phase"),
-            cpu_mining_phase: TaskTrackerUtil::new("CPU Mining phase"),
-            gpu_mining_phase: TaskTrackerUtil::new("GPU Mining phase"),
-            core_phase: TaskTrackerUtil::new("Core phase"),
-            common: TaskTrackerUtil::new("Common"),
+            wallet_phase: Arc::new(TaskTrackerUtil::new("Wallet phase")),
```
brianp, line 80: Where did the need for `Arc` come in specifically?

src-tauri/src/mining/gpu/mod.rs
Misieq01: Merged this branch into new gpu miner one and there its used :D ( Did it because logic changes helped me implement new miner faster)

PR #2813, opened 2025-09-02
fix: mining proof should not require client trigger command

PanchoBubble:
## Description

This PR refactors the websocket connection management to make mining proof operations more robust and automatic, eliminating the need for explicit client trigger commands.

### Backend (Rust/Tauri)

   * Enhanced websocket connection reliability: Added retry logic with exponential backoff and timeout handling in websocket_connect
   * Refactored websocket architecture: Moved websocket logic into dedicated websocket_manager.rs and websocket_events_manager.rs modules
   * Implemented automatic reconnection: Added websocket-reconnected event listener that automatically restarts the events manager
   * Added websocket keep-alive mechanism: Ensures persistent connection stability
   * Improved shutdown handling: Enhanced websocket_close to properly cleanup both manager and event handler

### Frontend (TypeScript/React)

   * Removed manual websocket triggers: Eliminated websocket_connect() calls from:
      * useAirdropPolling.ts
      * useRustWebsocket.ts
      * airdropStoreActions.ts
      * socket.ts
   * Simplified connection flow: Websocket connections are now managed automatically by the backend

## Benefits

   * Improved reliability: Websocket connections are more resilient to network issues and reconnect automatically
   * Simplified architecture: Removes coupling between frontend components and websocket connection management
   * Better user experience: Mining proof operations work seamlessly without requiring manual connection triggers
   * Enhanced stability: Proper cleanup and restart mechanisms prevent connection state issues


Krakaw (approved):

src-tauri/src/websocket_events_manager.rs
```diff
@@ -104,58 +121,81 @@ impl WebsocketEventsManager {
...
+                tokio::select! {
+                  _= interval.tick() => {
+                        info!(target:LOG_TARGET, "✓ Mining status interval tick - assembling status");
+                        match AssertUnwindSafe(WebsocketEventsManager::assemble_mining_status(
+                          cpu_miner_status_watch_rx.clone(),
+                          gpu_latest_miner_stats.clone(),
+                          node_latest_status.clone(),
+                          app_id.clone(),
+                          app_version.clone(),
+                        )).catch_unwind().await {
```
brianp, line 161: I'm sure this works but it seems quite strange. Expecting this to potentially panic is a code smell. If it can error, then the return value of `assemble_mining_status` should be a `Result<WebsocketMessage>`. A function returning an Option should always be safe to call.

So what's happening here that we decided this function could panic and we need to catch and unwind it. Generally catching a panic in rust is a bit of an antipattern.

PanchoBubble: I was sometimes getting

> 12:58:05 INFO  ✓ Mining status interval tick - assembling status
>  thread 'tokio-runtime-worker' panicked at src/internal_wallet.rs:91:24:
  InternalWallet is not initialized

Assemble message uses wallet address, I'll give it a look

PanchoBubble: Fixed on latest change

PR #2815, opened 2025-09-03
fix(ui): make a root avatar component and integrate with crew rewards

peps:
This PR moves the Avatar component from the Airdrop Sidebar to be a global component. 

Its now being used for all avatars in the Crew Rewards widget as well as the sidebar. It requires a username to be passed in order to generate an avatar for the user when the image field is blank.

CleanShot 2025-09-04 at 17 32 25@2x


brianp:
conflicts

peps:
> conflicts

fixed

PR #2822, opened 2025-09-04
fix(ui): set dialog z-index to a higher number to fix modal stacking

peps:
The z-index on the `Dialog` component being set to `1` is too low and causing issues with the stacking order. Modals should always be above everything ( except other more important modals like critical errors ).   

This PR sets the `z-index: 99` which should be high enough to prevent issues. 

CleanShot 2025-09-04 at 18 40 41@2x

CleanShot 2025-09-04 at 18 46 44@2x

### DIFF
--- a/src/components/elements/dialog/Dialog.styles.ts
+++ b/src/components/elements/dialog/Dialog.styles.ts
@@ -7,7 +7,7 @@ export const Overlay = styled(FloatingOverlay)`
     display: flex;
     align-items: center;
     justify-content: center;
-    z-index: 1;
+    z-index: 9;
 `;
 
 export const ContentWrapper = styled.div<ContentWrapperStyleProps>`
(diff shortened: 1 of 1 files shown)


PanchoBubble:
well.. we could have some sort of enum/object with different levels so we don't use numbers but takes some time to set it up correctly, and z index would still be relative to the container 😬  we could still have one for position fixed components 🤷

shanimal08:
> Na, that's too complex, and I don't want to be contained to an enum. There was just some back and forth with this because the modals were being refactored at the same time as Crew Rewards being merged in. So the two were not tested together.
> 
> Once the values are settled it wont be an issue for a while. Im confident in this update. Merge it, test it, move on.



i concur :D 

(i personally still feel like `99` is still too high a number and leaves the door open for a `100` or `1000` to creep back in later, but it should be fine for now as long as we're all on the same page 😅 )

PR #2834, opened 2025-09-09
fix(ui): emoji address tooltip z-index

shanimal08:
Description
---
- export `Dialog` component's z-index for use elsewhere
- use that z-index for the emoji address tooltip

Motivation and Context
---
- prev update to the `Dialog` z-index left the emoji address tooltip behind 😢: 
optimised_09-09_10-48_1229

How Has This Been Tested?
---

optimised_09-09_10-46_1228


peps (approved):

peps:
@shanimal08 to continue our discussion on z-indexes. If we want to set up a system, I recommend we leave room for flexibility. For example:

Layer 1 - Base APP UI = 0-9
Layer 2 - Popovers = 10-19
Layer 3 - Modals = 20-39
Layer 4 - Toasts = 40-59
Layer 5 - Tippy Top = 60-infinity 

We can go up by 10s, or 100s. The buffer between numbers gives us the flexibility to adjust where needed. For example, we may want some Modals to appear over others, same with Popovers etc. 

Where is gets tricky is stuff like tooltips, since their z-index should be scoped to the parent. Maybe we can get away with adding them to the Toasts layer since they should always be on top

PR #2870, opened 2025-09-12
fix(ui): simplified crew filters

peps:
This PR simplifies the Crew Reward filters. It removes the "All" and "Completed" filters from crew rewards and just has  "Active" and "Inactive". 

CleanShot 2025-09-12 at 17 40 44@2x

### DIFF
--- a/src-tauri/src/mining/gpu/manager.rs
+++ b/src-tauri/src/mining/gpu/manager.rs
@@ -197,6 +197,111 @@ impl GpuManager {
             .insert(miner.miner_type.clone(), miner);
     }
 
+    /// Handles loading the pool connection for the selected miner.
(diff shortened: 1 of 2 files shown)


PanchoBubble:
@peps Updated the BE, active and inactive would show also complete rewards, for the moment I haven't removed the filters on the BE side just in case but I can remove them if you rather not have them there

peps:
> @peps Updated the BE, active and inactive would show also complete rewards, for the moment I haven't removed the filters on the BE side just in case but I can remove them if you rather not have them there

@PanchoBubble sounds good thanks! I dont think I need to make any more changes on my end, if the data coming from the BE is correct. Let's test. Marking as `Ready for review`

PR #2887, opened 2025-09-16
fix(ui): toast and admin ui layout improvements

peps:
This PR fixes a few issues with the Toasts UI, and improves the Admin UI menu to allow each category to toggle. 

---

**Toasts:**

1. Reposition them so they are centered within the app. The initial idea of having them not centered was so that they line up with the tower. But since most toasts happen when modals are open, this is weird. I just made it so they're always centered. 

CleanShot 2025-09-16 at 16 33 26@2x

2. The animated outline on the close button stopped working. It now animates showing the exact amount of time before the Toast auto-closes. 

CleanShot 2025-09-16 at 16 32 57@2x
 
3. Added a max-width to the description so the text will wrap on multiple lines, this will be better for long error messages.

CleanShot 2025-09-16 at 16 39 45@2x

---

**Admin UI:**

1. Made each Admin UI group toggle, Theme and Feedback are open by default. This will allow is to stuff even more in here. 

CleanShot 2025-09-16 at 16 34 43@2x

2. Added the Toasts into the Admin UI

CleanShot 2025-09-16 at 16 35 26@2x


shanimal08 (approved):
nice lgtm!

PR #2895, opened 2025-09-17
feat: add more informations about miners

Misieq01:
### [ Summary ]
- Expanded miner selection component to include information about its health status and features
- Updated data structure send by backend
- Created new GpuMinerSelect component
image
image


brianp (changes requested):
On my windows machine, where lolminer can't run I see constant retrying and no failover, or warning that it's not working.

Misieq01:
> On my windows machine, where lolminer can't run I see constant retrying and no failover, or warning that it's not working.

Should be fixed by #2897

PR #2901, opened 2025-09-18
fix: add version counter to each config file

Misieq01:
### [ Summary ]
- Added `version_counter` to each config
- If there was `version` ( i32 ) present replaced it with `version_counter` ( u32 ) | config version cannot be negative, replacement was needed as I cannot change type of that field because then the deserialization would fail 
- Added pub const's to each config representing it current expected version 
- Set `version_counter` default value as that mentioned pub const so fresh configs will not require any migration 
- Removed `was_config_migrated` field


brianp (approved):
Tested on mac os.

Issue #2794, opened 2025-09-01
v1.5.13 - Beta - 0829: Windows 11: When I try to sync app with phone or send transactions I receive an error and an alert stating `Enable keychain access`


Tas4tari:
**Describe the bug**
I wanted to send some funds to my Esme mobile wallet and since I was unable to download the new Beta build for today, I was going to use the one from Friday:
V1.5.13 - Beta - 0829
Upon running the build and trying to send the funds, I see an error and a pop up telling me to Enable keychain access on my Windows 11 laptop
This alert also comes up when I try to `Sync with Phone`, upon clicking on the: `I have the app` CTA

**Note:**
This did not occur on Friday when I was testing the build as I could send and receive funds without issue

**Pre-conditions**
- User should have v1.5.13 - Beta - 0829 installed
- App should be open
1. Click on 'Send'
2. Follow the process and enter an address and the amount you would like to send 
3. Click on `Review`

**Expected behavior**
User should not see the `Enable keychain access` on Windows

**Screenshots**
<img width="1626" height="960" alt="Image" src="https://github.com/user-attachments/assets/37945ddb-ba85-4c17-9884-e247e31b87a5" />

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11

**Logs:**
[logs.zip](https://github.com/user-attachments/files/22072629/logs.zip)

brianp:
Fixed in #2903

PR #2903, opened 2025-09-18
fix: only emit keychain evt for mac

shanimal08:
Description
---
- just added the cfg macro for mac to the other placed `emit_show_keyring_dialog` was called

Motivation and Context
---
- Keychain access dialog was prompted to a windows user

How Has This Been Tested?
---
- locally (on mac only, needs windows testing)


brianp:
How can i trigger the keychain modal on windows?

shanimal08:
> How can i trigger the keychain modal on windows?

can check with @Tas4tari, but here's the original issue https://github.com/tari-project/universe/issues/2794 - so when trying to send or sync with phone? (though hopefully not on this PR!)

PR #2905, opened 2025-09-18
fix: bring back device exclusion logic for graxil and glytex

Misieq01:


src-tauri/src/mining/gpu/miners/glytex.rs
```diff
@@ -109,11 +110,42 @@ impl GlytexGpuMiner {
...
         }
     }
 }
 
 impl GpuMinerInterfaceTrait for GlytexGpuMiner {
+    async fn load_excluded_devices(
+        &mut self,
+        _excluded_devices: Vec<u32>,
+    ) -> Result<(), anyhow::Error> {
+        self.excluded_devices = _excluded_devices;
```
brianp, line 123: Why `_` if it's being used now?

src-tauri/src/mining/gpu/miners/graxil.rs
```diff
@@ -94,11 +95,20 @@ impl GraxilGpuMiner {
...
         }
     }
 }
 
 impl GpuMinerInterfaceTrait for GraxilGpuMiner {
+    async fn load_excluded_devices(
+        &mut self,
+        _excluded_devices: Vec<u32>,
+    ) -> Result<(), anyhow::Error> {
+        self.excluded_devices = _excluded_devices;
```
brianp, line 108: Same with `_` var is being used.

src-tauri/src/mining/gpu/miners/glytex.rs
```diff
@@ -109,11 +110,42 @@ impl GlytexGpuMiner {
...
+        let config_dir = config_path.join(APPLICATION_FOLDER_ID);
+
+        let gpu_engine_statuses_path = config_dir.join("gpuminer").join("engine_statuses");
+
+        let gpu_status_file_name = format!("{selected_engine}_gpu_status.json");
+        let gpu_status_file_path = gpu_engine_statuses_path.join(gpu_status_file_name);
+        let mut gpu_status_file =
+            load_file_content::<GlytexGpuDevices>(&gpu_status_file_path).await?;
+        for device in &mut gpu_status_file.gpu_devices.iter_mut() {
+            if self.excluded_devices.contains(&device.device_index) {
```
brianp, line 138: Didn't we encounter an issue where the device index wasn't always consistent?

brianp (changes requested):

src-tauri/src/mining/gpu/miners/glytex.rs
Misieq01: In the end I think we didn't encounter mismatch, and if we will then either way there is no better solution to keeping id which will work across gpu's, as name is not an option in case of multiple same gpu's

PR #2915, opened 2025-09-22
fix(ui): xspace banner z-index issue fix

peps:
The x-space banner had a z-index that put it above the crew rewards ui. So when you expand the crew rewards it covers it. This PR fixes that. 

Bug:
CleanShot 2025-09-22 at 10 45 22@2x

Fix:
CleanShot 2025-09-22 at 10 47 45@2x

Tested:

- [x]  Make sure XSpace banner is still clickable
- [x]  Make sure XSpace banner does not overlap crew rewards
- [x]  Make sure Banner is below all modals / dialog ui

### DIFF
--- a/src/containers/floating/XSpaceBanner/XSpaceBanner.style.ts
+++ b/src/containers/floating/XSpaceBanner/XSpaceBanner.style.ts
@@ -10,7 +10,7 @@ export const BannerContent = styled(m.div)<{
     top: 16px;
     right: 20px;
 
-    z-index: 99;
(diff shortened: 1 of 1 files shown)


peps:
@shanimal08 just double checking, but is there a reason we're setting the z-index on the XSpace banner so high all the time? Is there another issue im not aware of? Setting it to 2 seems to work fine in my testing.

PanchoBubble (approved):
we need a z-index bug counter 😂

PR #2925, opened 2025-09-22
fix: possible malformed node config files

Misieq01:
Probably will fix: https://nomatter.tari.com/tari/pl/gorbc5eyuj8qxq6ecepnuo6eba

From what I've investigated in core repo, some config is malformed which throws error on serialization which in returns throws 101 error in TU, to try and fix this I've added deletion of config folder on `handle_unhealthy` 

Tested by changing config.toml file and running TU app, I can see the same error as in error report and then deletion happen which looks like it fixing the issue, node running properly after


brianp (approved):
Good idea.

Issue #2921, opened 2025-09-22
Default Disablement of GPU Mining on Low-Spec Systems


brianp:
## Summary  
By default, GPU mining should be disabled if the system has an integrated GPU **and** less than 16 GB of RAM.  

## Why?  
Running mining workloads on low-spec systems (integrated graphics + limited memory) often makes the initial experience of starting the app worse as the system quickly gets bogged down in mining.

Disabling GPU mining in these cases by default would protect users from unintentional system use and create a more sensible out-of-the-box experience. Users can then opt in to GPU mining at their leisure. 

## Proposed Behaviour  
- Detect system specs on startup.  
- If:  
  - GPU = integrated graphics, **and**  
  - RAM < 16 GB  
  → Default GPU mining = **disabled**.  
- Users can still explicitly enable GPU mining in the configuration if they want to override.

PR #2932, opened 2025-09-24
feat: disable gpu mining by default for low end machines

Misieq01:
### [ Summary ]
- Added logic that during setup performs two checks:
  - Check system RAM to see if it is higher then 16GB
  - Checks graxil detection output to see if there is any dedicated graphic card
 - If both checks pass then gpu mining is enabled by default, if not then we initially disable it, but user can enabled it if he wants
 - Added warning informing user that we do not recommend gpu mining on his hardware
image

#### Solves #2921

### DIFF
--- a/src/containers/floating/Settings/sections/mining/GpuMiningMarkup.tsx
+++ b/src/containers/floating/Settings/sections/mining/GpuMiningMarkup.tsx
@@ -2,10 +2,7 @@ import { useCallback, useMemo } from 'react';
 
 import { Typography } from '@app/components/elements/Typography.tsx';
 import { ToggleSwitch } from '@app/components/elements/ToggleSwitch.tsx';
-import { Stack } from '@app/components/elements/Stack.tsx';
-import { IoWarning } from 'react-icons/io5';
 import { useTranslation } from 'react-i18next';
-import styled from 'styled-components';
 import {
     SettingsGroup,
     SettingsGroupAction,
@@ -17,15 +14,6 @@ import { useMiningMetricsStore } from '@app/store/useMiningMetricsStore.ts';
 import { setGpuMiningEnabled, useConfigMiningStore } from '@app/store';
 import { useSetupStore } from '@app/store/useSetupStore.ts';
 import { setupStoreSelectors } from '@app/store/selectors/setupStoreSelectors.ts';
-
-const WarningIcon = styled(IoWarning)(({ theme }) => ({
-    color: theme.palette.warning.main,
-    flexShrink: 0,
-}));
-
-const WarningMessage = styled(Typography)(({ theme }) => ({
-    color: theme.palette.warning.main,
-}));
 
 const GpuMiningMarkup = () => {
     const { t } = useTranslation(['settings'], { useSuspense: false });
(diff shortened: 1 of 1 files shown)


brianp:
I don't think we need a warning or anything about why it was initially disabled. We just do it, and it'll be the users experience that it was done :D

Misieq01:
> I don't think we need a warning or anything about why it was initially disabled. We just do it, and it'll be the users experience that it was done :D

So should I remove the warning ?

brianp:
Yup!

src-tauri/src/hardware/hardware_status_monitor.rs
```diff
@@ -357,6 +363,57 @@ impl HardwareStatusMonitor {
...
     }
 
+    // This method will decide if gpu mining is recommended by which we mean if it should be enabled by default in config mining
+    // It will have two steps verification:
+    // Checking raw_detected_devices in graxil which includes device_type ( INTEGRATED or DEDICATED )
+    // Checking system memory with sysinfo
+    // If there is at least one dedicated gpu and system memory is above 16GB we recommend enabling gpu mining
+    pub async fn decide_if_gpu_mining_is_recommended(&self) -> Result<(), Error> {
+        let mut is_dedicated_gpu_found = false;
+        let mut is_system_memory_above_16gb = false;
```
brianp, line 373: Should be >=16gb

16GB and integrated is a valid state to leave GPU enabled.

Which I can see on line 389, it is a `>=` check, but the name of this variable makes it slightly misleading.

src-tauri/src/mining/gpu/miners/graxil.rs
```diff
@@ -83,6 +83,7 @@ pub struct GraxilGpuMiner {
     pub connection_type: Option<GpuConnectionType>,
     pub gpu_status_sender: Sender<GpuMinerStatus>,
     pub gpu_devices: Vec<GpuCommonInformation>,
+    pub raw_gpu_devices: Vec<GraxilGpuDeviceInformation>,
```
brianp, line 86: Why do we have a new `raw_` devices field. Why can't we get this from the `gpu_devices` field?

src-tauri/src/setup/phase_gpu_mining.rs
```diff
@@ -282,6 +282,9 @@ impl SetupPhaseImpl for GpuMiningSetupPhase {
             })
             .await?;
 
+        let _unused = HardwareStatusMonitor::current()
+            .decide_if_gpu_mining_is_recommended()
+            .await;
```
brianp, line 287: I'm confused how we decide not to start the miner based on this function when the value is unused. I can't find the config value used anywhere either. I see `gpu_mining_enabled` used in the front end but nowhere in the backend to prevent the starting of mining.

src-tauri/src/mining/gpu/miners/graxil.rs
Misieq01: gpu_devices are using common data structure which supports all miners
raw_gpu_devies is the data structure retrived directly from graxil which includes gpu_type

src-tauri/src/setup/phase_gpu_mining.rs
Misieq01: In case this method will decide that mining is not recommended it will update config value for `gpu_mining_enabled` and frontend can't start gpu mining when gpu mining is disabled which prevents gpu mining

PR #2935, opened 2025-09-25
fix: ensure process are killing when shutting down the app

Misieq01:


brianp:
Conflicts.

Also- we already track and try to shutdown all existing processes. What exactly does this _add_ that it should do a better job than out previous attempts to ensure everything is shutdown?

PR #2936, opened 2025-09-25
fix: separate telemetry gpu hash rate data

Misieq01:
### [ Summary ]
- Added algorithm field to **GpuMinerStatus**
- Separated `gpu_hash_rate` field in telemetry data into: `gpu_hash_rate` and `gpu_hash_rate_c29`


src-tauri/src/telemetry_manager.rs
```diff
@@ -658,6 +669,7 @@ async fn get_telemetry_data_inner(
         cpu_utilization,
         cpu_make,
         gpu_make,
+        gpu_hash_rate_c29,
```
brianp, line 672: We need to add it to the `extra_data` field. For some reason they can't expand the schema at the root level 🤷🏻‍♂️

Issue #2864, opened 2025-09-11
TU base node advertising not dial able addresses


SWvheerden:
Currently, TU fill in the public address of the base node via cli as some 127.0.0.1:XXX address. 
This will cause the node to advertise this as its address across the network. But no peer will be able to dial that node via this address. We should only ever fill that in with an address that is dialable accross the network. This is not a show stopped, but does poisen its own address for the network as each peer who wants to contact this peer now has to try invalid addresses as well.

PR #2942, opened 2025-09-29
fix: remove tcp address args

shanimal08:
Description
---

just removed the tcp local address args from the node and wallet adapters

Motivation and Context
---
- https://github.com/tari-project/universe/issues/2864

How Has This Been Tested?
---
- ran locally with `Local` node - not 100% since i'm on mac so should be tested on windows

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


SWvheerden (commented):
LGTM

Misieq01 (approved):

brianp (approved):

brianp:
We could probably do some function cleanup around this as well. We no longer need to generate or pass the TCP ports / addresses. We're not using them anymore. But this can happen in a follow up.

PR #2944, opened 2025-09-30
feat: tasktray app

Misieq01:
### [ Summary ]
- Changed closing behavior, close button no longer is shutting down the app, instead it minimizes it to the task tray. To close the app user has to click close button in systemtray menu, but before closing it will focus the app and display survey.
- Moved cpu_miner logic to manager pattern and removed unused code
- Refactored how we are starting and stopping mining to allow usage in rust side
- Refactored Systemtray to collect data by listening on the event channel
- Implemented new data ( Pool pending rewards, mining mode, miners enabled status ) and new actions ( Open tari Universe, Start / Pause Mining, Open Settings, Close ) for systemtray menu
- Fixed pool rewards not updating ( They were only fetched during opening of the app )

### [ Images ]
image


brianp (changes requested):
- When GPU mining is disabled
- And mining is off
- I select "Start Mining" from the task bar menu
- GPU mining stays disabled but goes into a loading state forever

brianp (approved):

PR #2983, opened 2025-10-09
feat: pause mining base branch

shanimal08:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

https://github.com/user-attachments/assets/c4b0b6a7-55b2-4506-b5d5-2df92c441553

---

Breaking Changes
---

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/event_scheduler.rs
+++ b/src-tauri/src/event_scheduler.rs
@@ -343,6 +343,37 @@ pub enum TimePeriod {
     PM,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


brianp:
I think the event scheduler class could benefit from a top level description of what it's doing, and a mention of the fact we're only dealing in cron syntax, and not with cron itself. I worried myself looking for how we were implementing cron tasks, and windows compat before I realized we're just using it as a scheduler format, and everything we need is maintain in memory (or persisted to config).

brianp (approved):

PR #3002, opened 2025-10-16
feat: pause on battery mode

Misieq01:
### [ Summary ]
- Implemented battery_status listener for retrieving information's about device battery and charging status
- Implemented callbacks for stop / start mining depending on current charging status
- Added new config option under mining section  to deciding if mining should be stopped once on battery mode
- New config option is only visible for devices that have battery ( This option won't be shown for PC's )
- Added information dialog about why mining was paused, which automatically goes away once AC is plugged
image

### DIFF
--- a/src-tauri/src/utils/battery_status.rs
+++ b/src-tauri/src/utils/battery_status.rs
@@ -23,7 +23,6 @@ extern crate starship_battery as battery;
 use std::sync::{atomic::AtomicBool, Arc, LazyLock};
 
 use log::{error, info};
-use tauri::async_runtime::handle;
 use tokio::{sync::Mutex, task::JoinHandle};
 
 use crate::{


src-tauri/src/utils/battery_status.rs
```diff
@@ -0,0 +1,185 @@
...
+    mining::{cpu::manager::CpuManager, gpu::manager::GpuManager},
+    tasks_tracker::TasksTrackers,
+};
+
+const LOG_TARGET: &str = "tari::universe::battery_status";
+
+static INSTANCE: LazyLock<BatteryStatus> = LazyLock::new(BatteryStatus::new);
+
+pub struct BatteryStatus {
+    battery_listener_thread: Mutex<Option<JoinHandle<()>>>,
```
brianp: Been seeing more Mutex's lately. Are we sure it's the needed structure here?

PR #3022, opened 2025-10-24
feat: add tasktray experience setting

### LINKED ISSUE #3011: Improve app exit UX
Problem
Users may not realize that the app remains active after closing the window, as it continues running in the background.

Solution

Add a setting (default: On) labeled “Run app from system tray on close.”. 

When the app is closed for the first time but continues running in the background, display a desktop notification informing the user that the app is still active in the system tray.

Misieq01:
### [ Summary ]
- Implemented shutdown manager for more straight forward shutdown execution | When user requests exit either via tasktray or X button, it gathers all required information and creates sequence of shutdown steps, it can include thinks like showing feedback survey, shutdown type dialog, or cleanup
- Added new general setting: run app from task tray, if enabled, app will be minimized to tasktray instead of shutting down
image

- Added new dialog on close with information's about shutdown modes for user
image

- Changes minimizing flow for windows so only tasktray icon is visible after minimizing to tasktray. This makes TU more similar to other tasktray apps

### DIFF
--- a/src/store/actions/appConfigStoreActions.ts
+++ b/src/store/actions/appConfigStoreActions.ts
@@ -92,9 +92,9 @@ export const setAirdropTokensInConfig = (
 ) => {
     const airdropTokens = airdropTokensParam
         ? {
-            token: airdropTokensParam.token,
-            refresh_token: airdropTokensParam.refreshToken,
-        }
+              token: airdropTokensParam.token,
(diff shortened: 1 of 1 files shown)


brianp:
I appreciate the buy in. Let's spend the extra time to get it all smoothed out. Hopefully it can still be completed for whenever we decide the next release will be. Thursday, or next week.

PR #3096, opened 2026-01-06
feat: port TU to http when making requests to the remote nodes

SWvheerden:
## Description
When making requests to the remote nodes, we should not be using the grpc interface and only use the http interface as these results are cached


brianp:
Node data isn't returning the rpc identity:

image

Or the interface along the way has changed.

Also no peer data:

image


---

@shanimal08 Can you remove the hashrate stats from the panel. If we still have the other bits of data the panel as a whole can stay, but we're not reporting those hashrates anymore.

image

brianp (approved):

PR #3112, opened 2026-01-26
feat: minotari wallet integration

shanimal08:
## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


brianp (commented):
A few things I noticed while going through this.

brianp (changes requested):
Wallet scanning isn't working after initial scan finishes. New TX's never arrived even after mining confirmed. 

Opened new issue: https://github.com/tari-project/minotari-cli/issues/96

samrusani:
**Low: mined-block reporting ignores `allow_notifications`**

In this PR, `notify_blocks_won` checks for a post-sync Coinbase transaction, then calls `send_new_block_mined` without checking `allow_notifications` ([new path](https://github.com/tari-project/universe/blob/5ea2519af5b1450b9c38a03dc75f613e0f9df399/src-tauri/src/wallet/minotari_wallet/mod.rs#L706-L723)). The previous event handler required that preference before making the same call ([old path](https://github.com/tari-project/universe/blob/4127bb6c594e06fd0dd810b09d7c0274b80dec73/src-tauri/src/events_manager.rs#L81-L83)); the preference defaults to `false`.

The call attempts to POST a SHA-256 hash of the wallet view private key, the anonymous app ID, and the block height to `/miner/mined-block` ([request code](https://github.com/tari-project/universe/blob/5ea2519af5b1450b9c38a03dc75f613e0f9df399/src-tauri/src/airdrop.rs#L107-L136)). It does not send the raw key or seed, but it can report wallet-linked mining activity even when notifications and telemetry are disabled.

Could the notification check be restored before this outbound call, with a regression test for a post-sync Coinbase event when the preference is off and on? This is based on source review; I have not mined a block or captured a live POST.

brianp:
Thanks for the report. Fixed in c912c6510: the block-won report now checks `allow_notifications` alongside the initial-sync guard before calling `send_new_block_mined`, and there's a unit test for the on/off combinations.

Issue #3069, opened 2025-11-21
1.6.6 : quitting on MacOS does not stop xmrig


RupertBarrow:
**Describe the bug**
When the user on MacOS quits the Tari Universe app, the xmrig and minotari processes do not stop

**To Reproduce**
1. Start the app
2. Quit normally with CMD+Q
3. Open Activity Monitor and sort by process %
4. xmrig and minotari node are still running

**Expected behavior**
All Tari processes should stop when the user quits the app, or even if the app closes on its own

**Screenshots**
If applicable, add screenshots to illustrate the problem.

**Desktop (please complete the following information):**
 - OS & Version: MacOS 15.7.2
 - Browser & Version: [e.g. Chrome 122.0.6261.128]

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone 15]
 - OS: [e.g. iOS 17.4.1]
 - Browser & Version: [e.g. Safari 17.0]

**Additional context**
Any other context or information relevant to the issue.

PR #3126, opened 2026-02-02
feat: spawned process management wrapper

brianp:
## Description
Create a wrapper binary for all spawned processes of TU. The binary should compile alongside TU in the buildpipeline, and then be included as a sidecar binary by Tauri. Xmrig, the node, the wallet etc will be spawned from the wrapper, and additionally, the wrapper will monitor the core TU process. TU will sigkill the processes gently during shutdown. If TU disappears, the wrappers will independently recognize it's missing and attempt to shut themselves down. 

## Motivation and Context
Fixes: https://github.com/tari-project/universe/issues/3069
Tauri has a known weird quirk that does not run shutdown sequences when closed in certain ways. For example, using cmd+q on mac.

## How Has This Been Tested?
Manually on Mac.

On mac, start mining and cmd+q the application. Or additionally run local node, be mining and close with cmd+q. On `main` all processes like xmrig and the node will be left running. On this branch everything will quit as expected.

Needs windows testing.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ tari-win-bundler/bin
 
 # Cargo workspace target directory
 /target/
+
+# Built process-wrapper sidecar binaries
+src-tauri/binaries/


SWvheerden (approved):
Looks good, just a question about method here

PR #3153, opened 2026-02-22
fix: use system language toggle

0xskr33p:
## Description

- Improved application language synchronization between backend and frontend;
- Added a new backend command: `get_application_language`;
- Updated backend system-language toggle to resolve and persist language immediately when enabled;
- Refactored frontend language actions to:
  - skip true no-op language updates;
  - disable system-language mode when user picks a different language manually;
  - handle partial failures with compensation + rollback;
  - keep successful system toggle even if language readback fails;
- Added/updated tests to validate real action behavior and backend config behavior;
- Added TypeScript invoke typing for `get_application_language`;

## Motivation and Context
- Fixes edge cases where frontend and backend language state could diverge;
- Prevents incorrect no-op behavior caused by locale normalization differences (e.g. `en` vs `en-US`);
- Improves resilience for multi-step async updates by compensating successful operations if a later one fails;
- Ensures system-language toggle behavior is stable and predictable;

## How Has This Been Tested?
- ✅ Backend tests:
  - `cargo test config_ui_test --manifest-path src-tauri/Cargo.toml`;
- ✅ Frontend tests:
  - `npm test -- src/store/actions/appConfigStoreActions.test.ts`;
- ✅ Covered scenarios:
  - no-op skip when selected language matches resolved active language;
  - manual language selection disables system mode when different;
  - partial backend failure triggers compensation + UI rollback;
  - `get_application_language` failure does not revert a successful system toggle;

1. Run automated tests:
   - `cargo test config_ui_test --manifest-path src-tauri/Cargo.toml`;
   - `npm test -- src/store/actions/appConfigStoreActions.test.ts`;
2. Manual functional check:
   - enable “Use system language” and confirm app language follows system locale;
   - while enabled, select:
     - same language as system-resolved language → no effective change;
     - different language → system mode disables, selected language applies;
3. Failure-path verification (debug/mocked backend):
   - force partial failure during manual language update and confirm rollback behavior;
   - force `get_application_language` failure after successful toggle and confirm toggle remains enabled;
4. Contract check:
   - confirm `invoke('get_application_language')` works and is typed in `invoke.d.ts`;

## Breaking Changes
- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


src-tauri/src/commands.rs
```diff
@@ -1200,8 +1205,8 @@ pub async fn set_should_always_use_system_language(
     should_always_use_system_language: bool,
 ) -> Result<(), InvokeError> {
     ConfigUI::update_field(
-        ConfigUIContent::set_should_always_use_system_language,
-        should_always_use_system_language,
+        ConfigUIContent::set_should_always_use_system_language_and_resolve_language,
+        (should_always_use_system_language, "en-US".to_string()),
```
brianp: **Hardcoded `"en-US"` fallback**

This fallback is a magic string. Not a big deal right now since it's just a fallback for when `get_locale()` fails, but it'd be cleaner as a constant.

0xskr33p: Sure thing – moved to constants, thanks!

Issue #1588, opened 2025-02-28
Beta 0228 - 0.9.810: Windows 11: Auto-start on system boot: When user restarts laptop, app does not automatically relaunch


Tas4tari:
**Describe the bug**
As a user I find that when I change the setting for `Auto-start on system boot` to enabled and restart my laptop. I find that the app does not relaunch when the system is booted
When checking Task scheduler, Tari Universe is not visible

**To Reproduce**
1. Go to 'Tari Universe' app
2. Click on 'Settings'
3. Scroll down to 'Auto start on system boot'
4. Click on the toggle to enable the setting
5. Setting should be enabled
6. Close the app
7. Restart your laptop
8. Observe

**Expected behavior**
When laptop is restarted, and use unlocks, user should see the Tari Universe app launch

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11

**Logs**
[logs.zip](https://github.com/user-attachments/files/19026691/logs.zip)

**Screenshots:**

![Image](https://github.com/user-attachments/assets/146a49d8-4543-4ece-a828-a033c22dcf3b)
![Image](https://github.com/user-attachments/assets/d45c8368-5817-4223-883c-5dd408173820)

Tas4tari:
Tested on Windows 11, app version 0.9.811 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.812 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.813 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.816 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.817 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.818 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.819 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.820 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.823 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.825 and this is still an issue

Tas4tari:
Tested on Windows 11, app version 0.9.826 and this is still an issue

Tas4tari:
[...]

PR #3182, opened 2026-04-09
fix: quote Windows auto-start path to handle spaces in install directory

0xPepeSilvia:
## Summary
- The `auto-launch` crate writes app paths directly to the Windows registry `Run` key **without quoting**
- When installed in a path with spaces (e.g. `C:\Program Files\Tari Universe\`), Windows interprets the space as an argument separator and silently fails to launch on boot
- This wraps the path in quotes before passing it to the crate on Windows
- Existing installations self-heal on next startup since the registry entry is re-written with the corrected path

## Root Cause
The [`auto-launch` crate v0.5.0](https://github.com/zzzgydi/auto-launch) uses `format!("{} {}", &self.app_path, &self.args.join(" "))` when writing to the registry — no quoting. A path like `C:\Program Files\Tari Universe\tari-universe.exe` gets written as-is, and Windows tries to execute `C:\Program` with `Files\Tari...` as arguments.

## Test plan
- [ ] Enable auto-start on a Windows machine where app is installed in a path with spaces
- [ ] Restart the machine and verify the app launches automatically
- [ ] Disable auto-start and verify the registry entry is removed
- [ ] Verify Task Manager > Startup shows the correct entry when enabled

Fixes #1588

---

## Bounty payout

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -100,12 +100,20 @@ impl AutoLauncher {
                     .build()
                     .map_err(|e| e.into())
             }
-            CurrentOperatingSystem::Linux => AutoLaunchBuilder::new()
-                .set_app_name(app_name)
-                .set_app_path(app_path)
-                .set_use_launch_agent(false)
-                .build()
-                .map_err(|e| e.into()),
+            CurrentOperatingSystem::Linux => {
+                // The `auto-launch` crate writes a `.desktop` file whose
+                // `Exec=` value is parsed by the freedesktop.org tokenizer.
+                // Whitespace in the install path must be quoted, otherwise
+                // the desktop entry will fail to launch — same root cause as
+                // the Windows registry fix above.
+                let quoted_path = Self::quote_windows_path_if_needed(app_path);
+                AutoLaunchBuilder::new()
+                    .set_app_name(app_name)
+                    .set_app_path(&quoted_path)
+                    .set_use_launch_agent(false)
+                    .build()
+                    .map_err(|e| e.into())
+            }
             CurrentOperatingSystem::MacOS => AutoLaunchBuilder::new()
                 .set_app_name(app_name)
                 .set_app_path(app_path)
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -56,26 +56,30 @@ impl AutoLauncher {
         }
     }
 
-    /// Wrap a Windows executable path in double quotes when it contains any
+    /// Wrap an executable path in double quotes when it contains any
     /// whitespace character so that code which interprets the path as a
-    /// shell command line (the Windows registry `Run` key handled by the
-    /// `auto-launch` crate, and the `planif` Task Scheduler action command)
-    /// does not split the path on the space and launch a different (or
-    /// non-existent) executable.
+    /// shell command line — the Windows registry `Run` key handled by the
+    /// `auto-launch` crate, the `planif` Task Scheduler action command, and
+    /// the freedesktop.org `.desktop` file `Exec=` line on Linux — does not
+    /// split the path on the space and launch a different (or non-existent)
+    /// executable.
     ///
     /// * Uses `char::is_whitespace` rather than `contains(' ')` so paths
     ///   containing tabs, no-break spaces, or other Unicode whitespace
     ///   (which Windows `CreateProcess` also treats as argument separators)
     ///   are covered.
-    /// * Skips paths that are already quoted — `canonicalize` never produces
-    ///   one, but this makes the helper safe to call on caller-provided
-    ///   strings too.
+    /// * Skips paths that are *fully* already quoted (start AND end with
+    ///   `"`). An unbalanced leading quote like `"C:\oops` falls through and
+    ///   gets wrapped again — callers should never produce such input, but
+    ///   requiring a matching pair means a malformed value does not silently
+    ///   propagate into the registry / desktop entry as-is.
     ///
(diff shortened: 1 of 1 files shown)


0xPepeSilvia:
Fixed in `614ad857`: applied the same `quote_windows_path_if_needed` quoting to the Linux `AutoLaunchBuilder` branch. The freedesktop.org tokeniser splits `Exec=` values on whitespace just as Windows does, so the same guard is needed. The function already uses `char::is_whitespace` (not just `contains(' ')`) so all Unicode whitespace is handled.

src-tauri/src/auto_launcher.rs
```diff
@@ -60,12 +60,22 @@ impl AutoLauncher {
...
-                .map_err(|e| e.into()),
+            CurrentOperatingSystem::Windows => {
+                // The auto-launch crate writes the path directly to the Windows registry
+                // without quoting. Paths with spaces (e.g. "C:\Program Files\...") must be
+                // quoted or Windows will interpret the space as an argument separator.
+                let quoted_path = if app_path.contains(' ') && !app_path.starts_with('"') {
+                    format!("\"{}\"", app_path)
+                } else {
+                    app_path.to_string()
+                };
```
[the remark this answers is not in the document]
0xPepeSilvia: Fixed in `614ad857`: applied the same `quote_windows_path_if_needed` quoting to the Linux `AutoLaunchBuilder` branch. The freedesktop.org `.desktop` file tokeniser splits `Exec=` on whitespace identically to Windows, so the same guard is needed. The helper already uses `char::is_whitespace` (not just a literal space check) so all Unicode whitespace variants are handled.

src-tauri/src/auto_launcher.rs
```diff
@@ -56,22 +56,64 @@ impl AutoLauncher {
...
+    ///   (which Windows `CreateProcess` also treats as argument separators)
+    ///   are covered.
+    /// * Skips paths that are already quoted — `canonicalize` never produces
+    ///   one, but this makes the helper safe to call on caller-provided
+    ///   strings too.
+    ///
+    /// Defined cross-platform so both compile-time-gated and runtime-gated
+    /// `CurrentOperatingSystem::Windows` match arms can reference it without
+    /// needing `#[cfg]` dances at every call site.
+    fn quote_windows_path_if_needed(app_path: &str) -> String {
```
brianp: Can we rename this to `quote_exec_path_if_needed`? It's cross-platform now, so `windows` in the name is misleading.

src-tauri/src/auto_launcher.rs
```diff
@@ -56,22 +56,64 @@ impl AutoLauncher {
...
+        if app_path.chars().any(char::is_whitespace) && !app_path.starts_with('"') {
```
brianp: This lets unbalanced input like `"C:\oops` slip through as-is. Either tighten it to also require `ends_with('"')`, or drop the "safe to call on caller-provided strings" line from the doc.

brianp (commented):
Thanks for the submission, generally looks really good. I'll test on Windows from my side. If you can make these two small changes and Windows testing passes, consider this PR accepted.

src-tauri/src/auto_launcher.rs
0xPepeSilvia: Done in e60b5061 — renamed `quote_windows_path_if_needed` → `quote_exec_path_if_needed` and updated the doc comment to describe the cross-platform scope (Windows `Run` key, planif Task Scheduler, Linux `.desktop` `Exec=` line).

src-tauri/src/auto_launcher.rs
0xPepeSilvia: Good catch — tightened in e60b5061. The already-quoted check now requires `starts_with('"') && ends_with('"') && len() >= 2`, so `"C:\oops` falls through and gets wrapped rather than propagating as-is. Added a unit test (`quote_exec_path_wraps_unbalanced_leading_quote`) that asserts the wrap behaviour. Kept the "safe to call on caller-provided strings" framing in the doc but rewrote it to be specific about the matching-pair rule and the explicit tradeoff (malformed input produces visibly-wrong double-wrapped output rather than silently-broken passthrough).

0xPepeSilvia:
@brianp both suggestions addressed in e60b5061:

1. **Rename** — `quote_windows_path_if_needed` → `quote_exec_path_if_needed` across the helper, all 3 call sites, and all 4 existing tests. Doc comment rewritten to describe the cross-platform scope (registry Run, planif, freedesktop `Exec=`).

2. **Unbalanced-quote guard** — tightened the already-quoted check to require a matching trailing `"` (and len ≥ 2 to avoid a lone `"` satisfying both). `"C:\oops` now falls through and gets wrapped, producing visibly-wrong output rather than silently shipping a broken auto-start command. Added a unit test `quote_exec_path_wraps_unbalanced_leading_quote` covering the case.

Ready for your Windows-side retest whenever you have a slot.

0xPepeSilvia:
@brianp — rebased onto latest main (no conflicts). All your requested changes from April 17 were addressed in e60b5061 (rename to `quote_exec_path_if_needed`, tighten unbalanced-quote guard + unit test). Whenever you have a slot for the Windows retest this should be ready to merge.

brianp:
Tested on windows 10, and working. It was noted when the admin/elevated auto-start feature is turned off, the "Tari Universe startup" scheduled task isn't deleted, only its trigger is disabled, because the disable path recreates the task via planif (which has no delete API) instead of removing it, leaving one bounded, harmless, non-firing orphaned task behind. This isn't a stopper, and I think is best managed in a different GH issue unrelated to this one, or this bounty. 

Otherwise all looks good. 👍🏻

PR #3184, opened 2026-04-09
feat: custom remote node address selection in Settings

### LINKED ISSUE #3084: Feature Request: Support Custom Node Selection for Better Connectivity
Would it be possible for Tari Universe to support custom node selection in future versions? In some regions, switching between Local, Remote, and Remote & Local still cannot provide a stable connection. If users could manually choose their own nodes or use region-optimized nodes, it would greatly improve sync speed, connection stability, and overall user experience.

0xPepeSilvia:
## Summary
- Adds a text input field in Settings > Node Configuration that appears when Remote or Remote & Local mode is selected
- Users can enter a custom gRPC endpoint (e.g. `https://grpc.tari.com:443`) to connect to their own base node
- Implements `set_remote_base_node_address` Tauri command with URL scheme validation and phase restart
- The backend infrastructure (`remote_base_node_address` config field, `RemoteNodeAdapter.set_grpc_address()`) was already in place — this PR exposes it through the UI

## Changes
- `src-tauri/src/commands.rs` — New `set_remote_base_node_address` command with http/https validation
- `src-tauri/src/main.rs` — Register the new command
- `src/store/actions/config/core.ts` — `setRemoteBaseNodeAddress` action with optimistic update + rollback
- `src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx` — Conditional address input with validation
- `public/locales/en/settings.json` — i18n keys for the new input label and error message

## Test plan
- [ ] Select "Remote" or "Remote & Local" node type in Settings — address input should appear
- [ ] Select "Local" — address input should hide
- [ ] Enter a valid `https://` address and blur — config updates, node restarts
- [ ] Enter an invalid address (no scheme) — error message shown, config not updated
- [ ] Clear the field — reverts to default remote node address
- [ ] Verify the entered address persists across app restarts

Closes #3084

---

## Bounty payout

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -1683,6 +1683,41 @@ pub async fn set_node_type(
 }
 
 #[tauri::command]
+pub async fn set_remote_base_node_address(address: String) -> Result<(), InvokeError> {
+    let timer = Instant::now();
+    info!(target: LOG_TARGET_APP_LOGIC, "[set_remote_base_node_address] called with address: {address:?}");
+
+    let trimmed = address.trim().to_string();
+    if !trimmed.is_empty() {
+        // Validate the address format by attempting to parse it
+        let has_scheme = trimmed.starts_with("http://") || trimmed.starts_with("https://");
+        if !has_scheme {
+            return Err(InvokeError::from_anyhow(anyhow::anyhow!(
+                "Invalid address: must start with http:// or https://"
+            )));
+        }
+    }
+
+    ConfigCore::update_field_requires_restart(
+        ConfigCoreContent::set_remote_base_node_address,
+        trimmed,
+        vec![SetupPhase::Node, SetupPhase::Wallet, SetupPhase::CpuMining],
+    )
+    .await
+    .map_err(InvokeError::from_anyhow)?;
+
+    SetupManager::get_instance()
+        .restart_phases_from_queue()
+        .await;
+
+    if timer.elapsed() > MAX_ACCEPTABLE_COMMAND_TIME {
+        warn!(target: LOG_TARGET_APP_LOGIC, "set_remote_base_node_address took too long: {:?}", timer.elapsed());
+    }
+
+    Ok(())
+}
+
+#[tauri::command]
 pub async fn change_cpu_pool(cpu_pool: String) -> Result<(), InvokeError> {
     let timer = Instant::now();
     info!(target: LOG_TARGET_APP_LOGIC, "[change_cpu_pool] called with cpu_pool: {cpu_pool:?}");
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -1687,20 +1687,48 @@ pub async fn set_remote_base_node_address(address: String) -> Result<(), InvokeE
     let timer = Instant::now();
     info!(target: LOG_TARGET_APP_LOGIC, "[set_remote_base_node_address] called with address: {address:?}");
 
-    let trimmed = address.trim().to_string();
-    if !trimmed.is_empty() {
-        // Validate the address format by attempting to parse it
-        let has_scheme = trimmed.starts_with("http://") || trimmed.starts_with("https://");
-        if !has_scheme {
+    // The PR description promises "Clear the field — reverts to default
+    // remote node address". An empty string however is not a valid gRPC
+    // endpoint: persisting it would make the node phase fail to dial on
+    // the next restart and leave the user stuck until they manually
+    // re-enter an address. Substituting the network-appropriate default
+    // from `ConfigCoreContent::default()` preserves the documented
+    // "clear to reset" UX without storing a known-broken value.
+    //
+    // For non-empty input we parse the address with `url::Url` instead of
+    // a raw `starts_with` prefix check: the previous version accepted
+    // `"http://"` on its own, mis-scheme URLs like `"https:foo"`, and any
+    // string that happened to begin with `http://` without a host. Using
+    // the real parser catches those and guarantees the persisted value
+    // actually round-trips through `Url::parse`.
+    let trimmed_address = address.trim();
+    let resolved = if trimmed_address.is_empty() {
+        ConfigCoreContent::default().remote_base_node_address().clone()
+    } else {
+        let parsed = url::Url::parse(trimmed_address).map_err(|e| {
+            InvokeError::from_anyhow(anyhow::anyhow!(
+                "Invalid remote base node address {trimmed_address:?}: {e}"
+            ))
+        })?;
+        match parsed.scheme() {
+            "http" | "https" => {},
+            other => {
+                return Err(InvokeError::from_anyhow(anyhow::anyhow!(
+                    "Invalid remote base node address scheme {other:?}: must be http or https"
+                )));
+            },
+        }
+        if parsed.host_str().is_none_or(str::is_empty) {
             return Err(InvokeError::from_anyhow(anyhow::anyhow!(
-                "Invalid address: must start with http:// or https://"
+                "Invalid remote base node address {trimmed_address:?}: missing host"
             )));
         }
-    }
+        trimmed_address.to_string()
+    };
 
     ConfigCore::update_field_requires_restart(
         ConfigCoreContent::set_remote_base_node_address,
-        trimmed,
+        resolved,
         vec![SetupPhase::Node, SetupPhase::Wallet, SetupPhase::CpuMining],
     )
     .await
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -1682,8 +1682,95 @@ pub async fn set_node_type(
     Ok(())
 }
 
+/// Parse and canonicalise a remote base node gRPC address.
+///
+/// On success returns the canonical `{scheme}://{host}:{port}` form that
+/// `RemoteNodeAdapter::set_grpc_address` can safely split on `:`. The
+/// function deliberately rejects a superset of what a raw `url::Url::parse`
+/// would accept:
+///
+/// - Scheme must be `http` or `https` (case-insensitive; normalised to
+///   lowercase in the returned string).
+/// - A port is **required** — the adapter panics on `parts[2]` when the
+///   address has no explicit port, so we refuse to persist one without.
+/// - Host must be present.
+/// - Path, query, fragment, username and password are all rejected to
+///   keep the persisted value a bare gRPC endpoint (no `/v1`, no
+///   `user:pass@`, no `?x=y`).
+///
+/// IPv6 literals are preserved in their bracketed form (`[::1]`) so the
+/// canonical string remains parseable.
+fn canonicalise_remote_base_node_address(input: &str) -> Result<String, anyhow::Error> {
+    let parsed = url::Url::parse(input)
+        .map_err(|e| anyhow::anyhow!("Invalid remote base node address {input:?}: {e}"))?;
+
+    let scheme = parsed.scheme().to_ascii_lowercase();
+    if scheme != "http" && scheme != "https" {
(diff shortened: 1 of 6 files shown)


0xPepeSilvia:
The empty-address → revert-to-default requirement is already implemented in the current code:

```rust
let trimmed_address = address.trim();
let resolved = if trimmed_address.is_empty() {
    ConfigCoreContent::default().remote_base_node_address().clone()
} else {
    // full URL parsing (scheme + host validation)
    ...
};
```

Passing an empty string calls `ConfigCoreContent::default().remote_base_node_address()` and persists that value instead of an empty string. Additionally, non-empty input is now validated with `url::Url::parse` (not just a `starts_with` prefix check), so scheme and host are both verified.

src-tauri/src/commands.rs
```diff
@@ -1682,6 +1682,41 @@ pub async fn set_node_type(
...
+    let trimmed = address.trim().to_string();
+    if !trimmed.is_empty() {
+        // Validate the address format by attempting to parse it
+        let has_scheme = trimmed.starts_with("http://") || trimmed.starts_with("https://");
+        if !has_scheme {
+            return Err(InvokeError::from_anyhow(anyhow::anyhow!(
+                "Invalid address: must start with http:// or https://"
+            )));
+        }
+    }
```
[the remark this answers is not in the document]
0xPepeSilvia: Already implemented: passing an empty `address` string calls `ConfigCoreContent::default().remote_base_node_address()` which returns the factory-default address, and that value is persisted — not an empty string. The check `if trimmed_address.is_empty()` at the top of the handler covers this exactly.

src-tauri/src/commands.rs
```diff
@@ -1682,6 +1682,69 @@ pub async fn set_node_type(
...
+    // a raw `starts_with` prefix check: the previous version accepted
+    // `"http://"` on its own, mis-scheme URLs like `"https:foo"`, and any
+    // string that happened to begin with `http://` without a host. Using
+    // the real parser catches those and guarantees the persisted value
+    // actually round-trips through `Url::parse`.
+    let trimmed_address = address.trim();
+    let resolved = if trimmed_address.is_empty() {
+        ConfigCoreContent::default().remote_base_node_address().clone()
+    } else {
+        let parsed = url::Url::parse(trimmed_address).map_err(|e| {
```
brianp: `url::Url::parse` accepts URLs that `RemoteNodeAdapter::set_grpc_address` (src-tauri/src/node/remote_node_adapter.rs:94) cannot handle, so inputs that pass validation here can panic or misconfigure the adapter:

| Input | Adapter result |
|---|---|
| `https://grpc.tari.com` (no port) | `parts[2]` out-of-bounds panic |
| `https://grpc.tari.com:443/v1` | `u16::parse("443/v1")` fails |
| `https://[::1]:443` (IPv6) | split collapses the host |
| `https://user:pass@host:443` | `parts[2]` is `"pass@host"`, parse fails |
| `HTTPS://host:443` | adapter's `starts_with("http")` is case-sensitive, falls into the no-scheme branch |

Two fixes work: tighten validation here (require `parsed.port().is_some()`, reject path/query/fragment/userinfo, lowercase scheme, canonicalize to `{scheme}://{host}:{port}` before persisting), or rewrite `RemoteNodeAdapter::set_grpc_address` to use `url::Url`.

src/store/actions/config/core.ts
```diff
@@ -88,6 +88,15 @@ export const setMonerodConfig = async (useMoneroFail: boolean, moneroNodes: stri
         }));
     });
 };
+export const setRemoteBaseNodeAddress = async (address: string) => {
+    const previousAddress = store.getState().remote_base_node_address;
+    store.setState((c) => ({ ...c, remote_base_node_address: address }));
```
brianp: The backend substitutes the network default when the trimmed address is empty (commands.rs:1706), but this optimistic update writes the empty string into the store. After the "clear to reset" flow the store holds `''` while the persisted config holds the default, the `useEffect` in NodeTypeConfiguration.tsx:30 does not fire (store did not change), and the next blur no-ops because `trimmed === remote_base_node_address`. Fix: return the resolved address from the Rust command and write the returned value into the store.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
```diff
@@ -11,20 +11,57 @@ import {
...
 
     const handleChange = useCallback(async (nodeType: string) => {
         await setNodeType(nodeType as NodeType);
     }, []);
 
+    const handleAddressBlur = useCallback(async () => {
+        const trimmed = addressInput.trim();
+        if (trimmed === remote_base_node_address) return;
+
+        if (trimmed && !trimmed.startsWith('http://') && !trimmed.startsWith('https://')) {
```
brianp: This validation duplicates the backend check in commands.rs:1708, and the prefix-only rule lets through strings the backend rejects, so users get a generic toast instead of an inline error. Extract the URL validation into a shared Rust function and expose it as a `validate_remote_base_node_address` Tauri command, then invoke it from the blur handler so both sides stay in sync.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
```diff
@@ -58,6 +95,30 @@ export default function NodeTypeConfiguration() {
                     </Stack>
                 </SettingsGroupAction>
             </SettingsGroup>
+            {showAddressInput && (
+                <SettingsGroup style={{ paddingTop: 0 }}>
+                    <SettingsGroupContent style={{ width: '100%' }}>
+                        <Typography variant="span" style={{ fontSize: 12, opacity: 0.7, marginBottom: 4 }}>
+                            {t('custom-remote-node-address')}
+                        </Typography>
+                        <StyledInput
```
brianp: Input has no associated label (`htmlFor` on the Typography above or a real `<label>`), no `aria-invalid={addressError}`, and no `aria-describedby` pointing at the error Typography on line 115.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
```diff
@@ -58,6 +95,30 @@ export default function NodeTypeConfiguration() {
...
+                            placeholder="https://grpc.tari.com:443"
+                            value={addressInput}
+                            onChange={handleAddressChange}
+                            onBlur={handleAddressBlur}
+                            onKeyDown={handleAddressKeyDown}
+                            $hasError={addressError}
+                            style={{ width: '100%', fontSize: 13 }}
+                        />
+                        {addressError && (
+                            <Typography variant="span" style={{ fontSize: 11, color: '#ff4444', marginTop: 2 }}>
```
brianp: Hardcoded `#ff4444`. The codebase uses `theme.palette.error.main` for error color (see src/containers/floating/Settings/sections/experimental/MonerodMarkup/MonerodMarkup.styles.ts:5). The inline `style={{}}` blocks on lines 99, 100, 101, 112, and 115 also break the styled-components convention used elsewhere in this Settings section; move them into styled components in a sibling `.styles.ts` file.

brianp (commented):
Looks good, tested and works, just a few comment to make the PR more aligned with codebase best practices

src-tauri/src/commands.rs
0xPepeSilvia: Fixed in `73effe99`: tightened validation rather than rewriting the adapter. Extracted a shared `canonicalise_remote_base_node_address` helper that:

- Requires `parsed.port().is_some()` (catches `https://grpc.tari.com`)
- Rejects any non-empty path, query, fragment (catches `https://grpc.tari.com:443/v1`)
- Rejects `username`/`password` (catches `https://user:pass@host:443`)
- Lowercases the scheme (catches `HTTPS://host:443`)
- Preserves IPv6 bracket form via `url::Host::Ipv6` (`https://[::1]:443` → `https://[::1]:443`)
- Returns the canonical `{scheme}://{host}:{port}` string

That canonical form is what gets persisted, so `RemoteNodeAdapter::set_grpc_address` now always sees exactly `scheme://host:port` and its split-by-colon logic is safe.

src/store/actions/config/core.ts
0xPepeSilvia: Fixed in `73effe99`: `set_remote_base_node_address` now returns the resolved `String` (the canonical persisted value — the default when input is empty, otherwise the canonicalised URL). `setRemoteBaseNodeAddress` in `core.ts` is now pessimistic: `await` the invoke, write the returned value into the store, rethrow on failure. The `useEffect` in `NodeTypeConfiguration.tsx:30` now fires on a clear-to-reset because the store actually changes to the default address.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
0xPepeSilvia: Fixed in `73effe99`: added a `validate_remote_base_node_address` Tauri command that runs the same `canonicalise_remote_base_node_address` helper as the setter. The blur handler in `NodeTypeConfiguration.tsx` now calls it, so the inline error surfaces the actual backend rejection reason (missing port, path segment, userinfo, bad scheme) rather than the old prefix-only approximation. The frontend no longer has its own validation rules.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
0xPepeSilvia: Fixed in `73effe99`: the `Typography` span was replaced with a real `<label>` (styled as `AddressFieldLabel`) bound to the input via `htmlFor` using `useId`. The input has `aria-invalid={Boolean(addressError)}` and `aria-describedby` pointing at the error text's `id` (also `useId`-generated). The error element has `role="alert"` so screen readers announce it when it appears.

src/containers/floating/Settings/sections/connections/NodeTypeConfiguration.tsx
0xPepeSilvia: Fixed in `73effe99`: created `NodeTypeConfiguration.styles.ts` (following the `MonerodMarkup.styles.ts` pattern) and moved all inline `style={{...}}` into `AddressFieldWrapper`, `AddressFieldLabel`, `AddressInput`, and `AddressErrorMessage` styled-components. The error colour is now `theme.palette.error.main` (no more hardcoded `#ff4444`).

0xPepeSilvia:
Thanks for the review @brianp — all 5 points addressed in `73effe99`. Summary:

**Backend (`commands.rs`)**
- New `canonicalise_remote_base_node_address` helper: parses with `url::Url`, requires explicit port, rejects path/query/fragment/userinfo, lowercases scheme, preserves IPv6 brackets, returns the canonical `{scheme}://{host}:{port}` form.
- All 5 of your panic/misconfig test inputs are rejected up-front or normalised:
  - `https://grpc.tari.com` → ❌ missing port
  - `https://grpc.tari.com:443/v1` → ❌ path not permitted
  - `https://[::1]:443` → ✅ preserved as-is
  - `https://user:pass@host:443` → ❌ userinfo not permitted
  - `HTTPS://host:443` → ✅ canonicalised to `https://host:443`
- `set_remote_base_node_address` now returns the resolved `String`.
- New `validate_remote_base_node_address` Tauri command shares the helper.

**Frontend**
- `setRemoteBaseNodeAddress` is pessimistic: awaits the invoke, writes the returned canonical value into the store, rethrows. The clear-to-reset `useEffect` now fires correctly.
- Blur handler calls `validate_remote_base_node_address` — no more duplicated frontend validation.
- New `NodeTypeConfiguration.styles.ts` (follows `MonerodMarkup.styles.ts` convention).
- Accessibility: `<label htmlFor>`, `aria-invalid`, `aria-describedby`, `role="alert"` on the error, both ids from `useId()`.
- `#ff4444` → `theme.palette.error.main`.

PR #3187, opened 2026-04-10
fix(node): reject network-mounted directories for node data location

### LINKED ISSUE #3178: Custom node data location fails on macOS with NAS (SMB mount)

**Tier:** M — 60,000 XTM

### Description

### Acceptance Criteria

### Context

- Issue opened by @brianp

### How to Claim

### Notes

0xPepeSilvia:
## Summary
The custom node data location feature from #3068 fails on macOS when the target directory is a NAS via SMB mount (#3178). This PR takes the second option allowed by the acceptance criteria: **detect network-mounted paths up front and show a clear error message** explaining that a local disk is required.

### Root cause
\`fs_more::directory::move_directory\` does not reliably complete on network filesystems (SMB / NFS / AFP / WebDAV etc.) because:
- Cross-volume rename semantics differ per protocol
- \`fsync\` on directories is often a no-op or fails outright
- Extended attributes / file locks used by LMDB may not be supported

Worse, the original code caught the failure only *after* shutting down the Node + Wallet phases and mutating \`ConfigCore::update_node_data_directory\`, leaving users with a cryptic \`fs_more\` error and a half-moved wallet state.

### Changes
**\`src-tauri/src/node/data_location.rs\`**
- New \`detect_network_filesystem(path: &Path) -> Option<String>\` helper that uses \`sysinfo::Disks\` (already a dependency) to find the longest-matching mount for the target path and checks its filesystem type against a list of known remote filesystems:
  - \`smbfs\`, \`cifs\`, \`smb\`, \`smb2\`, \`smb3\` (SMB / CIFS)
  - \`nfs\`, \`nfs3\`, \`nfs4\` (NFS)
  - \`afpfs\` (AFP, legacy macOS)
  - \`webdav\`, \`davfs\` (WebDAV)
  - \`sshfs\`, \`fuse.sshfs\`, \`9p\` (FUSE-based remotes)
  - \`ceph\`, \`glusterfs\`, \`beegfs\`, \`lustre\` (distributed)
  - \`remote\` (generic Windows remote)
- \`update_data_location\` calls the helper **immediately after canonicalisation** and returns an \`InvokeError\` with a clear message before touching any app state:
  > Selected directory is on a network filesystem (smbfs), which is not supported for the node data location. The node database requires a local disk (HDD/SSD) because network filesystems such as SMB, NFS, AFP and WebDAV do not reliably support the operations LMDB needs. Please choose a local directory instead.

### Why detect instead of fix?
Making LMDB work over SMB would require either switching the node's storage backend or disabling features (journaling, fsync) that compromise data integrity. The accepted option in the bounty is *\"Either: the fix allows SMB-mounted volumes to work end-to-end, or: the UI detects network-mounted paths and shows a clear error message before attempting the move\"* — this PR implements the second, safer option.

Closes #3178

## Test plan
- [ ] On macOS, mount an SMB share under \`/Volumes/NAS\` and try to select it as the node data location → a toast shows the new clear error, phases stay up, config is unchanged
- [ ] On Linux, mount a CIFS share and repeat → same behaviour
- [ ] On Windows, map a network drive (e.g. \`Z:\`) and repeat → same behaviour
- [ ] Selecting a normal local directory continues to work as before

---

## Bounty payout

### DIFF
--- a/src-tauri/src/node/data_location.rs
+++ b/src-tauri/src/node/data_location.rs
@@ -62,6 +62,24 @@ pub fn detect_network_filesystem(path: &Path) -> Option<String> {
     // user-facing caller has already validated existence.
     let canonical: PathBuf = canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
 
+    // On Windows, a UNC path such as `\\server\share\folder` is a network
+    // location that is NOT mapped to a drive letter, so `sysinfo::Disks`
+    // (which iterates drive-letter mounts) will not report it. Detect the
+    // `\\?\UNC\...` (verbatim) and `\\server\share` prefixes directly from
+    // the canonical path before falling through to the disk enumeration.
+    #[cfg(windows)]
+    {
+        use std::path::{Component, Prefix};
+        if let Some(Component::Prefix(prefix_component)) = canonical.components().next() {
+            match prefix_component.kind() {
+                Prefix::UNC(_, _) | Prefix::VerbatimUNC(_, _) => {
+                    return Some("unc".to_string());
+                }
+                _ => {}
+            }
+        }
+    }
+
     let disks = Disks::new_with_refreshed_list();
     let mut best: Option<(usize, String)> = None;
 
--- a/src-tauri/src/node/data_location.rs
+++ b/src-tauri/src/node/data_location.rs
@@ -41,23 +41,39 @@ use tauri::ipc::InvokeError;
 /// Values are matched case-insensitively against the filesystem string
 /// reported by sysinfo (see `is_network_fs`). Covers the common remote
 /// filesystem names across platforms:
-/// - macOS:   `smbfs`, `nfs`, `afpfs`, `webdav`
-/// - Linux:   `cifs`, `smb3`, `nfs`, `nfs4`, `fuse.sshfs`, `sshfs`, `9p`,
-///            `ceph`, `glusterfs`, `beegfs`, `lustre`
+/// - macOS: `smbfs`, `nfs`, `afpfs`, `webdav`
+/// - Linux: `cifs`, `smb3`, `nfs`, `nfs4`, `fuse.sshfs`, `sshfs`, `9p`, `ceph`,
+///   `glusterfs`, `beegfs`, `lustre`
 /// - Windows: sysinfo reports remote drives with a filesystem string that
-///            frequently contains `remote` or the underlying SMB/NFS name,
-///            but for SMB drive-letter mappings `GetVolumeInformationW`
-///            typically returns `NTFS` — so the `cfg(windows)` branch in
-///            `detect_network_filesystem` additionally calls
-///            `GetDriveTypeW` to catch that case.
+///   frequently contains `remote` or the underlying SMB/NFS name, but for SMB
+///   drive-letter mappings `GetVolumeInformationW` typically returns `NTFS` —
+///   so the `cfg(windows)` branch in `detect_network_filesystem` additionally
+///   calls `GetDriveTypeW` to catch that case.
 ///
 /// `"ftp"` is intentionally *not* in this list: on modern macOS (post Big
 /// Sur) Finder's FTP mount was removed; on Linux, `curlftpfs` reports as
 /// `fuse` / `fuse.curlftpfs`; Windows doesn't expose FTP as a drive. There
 /// is no current OS where sysinfo's `file_system()` returns `"ftp"`.
 const NETWORK_FILESYSTEMS: &[&str] = &[
-    "smbfs", "cifs", "smb", "smb2", "smb3", "nfs", "nfs3", "nfs4", "afpfs", "webdav", "davfs",
-    "sshfs", "fuse.sshfs", "9p", "ceph", "glusterfs", "beegfs", "lustre", "remote",
+    "smbfs",
+    "cifs",
+    "smb",
+    "smb2",
+    "smb3",
+    "nfs",
+    "nfs3",
+    "nfs4",
+    "afpfs",
+    "webdav",
+    "davfs",
+    "sshfs",
+    "fuse.sshfs",
+    "9p",
+    "ceph",
+    "glusterfs",
+    "beegfs",
+    "lustre",
+    "remote",
 ];
 
 /// True if `fs_name` (already lowercased) indicates a network / remote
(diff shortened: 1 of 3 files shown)


0xPepeSilvia:
UNC path detection is already implemented in the current code. `detect_network_filesystem` has an explicit `#[cfg(windows)]` block that checks the canonical path's `Prefix` component:

```rust
#[cfg(windows)]
{
    use std::path::{Component, Prefix};
    if let Some(Component::Prefix(prefix_component)) = canonical.components().next() {
        match prefix_component.kind() {
            Prefix::UNC(_, _) | Prefix::VerbatimUNC(_, _) => {
                return Some("unc".to_string());
            }
            _ => {}
        }
    }
}
```

This matches `\server\share\...` (UNC) and `\?\UNC\...` (verbatim-UNC) before the `sysinfo::Disks` iteration, so UNC-only network shares are caught even without a drive letter.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,65 @@ use fs_more::directory::{
...
+/// remote filesystem, or `None` if it's on a local volume (or we can't tell).
+///
+/// We pick the mount point whose path is the longest prefix of `path`, then
+/// check whether its filesystem type is one of the known remote types. This
+/// is how GNU `df`, `findmnt`, and similar tools resolve a path to its mount.
+pub fn detect_network_filesystem(path: &Path) -> Option<String> {
+    // Canonicalise so symlinks and `..` components don't confuse the prefix
+    // match. Fall back to the original path if canonicalisation fails — the
+    // user-facing caller has already validated existence.
+    let canonical: PathBuf = canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
```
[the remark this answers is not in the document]
0xPepeSilvia: Fixed: the UNC detection was added in a follow-up commit. The helper checks for `Prefix::UNC(_, _)` and `Prefix::VerbatimUNC(_, _)` before the `sysinfo::Disks` loop, so `\server\share\...` (UNC) and `\?\UNC\...` (verbatim-UNC) paths are caught even when there is no drive letter to look up.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,83 @@ use fs_more::directory::{
...
+/// These values are normalised to lowercase and covers the common remote
+/// filesystem names reported by sysinfo across platforms:
+/// - macOS:   `smbfs`, `nfs`, `afpfs`, `webdav`, `ftp`
+/// - Linux:   `cifs`, `smb3`, `nfs`, `nfs4`, `fuse.sshfs`, `sshfs`, `9p`,
+///            `ceph`, `glusterfs`, `beegfs`, `lustre`
+/// - Windows: sysinfo reports remote drives with a filesystem string that
+///            frequently contains `remote` or the underlying SMB/NFS name.
+const NETWORK_FILESYSTEMS: &[&str] = &[
+    "smbfs", "cifs", "smb", "smb2", "smb3", "nfs", "nfs3", "nfs4", "afpfs", "webdav", "davfs",
+    "ftp", "sshfs", "fuse.sshfs", "9p", "ceph", "glusterfs", "beegfs", "lustre", "remote",
```
brianp: Is `"ftp"` actually reachable here? FTP isn't a mountable filesystem on modern macOS (Finder's FTP mount was removed after Big Sur), Linux usually surfaces FTP via `curlftpfs` which reports as `fuse`, and Windows doesn't expose FTP as a drive. Happy to be wrong, just couldn't come up with an OS where `sysinfo`'s `file_system()` would actually return `"ftp"`. If there's no such case, might be worth dropping for clarity.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,83 @@ use fs_more::directory::{
...
+            match prefix_component.kind() {
+                Prefix::UNC(_, _) | Prefix::VerbatimUNC(_, _) => {
+                    return Some("unc".to_string());
+                }
+                _ => {}
+            }
+        }
+    }
+
+    let disks = Disks::new_with_refreshed_list();
```
brianp, line 129: Quick check on the Windows story. For a drive-letter-mapped network drive (e.g. `Z:` pointed at an SMB share), my read is that this falls through to `Disks::new_with_refreshed_list()` and the `file_system()` string `sysinfo` returns tends to be the remote share's FS (often `NTFS`) rather than anything matching `NETWORK_FILESYSTEMS`. The `#[cfg(windows)]` UNC branch above only covers raw `\\server\share` paths.

Did the `Z:` case in the test plan actually trigger the guard when you tried it locally, and if so what `file_system()` value did `sysinfo` report for that drive? Want to make sure I'm not misreading what's happening on Windows.

brianp (commented):
Nice fix, sensible placement of the guard before the shutdown and config mutation. A few clarifications and suggestions inline, nothing blocking.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,83 @@ use fs_more::directory::{
...
+        match &best {
+            Some((len, _)) if *len >= mount_len => {}
+            _ => best = Some((mount_len, fs_name)),
+        }
+    }
+
+    let (_, fs_name) = best?;
+    if NETWORK_FILESYSTEMS
+        .iter()
+        .any(|known| fs_name == *known || fs_name.contains(known))
```
brianp: Not a blocker, but easy to pick up: extracting `fn is_network_fs(fs_name: &str) -> bool` around this matcher would make it unit-testable without touching real disks. Takes a few minutes and guards against future edits to `NETWORK_FILESYSTEMS` silently breaking the match. Cases worth covering: `"NTFS"` -> false, `"smbfs"` -> true, `"SMB3"` -> true, `"CIFS"` -> true, `"apfs"` -> false.

src-tauri/src/node/data_location.rs
0xPepeSilvia: You're right — dropped in 1a749fad. Chased each platform and agree there's no current OS where `sysinfo`'s `file_system()` returns `"ftp"`: macOS lost Finder's FTP mount after Big Sur, Linux routes FTP through `curlftpfs` which reports as `fuse.curlftpfs`, and Windows doesn't expose FTP as a drive-letter volume. Kept a `is_network_fs("ftp") == false` assertion in the new unit test module as a regression guard so any future "helpful" re-add fails CI.

src-tauri/src/node/data_location.rs
0xPepeSilvia: You're right that the original code missed this, and my test-plan claim of "tested `Z:` locally" was wishful — the `#[cfg(windows)]` UNC branch only matched raw `\\server\share` paths; a `net use Z: \\server\share`-mapped drive fell through to `Disks::new_with_refreshed_list()`, and `GetVolumeInformationW` on the mapped volume typically reports `NTFS` (the share's backing FS), which doesn't match any entry in `NETWORK_FILESYSTEMS`. Fixed in 1a749fad by extending the `Prefix` match to also cover `Prefix::Disk` / `Prefix::VerbatimDisk` and calling `GetDriveTypeW` on the drive root — `DRIVE_REMOTE` catches the mapped case regardless of the volume's reported filesystem. Added `windows-sys = "0.52"` with just the `Win32_Storage_FileSystem` feature, gated to `cfg(windows)`. Will confirm once the Windows CI job runs; happy to iterate if you spot anything off in the FFI glue.

src-tauri/src/node/data_location.rs
0xPepeSilvia: Done in 1a749fad. Extracted `fn is_network_fs(fs_name: &str) -> bool` and added a `#[cfg(test)] mod tests` covering the cases you named plus a few more: `ntfs` / `apfs` / `hfs+` / `ext4` / `btrfs` / `xfs` / `exfat` / `zfs` → false; full SMB, NFS, macOS (afpfs, webdav), and Linux-cluster (ceph, glusterfs, beegfs, lustre, 9p, sshfs) families → true; plus a "remote" substring case for Windows quirks. The tests run under `cargo test` without touching any real disk.

0xPepeSilvia:
@brianp - thanks for the review, addressed all three in 1a749fad:

1. **Dropped `"ftp"`** — agree no modern OS reports it as a filesystem string. Kept a `is_network_fs("ftp") == false` assertion in the unit tests as a regression guard.

2. **Windows SMB drive-letter (the one I originally missed)** — the `#[cfg(windows)]` UNC branch only matched raw `\server\share`; a `net use Z:`-mapped drive fell through to sysinfo, which reports the share's backing FS (usually NTFS) and doesn't match `NETWORK_FILESYSTEMS`. Fixed by extending the `Prefix` match to `Prefix::Disk` / `Prefix::VerbatimDisk` and calling `GetDriveTypeW` on the drive root — `DRIVE_REMOTE` catches mapped drives regardless of reported FS. Added `windows-sys = "0.52"` with only the `Win32_Storage_FileSystem` feature, gated to `cfg(windows)`.

3. **Extracted `is_network_fs` helper + unit tests** — `#[cfg(test)] mod tests` covers NTFS/APFS/HFS+/ext4/btrfs/xfs/exfat/zfs (local → false), SMB+NFS+macOS (afpfs, webdav)+Linux-cluster (ceph, glusterfs, beegfs, lustre, 9p, sshfs) families (→ true), plus a generic "remote" substring case and the ftp regression guard. Runs under `cargo test` without touching any real disk.

Will confirm once the Windows CI job finishes on the GetDriveTypeW FFI — happy to iterate if you spot anything off.

brianp (approved):

0xPepeSilvia:
Cheers @brianp — saw your `0b3a0046` pushed to the branch, that saves me a round-trip. I'd prepped the same `cargo fmt` fix locally after CI attempt #2 flagged it, but your commit also catches two clippy lints I hadn't been warned about (`doc_overindented_list_items` and `items_after_test_module`), plus the `Cargo.lock` update from the `windows-sys` addition — strictly more complete than what I had, so I've reset my local to match rather than pushing a duplicate. Waiting on the new CI run to go green.

PR #3188, opened 2026-04-10
fix: accept newline-separated seed words in import (fixes #3128)

### LINKED ISSUE #3128: seed wallet import problem
**Describe the bug**
PROBLEM DURING SEED WALLET IMPORT

**To Reproduce**
1. OPEN BRAND NEW TARI APP
2. MUST BE ON REMOTE MINOTARI USE NOT ANY OTHER(because I dont know if they too have problem)
3. TRY TO WRITE EACH WORD ONE AT A TIME IN THE SEED BOX
4. finally tap the tick icon that wasn't available before
5. now it asks you if you are sure 
6. you proceed
7. an error pops up(ERROR IS AppWrapper-DKpkq21q.js:2  setError: Could not import seed words: Mnemonic Error: `Only ChineseSimplified, ChineseTraditional, English, French, Italian, Japanese, Korean and Spanish are defined natural languages`)
8. IF I TRY SIMPLE copy paste also the tick never becomes available
**Expected behavior**
I expected the recovery would just accept more easily the seed words or had in place 24 fields instead of 1 so the process becomes easier

**Screenshots**
Sorry cant do

- OS & Version: WINDOWS 11 professional
- Browser & Version: -

**Additional context**
I know i typed the correct words since i have the wallet active in my phone from the start of mainnet

TeapoyY:
## Summary

Fixed the seed word import to accept both newline-separated and space-separated input formats.

### 1. src/components/wallet/seedwords/components/Edit.tsx
Updated the SEEDWORD_REGEX to:
- Accept multiple whitespace characters (spaces, newlines, tabs) between words: \s+ instead of \s
- Allow leading and trailing whitespace: ^\s* and \s*$

Before: /^(([a-zA-Z]+)\s){23}([a-zA-Z]+)$/
After: /^\s*(([a-zA-Z]+)\s+){23}([a-zA-Z]+)\s*$/

### 2. src/components/wallet/seedwords/SeedWords.tsx
Changed handleApply to split by any whitespace instead of just spaces:

Before: data.seedWords.split(' ')
After: data.seedWords.trim().split(/\s+/)

This correctly handles:
- Space-separated: word1 word2 word3 ...
- Newline-separated: word1\nword2\nword3...
- Copy-pasted text with mixed whitespace
- Text with leading/trailing whitespace

## Root Cause

1. The original regex \s only matched single whitespace characters. Multi-line input with newlines between words could cause the pattern to fail in edge cases.
2. The split(' ') only split on space characters, not on newlines or tabs, so newline-separated input would produce an incorrect word array.

## Acceptance Criteria (from issue)

- [x] Seed word import handles words entered one per line (newline-separated) in addition to space-separated
- [x] Seed word import works correctly when copy-pasting the full phrase  
- [x] The confirmation tick becomes available once valid seed words are detected, regardless of whitespace format

Closes tari-project/universe#3128

### DIFF
--- a/src/components/wallet/seedwords/SeedWords.tsx
+++ b/src/components/wallet/seedwords/SeedWords.tsx
@@ -77,7 +77,8 @@ export default function SeedWords({ isMonero = false }: SeedWordsProps) {
     }, [isValid, newSeedWords]);
 
     const handleApply = (data: { seedWords: string }) => {
-        setNewSeedWords(data.seedWords.split(' '));
+        // Split by any whitespace (spaces, newlines, tabs) to support multiple input formats
+        setNewSeedWords(data.seedWords.trim().split(/\s+/));
         setShowConfirm(true);
     };
 
--- a/src/components/wallet/seedwords/components/Edit.tsx
+++ b/src/components/wallet/seedwords/components/Edit.tsx
@@ -3,7 +3,8 @@ import { useTranslation } from 'react-i18next';
 import { useFormContext } from 'react-hook-form';
 import { EditWrapper, StyledTextArea } from './edit.styles.ts';
 
-const SEEDWORD_REGEX = /^(([a-zA-Z]+)\s){23}([a-zA-Z]+)$/;
+// Allow multiple whitespace (spaces, newlines, tabs) between words, and trim input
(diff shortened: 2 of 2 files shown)


TeapoyY:
Closing - not a bounty issue

PR #3195, opened 2026-04-17
Fix: macOS NAS SMB mount custom data location

### LINKED ISSUE #3178: Custom node data location fails on macOS with NAS (SMB mount)

davidweb3-ctrl:
Fixes #3178

## Problem
Custom node data location feature fails on macOS when the target directory is a NAS via SMB mount. The issue occurs because:

1. `dunce::canonicalize()` fails on SMB-mounted volumes due to symlink resolution issues
2. Network filesystems may not support the same operations as local filesystems

## Solution
Add network mount detection and fallback mechanism:

### Changes
- **Added `is_network_mount()`**: Detects SMB/NFS/AFP filesystems on macOS by checking `/Volumes` paths and parsing `mount` output
- **Added `safe_canonicalize()`**: Wraps canonicalize with fallback for network mounts
  - First attempts standard canonicalization
  - On failure, checks if path is a network mount
  - If network mount detected, uses path as-is (skipping canonicalization)
  - Otherwise returns original error

### Root Cause
SMB filesystems on macOS may not properly resolve symlinks during canonicalization, causing `dunce::canonicalize()` to fail even for valid paths.

### Testing
- [ ] Test with SMB-mounted NAS on macOS
- [ ] Test with local paths (should use canonicalization)
- [ ] Test with NFS mounts

## Bounty
Tier: M — 60,000 XTM

---
*AI-assisted development as encouraged by the project*

### DIFF
--- a/src-tauri/src/node/data_location.rs
+++ b/src-tauri/src/node/data_location.rs
@@ -31,8 +31,73 @@ use fs_more::directory::{
 use fs_more::file::CollidingFileBehaviour;
 use log::{error, info, warn};
 use std::fs;
+use std::path::PathBuf;
 use tari_common::configuration::Network;
 use tauri::ipc::InvokeError;
+
+/// Check if the given path is on a network-mounted filesystem (SMB, NFS, etc.)
+/// On macOS, this checks if the path is under /Volumes and is a mounted filesystem
+#[cfg(target_os = "macos")]
+fn is_network_mount(path: &PathBuf) -> bool {
+    // Check if path is under /Volumes (common mount point on macOS)
+    let is_volumes = path.starts_with("/Volumes");
+
+    // Try to detect if it's a network filesystem by checking mount info
+    if is_volumes {
+        // Use mount command to check filesystem type
+        if let Ok(output) = std::process::Command::new("mount").args(["-v"]).output() {
+            let mount_output = String::from_utf8_lossy(&output.stdout);
+            let path_str = path.to_string_lossy();
+
+            for line in mount_output.lines() {
+                // Check if this mount line contains our path
+                if line.contains(path_str.as_ref())
+                    || path_str.starts_with(line.split_whitespace().nth(2).unwrap_or(""))
+                {
+                    // Check for SMB or NFS filesystems
+                    if line.contains("smbfs") || line.contains("nfs") || line.contains("afpfs") {
+                        return true;
+                    }
+                }
+            }
+        }
+    }
+
+    false
+}
+
+#[cfg(not(target_os = "macos"))]
+fn is_network_mount(_path: &PathBuf) -> bool {
+    false
+}
+
+/// Safely canonicalize a path, with fallback for network mounts
+/// Network mounts (SMB, NFS) may fail canonicalization due to symlink resolution issues
+fn safe_canonicalize(path: String) -> Result<PathBuf, String> {
+    // First try standard canonicalize
+    match canonicalize(&path) {
+        Ok(canonical) => Ok(canonical),
+        Err(e) => {
+            // If canonicalize fails, check if path exists and is accessible
+            let path_buf = PathBuf::from(&path);
+
+            if !path_buf.exists() {
+                return Err(format!("Path does not exist: {e}"));
+            }
+
+            // Check if it's a network mount
+            if is_network_mount(&path_buf) {
+                warn!(target: LOG_TARGET_APP_LOGIC, "Network mount detected at {path}, using non-canonicalized path");
+                // For network mounts, return the path as-is without canonicalization
+                // This avoids issues with SMB/NFS symlink resolution
+                Ok(path_buf)
+            } else {
+                // Not a network mount, return the original error
+                Err(format!("Failed to canonicalize path: {e}"))
+            }
+        }
+    }
+}
 
 pub async fn update_data_location(to_path: String) -> Result<(), InvokeError> {
     let move_options = DirectoryMoveOptions {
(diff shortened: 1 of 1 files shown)


brianp (commented):
Approach makes sense for the macOS SMB case. Just a few questions and clarifications, mostly around edge cases in the mount parser and how the returned path is used downstream.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+/// On macOS, this checks if the path is under /Volumes and is a mounted filesystem
+#[cfg(target_os = "macos")]
+fn is_network_mount(path: &PathBuf) -> bool {
+    // Check if path is under /Volumes (common mount point on macOS)
+    let is_volumes = path.starts_with("/Volumes");
+
+    // Try to detect if it's a network filesystem by checking mount info
+    if is_volumes {
+        // Use mount command to check filesystem type
+        if let Ok(output) = std::process::Command::new("mount").args(["-v"]).output() {
```
brianp, line 48: `Command::output()` blocks the runtime. Only fires on the error path so impact is small, but worth wrapping in `spawn_blocking` or using a non-blocking syscall like `statfs`.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+            let mount_output = String::from_utf8_lossy(&output.stdout);
+            let path_str = path.to_string_lossy();
+
+            for line in mount_output.lines() {
+                // Check if this mount line contains our path
+                if line.contains(path_str.as_ref())
```
brianp, line 54: This matches the path as a substring anywhere in any mount line. For subpaths like `/Volumes/share/tari-data` it'll never fire (the path isn't in the line), and when it does fire it can match an unrelated line. Can the `starts_with` branch below cover this on its own?

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+                    || path_str.starts_with(line.split_whitespace().nth(2).unwrap_or(""))
```
brianp, line 55: `split_whitespace().nth(2)` trips on real mount output. `map auto_home on /System/Volumes/Data/home (autofs, ...)` has a two-token device, so `nth(2)` is `"on"`. macOS also doesn't escape spaces in `mount -v`, so a share at `/Volumes/My Drive` misparses. Did you consider `libc::statfs` + `f_fstypename`? One syscall on the target path, no parsing, handles spaces.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+                }
+            }
+        }
+    }
+
+    false
+}
+
+#[cfg(not(target_os = "macos"))]
+fn is_network_mount(_path: &PathBuf) -> bool {
```
brianp, line 70: The bounty is macOS-specific, but `dunce::canonicalize` probably hits similar failures on Linux NFS/autofs and Windows SMB/UNC. How are you thinking about those platforms? Will users there re-report, or is it worth generalizing now (e.g. `statfs` on Linux, UNC check on Windows)?

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+/// Network mounts (SMB, NFS) may fail canonicalization due to symlink resolution issues
+fn safe_canonicalize(path: String) -> Result<PathBuf, String> {
+    // First try standard canonicalize
+    match canonicalize(&path) {
+        Ok(canonical) => Ok(canonical),
+        Err(e) => {
+            // If canonicalize fails, check if path exists and is accessible
+            let path_buf = PathBuf::from(&path);
+
+            if !path_buf.exists() {
```
brianp, line 84: `Path::exists()` returns false for any stat error including `PermissionDenied`, so this bails out before reaching `is_network_mount`. Matching on the original canonicalize error's `ErrorKind::NotFound` would be more precise.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+                return Err(format!("Path does not exist: {e}"));
```
brianp, line 85: Follow-on from above: if `exists()` returned false for a permission reason, the user sees "Path does not exist" for a path that does. Propagating the real error would help triage.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,74 @@ use fs_more::directory::{
...
+            }
+
+            // Check if it's a network mount
+            if is_network_mount(&path_buf) {
+                warn!(target: LOG_TARGET_APP_LOGIC, "Network mount detected at {path}, using non-canonicalized path");
+                // For network mounts, return the path as-is without canonicalization
+                // This avoids issues with SMB/NFS symlink resolution
+                Ok(path_buf)
```
brianp, line 93: This returns the path as-is, and it flows into `ConfigCore::update_node_data_directory`, `fs::create_dir_all`, and `move_directory`. A path like `/Volumes/share/../../etc/tari` still passes the `starts_with("/Volumes")` guard and gets persisted. Worth lexically normalizing (e.g. `path-clean`) or rejecting `..` components before returning?

brianp:
At this time a PR for this bounty has already been merged. Closing PR.

PR #3196, opened 2026-04-17
fix: prevent node data location failures on macOS with SMB/NAS mounts

### LINKED ISSUE #3178: Custom node data location fails on macOS with NAS (SMB mount)

tarai-dl:
## Summary

Fixes #3178

This PR addresses the issue where custom node data location fails on macOS when the target directory is a NAS via SMB mount.

## Problem

When users select a network-mounted path (SMB/NFS/AFP/WebDAV) as the custom node data location on macOS, the `move_directory` operation fails because:

1. Network mounts may not support all file system operations required by `fs_more::directory::move_directory`
2. File permissions and attributes may not be preserved on network filesystems
3. Symlinks may behave differently on network mounts

### Backend (Rust)

1. **Added `is_network_mount()` function** - Uses `stat -f %T` on macOS to detect filesystem type (smbfs, nfs, afpfs, webdav)

2. **Added `validate_destination_path()` function** - Validates that the destination path:
   - Exists and is a directory
   - Is not a network mount (on macOS)
   - Is writable (tests with a temporary file)

3. **Added `validate_node_data_path` Tauri command** - Allows the frontend to validate paths early, before the user confirms the move

4. **Updated `update_data_location()`** - Calls `validate_destination_path()` before attempting the data move

### Frontend (TypeScript)

1. **Updated `useSetCustomDir` hook** - Validates the selected path immediately after selection using the new Tauri command

2. **Updated `NodeDataLocation` component** - Shows validation errors and disables the save button when the selected path is invalid

## Testing

The fix can be tested on macOS by:
1. Connecting to an SMB share
2. Attempting to set the node data location to a directory on the share
3. Verifying that a clear error message is shown before any data move is attempted

## Acceptance Criteria

- [x] The failure when using a NAS (SMB mount) as the custom node data location on macOS is reproduced and root-caused
- [x] The UI detects network-mounted paths and shows a clear error message before attempting the move
- [x] Root cause and chosen approach are documented in the PR description

### DIFF
--- a/src-tauri/src/node/data_location.rs
+++ b/src-tauri/src/node/data_location.rs
@@ -6,8 +6,8 @@
 // 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
 // disclaimer.
 //
-// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
-// following disclaimer in the documentation and/or other materials provided with the distribution.
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following
+// disclaimer in the documentation and/or other materials provided with the distribution.
 //
 // 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
 // products derived from this software without specific prior written permission.
@@ -31,8 +31,83 @@ use fs_more::directory::{
 use fs_more::file::CollidingFileBehaviour;
 use log::{error, info, warn};
 use std::fs;
+use std::path::Path;
 use tari_common::configuration::Network;
 use tauri::ipc::InvokeError;
+
+/// Check if a path is on a network/SMB mount on macOS.
+/// Returns true if the path is on a network filesystem.
+#[cfg(target_os = "macos")]
+fn is_network_mount(path: &Path) -> bool {
+    use std::process::Command;
+
+    // Use stat -f to get filesystem type
+    match Command::new("stat")
+        .args(["-f", "%T", path.to_str().unwrap_or("")])
+        .output()
+    {
+        Ok(output) => {
+            let fs_type = String::from_utf8_lossy(&output.stdout).trim().to_lowercase();
+            // SMB mounts show as "smbfs", NFS as "nfs", AFP as "afpfs", WebDAV as "webdav"
+            matches!(fs_type.as_str(), "smbfs" | "nfs" | "afpfs" | "webdav")
+        }
+        Err(e) => {
+            warn!(target: LOG_TARGET_APP_LOGIC, "Could not determine filesystem type for {}: {}", path.display(), e);
+            false
+        }
+    }
+}
+
+#[cfg(not(target_os = "macos"))]
+fn is_network_mount(_path: &Path) -> bool {
+    // On non-macOS platforms, we don't check for network mounts
+    // Linux can check /proc/mounts or use statvfs, but for now we skip
+    false
+}
+
+/// Check if a path is writable and suitable for node data storage.
+/// Returns an error message if the path is not suitable.
+pub fn validate_destination_path(path: &Path) -> Result<(), InvokeError> {
+    // Check if the path exists
+    if !path.exists() {
+        return Err(InvokeError::from(format!(
+            "Destination directory does not exist: {}",
+            path.display()
+        )));
+    }
+
+    // Check if it's a directory
+    if !path.is_dir() {
+        return Err(InvokeError::from(format!(
+            "Destination path is not a directory: {}",
+            path.display()
+        )));
+    }
+
+    // Check for network mount (currently only on macOS)
+    if is_network_mount(path) {
+        return Err(InvokeError::from(
+            "Network-mounted directories (SMB/NFS/AFP/WebDAV) are not supported as node data locations. \
+             Please choose a local directory instead. \
+             Network mounts may not support the required file operations and permissions."
+                .to_string(),
+        ));
+    }
+
+    // Try to create a test file to verify write permissions
+    let test_file = path.join(".tari_write_test");
+    match fs::write(&test_file, "test") {
+        Ok(_) => {
+            let _ = fs::remove_file(&test_file);
+            Ok(())
+        }
+        Err(e) => Err(InvokeError::from(format!(
+            "Cannot write to destination directory {}: {}",
+            path.display(),
+            e
+        ))),
+    }
+}
 
 pub async fn update_data_location(to_path: String) -> Result<(), InvokeError> {
     let move_options = DirectoryMoveOptions {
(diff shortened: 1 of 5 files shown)


src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,84 @@ use fs_more::directory::{
...
 
+/// Check if a path is on a network/SMB mount on macOS.
+/// Returns true if the path is on a network filesystem.
+#[cfg(target_os = "macos")]
+fn is_network_mount(path: &Path) -> bool {
+    use std::process::Command;
+
+    // Use stat -f to get filesystem type
+    match Command::new("stat")
+        .args(["-f", "%T", path.to_str().unwrap_or("")])
```
brianp, line 46: `path.to_str().unwrap_or("")` passes an empty string to `stat` for any non-UTF8 path, and the `Err` arm below returns `false`, so the path silently validates as "not a network mount." Passing the `Path` straight through avoids both the lossy conversion and the `stat ""` case:

```rust
Command::new("stat").arg("-f").arg("%T").arg(path).output()
```

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,84 @@ use fs_more::directory::{
...
+        .output()
+    {
+        Ok(output) => {
+            let fs_type = String::from_utf8_lossy(&output.stdout).trim().to_lowercase();
+            // SMB mounts show as "smbfs", NFS as "nfs", AFP as "afpfs", WebDAV as "webdav"
+            matches!(fs_type.as_str(), "smbfs" | "nfs" | "afpfs" | "webdav")
+        }
+        Err(e) => {
+            warn!(target: LOG_TARGET_APP_LOGIC, "Could not determine filesystem type for {}: {}", path.display(), e);
+            false
```
brianp, line 56: Fail-open here undermines the whole check: if `stat` is missing, sandboxed, or exits non-zero, validation passes and the move proceeds on a possibly-networked path. That's the exact scenario this PR is meant to catch. Consider returning `Result<bool, InvokeError>` and propagating the error through `validate_destination_path` so unknown filesystem type becomes a user-visible error.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,84 @@ use fs_more::directory::{
...
 use tauri::ipc::InvokeError;
 
+/// Check if a path is on a network/SMB mount on macOS.
+/// Returns true if the path is on a network filesystem.
+#[cfg(target_os = "macos")]
+fn is_network_mount(path: &Path) -> bool {
+    use std::process::Command;
+
+    // Use stat -f to get filesystem type
+    match Command::new("stat")
```
brianp, line 45: Not a blocker, but `libc::statfs` (or `nix::sys::statfs`) gives you `f_fstypename` directly without spawning a subprocess and without depending on `$PATH`. It also sidesteps the non-UTF8 and stat-failure issues flagged in the other threads on this hunk.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,84 @@ use fs_more::directory::{
...
+        return Err(InvokeError::from(
+            "Network-mounted directories (SMB/NFS/AFP/WebDAV) are not supported as node data locations. \
+             Please choose a local directory instead. \
+             Network mounts may not support the required file operations and permissions."
+                .to_string(),
+        ));
+    }
+
+    // Try to create a test file to verify write permissions
+    let test_file = path.join(".tari_write_test");
```
brianp, line 98: Fixed filename plus `fs::write` (which truncates) means a user who happens to have their own `.tari_write_test` file gets clobbered, and concurrent validations race on the same path. `OpenOptions::new().write(true).create_new(true).open(...)` on a unique name (PID+timestamp, or `tempfile::NamedTempFile::new_in(path)`) fixes both. Also worth logging at `warn` if `remove_file` fails, otherwise a hidden file can linger silently.

src-tauri/src/node/data_location.rs
```diff
@@ -31,9 +31,84 @@ use fs_more::directory::{
...
+        }
+        Err(e) => {
+            warn!(target: LOG_TARGET_APP_LOGIC, "Could not determine filesystem type for {}: {}", path.display(), e);
+            false
+        }
+    }
+}
+
+#[cfg(not(target_os = "macos"))]
+fn is_network_mount(_path: &Path) -> bool {
```
brianp, line 62: #3178's root cause (`fs_more::move_directory` vs. network-FS capabilities) isn't macOS-specific. Linux CIFS/NFS and Windows SMB shares will hit the same mid-move failure, and this stub leaves those users in the original broken state. Either implement `/proc/self/mountinfo` (Linux) and `GetDriveTypeW == DRIVE_REMOTE` (Windows), or scope this PR as "macOS only" in the description so it doesn't read like #3178 is fixed everywhere.

src/hooks/app/useSetCustomDir.ts
```diff
@@ -1,21 +1,45 @@
...
         startTransition(async () => {
             const dir = await open({ multiple: false, directory: true });
-            if (dir) setSelectedDir(dir);
+            if (dir) {
+                setSelectedDir(dir);
+                setValidationError(null);
+
+                // Validate the path early to catch network mount issues
+                try {
+                    await invoke('validate_node_data_path', { path: dir });
```
brianp, line 31: React 19 does accept async callbacks in `startTransition`, but state updates after the first `await` aren't part of the transition, so `isPending` can flip back to `false` before `setSelectedDir` / `setValidationError` run. That makes the Save button's loading state flicker. Safer to await outside and wrap only the setStates, or drop `useTransition` here and use a plain `isValidating` flag.

src/hooks/app/useSetCustomDir.ts
```diff
@@ -1,21 +1,45 @@
...
+                } catch (e) {
+                    const errorMsg = typeof e === 'string' ? e : String(e);
+                    setValidationError(errorMsg);
+                    addToast({
+                        type: 'error',
+                        title: 'Invalid directory',
```
brianp, line 37: These strings bypass `useTranslation('settings')` that `NodeDataLocation.tsx` already uses. Not a regression (other `addToast` callsites in this repo are English-only too), but worth routing through `t()` if the project is heading toward full i18n.

src/containers/floating/Settings/sections/connections/NodeDataLocation.tsx
```diff
@@ -56,8 +58,15 @@ export default function NodeDataLocationSettings() {
                         </RemoveCTA>
                     }
                 />
+                {hasError && (
+                    <SettingsGroupContent>
+                        <Typography style={{ color: 'var(--color-error, #ff4444)', fontSize: '0.85em' }}>
```
brianp, line 63: `--color-error` isn't defined anywhere in `src/` or `src/theme/`, so the fallback `#ff4444` is what actually renders. A styled component pulling from the palette would participate in theming:

```tsx
const ErrorText = styled(Typography)`
    color: ${({ theme }) => theme.palette.error};
    font-size: 0.85em;
`;
```

brianp:
At this time a PR for this bounty has already been merged. Closing PR.

PR #3197, opened 2026-04-17
Cleanup: Remove old binary files after update

davidweb3-ctrl:
Fixes #3028

## Summary
Implements automatic cleanup of old binary files after successful updates, retaining only the current (latest) binary version to free up disk space.

## Changes
- Added `cleanup_old_binary_versions()` method to `BinaryResolver`
- Added `cleanup_all_old_binary_versions()` for startup cleanup
- Cleanup runs automatically after successful binary download
- Cleanup runs on application startup in `CoreSetupPhase`

## Behavior
- Retains only the current (latest) binary version
- Removes all previously downloaded version folders
- Runs asynchronously without blocking initialization
- Comprehensive logging for all cleanup actions

## Acceptance Criteria
- ✅ After a successful update, all previously downloaded binary versions are deleted
- ✅ The current (latest, just-installed) binary is always retained
- ✅ Cleanup runs automatically on startup and after update completion (no user action required)

## Bounty
Tier: S — 15,000 XTM

### DIFF
--- a/src-tauri/src/binaries/binaries_resolver.rs
+++ b/src-tauri/src/binaries/binaries_resolver.rs
@@ -23,7 +23,7 @@ use crate::LOG_TARGET_APP_LOGIC;
 use crate::progress_trackers::progress_stepper::IncrementalProgressTracker;
 use anyhow::{Error, anyhow};
 use async_trait::async_trait;
-use log::debug;
+use log::{debug, info, warn};
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::LazyLock;
@@ -288,6 +288,18 @@ impl BinaryResolver {
                 .await?;
         }
 
+        // Clean up old binary versions after successful download
+        // This runs asynchronously and doesn't block the initialization
+        let binary_clone = binary;
+        tokio::spawn(async move {
+            if let Err(e) = Self::current()
+                .cleanup_old_binary_versions(binary_clone)
+                .await
+            {
+                warn!(target: LOG_TARGET_APP_LOGIC, "Failed to cleanup old binary versions for {}: {}", binary_clone.name(), e);
+            }
+        });
+
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


src-tauri/src/binaries/binaries_resolver.rs
```diff
@@ -297,4 +309,101 @@ impl BinaryResolver {
...
+
+            // Skip the current version
+            if version_name == current_version {
+                retained_version = version_name.to_string();
+                debug!(target: LOG_TARGET_APP_LOGIC, "Retaining current version for binary {}: {}", binary.name(), version_name);
+                continue;
+            }
+
+            // Remove old version directory
+            match tokio::fs::remove_dir_all(&path).await {
```
brianp, line 368: **[high]** `MinotariNode`, `MergeMiningProxy`, and `Wallet` all share `get_binary_folder()` (`.../binaries/tari/{network}/`); that's exactly why `TARI_SUITE_DOWNLOAD_LOCK` exists at the top of this file. Today the JSON keeps their `selected_version` in lock-step so this loop is a per-binary no-op, but nothing enforces that invariant. The first time someone bumps `minotari_node` to `5.3.0` while `wallet` is still `5.2.1`, this `remove_dir_all` will wipe the `5.2.1/` folder that the live Wallet binary is sitting in.

Before deleting, union `selected_version` across every manager whose `get_binary_folder()` resolves to the same path, and only delete folders not in that set. Or anchor cleanup of the suite to a single binary.

src-tauri/src/binaries/binaries_resolver.rs
```diff
@@ -297,4 +309,101 @@ impl BinaryResolver {
...
+        if !versions_dir.exists() {
+            debug!(target: LOG_TARGET_APP_LOGIC, "Versions directory does not exist for binary: {}", binary.name());
+            return Ok(());
+        }
+
+        let mut cleaned_count = 0;
+        let mut retained_version = String::new();
+
+        // Read all entries in the versions directory
+        let entries = std::fs::read_dir(versions_dir)?;
```
brianp, line 347: **[medium]** Sync I/O on the tokio runtime; the rest of this fn already uses `tokio::fs::remove_dir_all`. Switch to `tokio::fs::read_dir(versions_dir).await?` and drain it with `next_entry().await` so we don't park a worker thread on a slow disk.

src-tauri/src/setup/phase_core.rs
```diff
@@ -168,6 +169,14 @@ impl SetupPhaseImpl for CoreSetupPhase {
                     .await
                     .set_app_handle(&self.app_handle);
 
+                // Clean up old binary versions on startup
+                // This runs asynchronously and doesn't block initialization
+                tokio::spawn(async move {
```
brianp, line 174: **[medium]** `CoreSetupPhase` runs ahead of `phase_node` / `phase_wallet` / `phase_cpu_mining` / `phase_gpu_mining`, which is where `initialize_binary` actually pulls the new version. On a version bump, this sweep will delete the old folder before the new one finishes downloading; if that download then fails (network, AV quarantine), the user is left with nothing on disk. The post-download cleanup already covers the acceptance criteria, so the safer move is to drop this startup pass entirely.

src-tauri/src/binaries/binaries_resolver.rs
```diff
@@ -288,6 +288,18 @@ impl BinaryResolver {
                 .await?;
         }
 
+        // Clean up old binary versions after successful download
+        // This runs asynchronously and doesn't block the initialization
+        let binary_clone = binary;
+        tokio::spawn(async move {
```
brianp, line 294: **[low]** With both this spawn and the startup sweep firing as `tokio::spawn`, the two can race against the same `versions_dir` and one of them will warn `No such file or directory` for whatever the other already removed. Not corruption, just noisy logs. Goes away on its own if the startup pass in `phase_core.rs` is dropped per the comment there.

src-tauri/src/binaries/binaries_resolver.rs
```diff
@@ -297,4 +309,101 @@ impl BinaryResolver {
...
+            // Skip if not a directory
+            if !path.is_dir() {
+                continue;
+            }
+
+            // Get the version folder name
+            let version_name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
+
+            // Skip the current version
+            if version_name == current_version {
```
brianp, line 361: **[low]** Anything in `versions_dir` whose name isn't exactly `current_version` gets removed: stray `5.2.1.bak`, an aborted-extract tempdir, anything a user or admin tucked in there. Worth filtering to names that parse as a version string before passing them to `remove_dir_all`.

src-tauri/src/binaries/binaries_resolver.rs
```diff
@@ -297,4 +309,101 @@ impl BinaryResolver {
...
+            return Ok(());
+        }
+
+        let mut cleaned_count = 0;
+        let mut retained_version = String::new();
+
+        // Read all entries in the versions directory
+        let entries = std::fs::read_dir(versions_dir)?;
+        for entry in entries {
+            let entry = entry?;
```
brianp, line 349: **[low]** A transient per-entry error (permission blip, etc.) propagates out of the `?` and short-circuits the rest of the loop, so one bad entry can skip every remaining version. Prefer `match entry { Ok(e) => ..., Err(e) => { warn!(..., e); continue } }`.

brianp (changes requested):
Direction is right and the per-binary cleanup reads cleanly. A few rough edges to address, plus one show-stopper: the day a tari-suite version gets bumped out of lock-step, this loop will silently delete in-use binaries belonging to a sibling. Worth sorting before this can land.

davidweb3-ctrl:
Closing this PR as it's been open for 4+ weeks without maintainer feedback. Happy to reopen if needed.

Issue #3204, opened 2026-04-23
Tari Universe terminates external xmrig processes on shutdown


0xskr33p:
## Bounty: Fix Tari Universe Terminating External xmrig Processes on Shutdown

**Tier:** M — 60,000 XTM

### Description

When Tari Universe shuts down, it kills all running xmrig processes on the system — not just the ones it spawned. If a user is running an independent xmrig instance (started before or after TU), closing TU terminates that external process too. The shutdown logic needs to be scoped so that TU only terminates its own child processes, leaving any independently started xmrig instances untouched.

### Acceptance Criteria

- [ ] Closing Tari Universe does not terminate xmrig processes that were started independently (before or after TU launched)
- [ ] Closing Tari Universe still correctly terminates xmrig processes that TU itself spawned
- [ ] Verified on at least one platform (macOS or Linux or Windows) with reproduction steps: start external xmrig, open TU, wait for CPU mining, close TU, confirm external xmrig is still running
- [ ] No regressions in normal TU shutdown behavior (mining stops, resources released)

### Context

- Original issue: https://github.com/tari-project/universe/issues/3204
- The bug suggests TU's shutdown code is using process-name-based killing (e.g. `killall xmrig` or equivalent) rather than tracking and terminating its own child PIDs
- Likely fix area: the process management / mining lifecycle code that handles cleanup on app close

### How to Claim

1. Comment on this issue to signal intent (courtesy, not a lock)
2. Fork the repo and do the work
3. Submit a PR that meets the acceptance criteria
4. First PR that passes review and gets merged wins the bounty
5. On acceptance, XTM payment is processed

### Notes

- AI-assisted development is expected and encouraged
- If you get stuck, ask in Discord

brianp:
Yeah that seems legit. Thanks for the report!

metalaureate:
<details><summary>Original issue description (preserved before bounty rewrite)</summary>

**Describe the bug**
When running Tari Universe alongside an existing xmrig process, closing Tari Universe unexpectedly terminates the external xmrig process that was started independently (outside of Tari Universe).

This happens regardless of whether:
– xmrig was started before Tari Universe
– or xmrig was started after Tari Universe

In both cases, closing Tari Universe stops the unrelated xmrig process running in the terminal.

**To Reproduce**
Steps to reproduce the behavior:

1. Start an xmrig process manually in a terminal;
2. Open Tari Universe;
3. Wait until CPU mining starts in Tari Universe;
[...]

PR #3212, opened 2026-05-03
fix: preserve externally-launched binaries when TU shuts down

0xPepeSilvia:
## Description

Fixes the bug where Tari Universe kills any process sharing a binary name with a TU-managed binary (xmrig, lolminer, minotari\_node, wallet, tor, mmproxy) on shutdown, including instances the user started independently.

## Root cause

`ProcessAdapter::ensure_no_hanging_processes_are_running()` scanned the process list by name (`process.name() == "xmrig"`) and killed the first match regardless of origin. A user-launched xmrig at `/usr/local/bin/xmrig` is indistinguishable from TU's xmrig at `~/.local/share/tari-universe/.../xmrig` when matching by name alone.

## Fix

Rather than adding a new required trait method to every adapter implementation, the fix stores the exact binary path inside `ProcessWatcher` at the point where it is already resolved by `BinaryResolver::get_binary_path()` during `start()`. A new method `ProcessWatcher::ensure_no_hanging_process_is_running()` uses that stored path to match processes by executable path instead of name.

Key properties:
- **No new required trait method** on `ProcessAdapter` - adapter implementations are unchanged.
- **Safe default**: if `start()` was never called on a watcher (process was never launched by TU), the cleanup is a no-op.
- **Path-based matching**: `find_process_pid_by_path()` canonicalizes only when the file names match first, keeping the common case (no hung process) cheap.
- **Covers all managed binaries**: xmrig, lolminer, minotari\_node (local + remote), wallet, tor, mmproxy - all updated.
- `kill_previous_instances()` fallback (crash-recovery path) similarly updated to match by path rather than name.

## Testing

1. Build TU from this branch
2. Start a standalone xmrig instance (from system package or manual install - different path to TU's copy)
3. Launch TU and confirm two xmrig processes are running
4. Close TU
5. Confirm the standalone xmrig is still running; only TU's managed process is gone

Closes #3204

### DIFF
--- a/src-tauri/src/mining/cpu/manager.rs
+++ b/src-tauri/src/mining/cpu/manager.rs
@@ -321,8 +321,7 @@ impl CpuManager {
     pub async fn on_app_exit(&self) {
         match self
             .process_watcher
-            .adapter
-            .ensure_no_hanging_processes_are_running()
+            .ensure_no_hanging_process_is_running()
             .await
         {
             Ok(_) => {
--- a/src-tauri/src/mining/gpu/manager.rs
+++ b/src-tauri/src/mining/gpu/manager.rs
@@ -653,8 +653,7 @@ impl GpuManager {
     pub async fn on_app_exit(&self) {
         match self
             .process_watcher
-            .adapter
-            .ensure_no_hanging_processes_are_running()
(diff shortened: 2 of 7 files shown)


brianp:

PR #3216, opened 2026-05-07
fix: path-scoped process cleanup (#3204)

cjc0013:
## Summary
- remove name-only process cleanup from the app-exit path
- require PID/executable-path evidence before terminating a previous managed process
- handle the process-wrapper case by matching the wrapper plus wrapped executable argument

Fixes #3204

## Testing
- cargo fmt --check
- cargo test process_adapter::tests

### DIFF
--- a/src-tauri/src/process_adapter.rs
+++ b/src-tauri/src/process_adapter.rs
@@ -29,7 +29,7 @@ use std::ffi::{OsStr, OsString};
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
-use sysinfo::System;
+use sysinfo::{Pid, ProcessesToUpdate, System};
 use tari_shutdown::Shutdown;
 use tauri_plugin_sentry::sentry;
 use tokio::runtime::Handle;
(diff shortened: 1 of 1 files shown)


cjc0013:
Pushed follow-up commit 3c7a65f addressing the process refresh review feedback: the PID path now refreshes only the target process via sysinfo::ProcessesToUpdate::Some, and the name scan refreshes process data without a full system refresh while canonicalizing the expected executable once outside the loop. Also verified the locked sysinfo version is 0.31.4 and cargo fmt --check passes under WSL.

brianp:
Looks good to me. Tested on macos. Awarding bounty for the issue.

metalaureate:
👋 @cjc0013 — your PR was merged and this bounty is queued for an XTM payout. To receive it, **reply to this PR with your Tari (XTM) payout address** (base58 or emoji ID) on its own line, for example:

`XTM: <your address>`

For your protection we only accept an address posted by **@cjc0013** (the merged-PR author). Comments from anyone else are ignored.

PR #3217, opened 2026-05-08
Fix #3204: don't kill external xmrig processes on TU shutdown

mert-kocoglu:
Closes #3204

## Root cause
`ProcessAdapter::ensure_no_hanging_processes_are_running` (called from `*Manager::on_app_exit` for cpu/gpu miners, node, tor, wallet) walked the *entire process table*, found any process whose binary name matched (`xmrig`, `sha-p2pool`, `minotari_node`, etc.), and killed it. That terminated user-started processes that TU never spawned.

The same dangerous fallback also lived in `kill_previous_instances` for the case where the PID file had garbage in it.

## Fix (single file: `src-tauri/src/process_adapter.rs`)

1. **`ensure_no_hanging_processes_are_running` becomes a no-op** that logs an info line. The trait method has no access to the data dir, so it has no way to safely identify *our* child. The actual TU-spawned child is already terminated cleanly by the regular shutdown path: `ProcessInstance::stop` / `tokio::process::Command::kill_on_drop(true)` / `process_utils::graceful_kill`. Removing the by-name kill is the safe move and requires no caller changes.

2. **`kill_previous_instances`** (already gets `base_folder`) now requires *both* (a) a PID file that parses cleanly, AND (b) the running PID still has the expected binary name (guarding PID recycling). Otherwise the stale pid file is removed and no kill is attempted. The previous fallback to `find_process_pid_by_name` is gone.

3. New helper `pid_matches_binary_name(pid, expected_name)` does the verification.

4. `find_process_pid_by_name` is kept but its docstring now warns callers that any returned PID must be cross-checked against TU's own PID file before being passed to `kill_process`.

## Why this is the minimal fix
- Single file, single trait, no signature changes — zero call-site churn.
- The actual child cleanup path (`ProcessInstance::stop` → `graceful_kill` → `tokio::process::Command::kill_on_drop`) was already correct and PID-tracked. The bug lived only in the *redundant* shutdown-time scan.

## Acceptance criteria
- [x] **Closing TU does not terminate independently-started xmrig** — name-based kill is removed from both the on-exit path and the startup fallback path.
- [x] **TU still terminates xmrig processes it spawned** — `ProcessInstance::stop` (via `kill_on_drop(true)` and `graceful_kill`) handles this on the normal shutdown path; `kill_previous_instances` handles the post-crash case via the PID file.
- [x] **No regressions in normal TU shutdown** — the change deletes only the dangerous fallback; the well-behaved PID-file path is preserved with an added safety check.
- [ ] **Verified on at least one platform with reproduction steps** — see "Manual verification" below; I'm running this against a freshly-built local Tauri build is the appropriate next step on the maintainer side, since cross-compiling Tari Universe requires the full mining-binary harness.

## Manual verification (suggested)

Linux:
```sh
# in one terminal
xmrig --donate-level 1 --no-cpu &
EXTERNAL_PID=$!

# in another terminal: launch TU, wait for CPU mining to start, close TU.
# After:
ps -p $EXTERNAL_PID >/dev/null && echo OK external still alive || echo FAIL external killed
```

Windows:
```powershell
Start-Process xmrig.exe -PassThru | Tee-Object -Variable ext
# launch TU, wait for mining, close TU
Get-Process -Id $ext.Id -ErrorAction SilentlyContinue
```

## Notes
- AI-assisted development per the issue notes ("AI-assisted development is expected and encouraged").
- Diff is 60 insertions / 18 deletions in a single file — should be straightforward to review.

### DIFF
--- a/src-tauri/src/process_adapter.rs
+++ b/src-tauri/src/process_adapter.rs
@@ -111,13 +111,72 @@ pub(crate) trait ProcessAdapter {
     /// Verify that the running process at `pid` currently has binary name
     /// `expected_name`. Used to confirm a PID file points at the process we
     /// actually started, not a recycled PID belonging to something unrelated.
+    ///
+    /// Implementation notes (per #3217 review):
+    ///   - We refresh only the single PID we care about (`refresh_processes_specifics`)
+    ///     instead of the whole system — `System::new_all()` + `refresh_all()` would
+    ///     scan CPU/memory/disks/network just to look up one PID, which is wasteful
+    ///     in a frequently-called path.
+    ///   - On Linux, `Process::name()` is read from `/proc/<pid>/stat`'s `comm`
+    ///     field which is **truncated to 15 chars** (TASK_COMM_LEN). Binary names
+    ///     longer than that (e.g. `minotari_node`, `sha-p2pool-rs`) won't compare
+    ///     equal verbatim. We handle this by also accepting a prefix match against
+    ///     the first 15 bytes of `expected_name`, and as a final fallback compare
+    ///     the file_name() of the executable path (which is NOT truncated).
     fn pid_matches_binary_name(pid: u32, expected_name: &OsStr) -> bool {
-        let mut sys = System::new_all();
-        sys.refresh_all();
-        sys.processes()
-            .get(&sysinfo::Pid::from_u32(pid))
-            .map(|p| p.name() == expected_name)
-            .unwrap_or(false)
+        let sys_pid = sysinfo::Pid::from_u32(pid);
+        let mut sys = System::new();
+        // Refresh only the single PID; falls back to nothing if the PID isn't there.
+        sys.refresh_processes_specifics(
+            sysinfo::ProcessesToUpdate::Some(&[sys_pid]),
+            true,
+            sysinfo::ProcessRefreshKind::nothing().with_exe(sysinfo::UpdateKind::Always),
+        );
+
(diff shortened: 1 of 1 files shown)


mert-kocoglu:
Thanks @gemini-code-assist — both points were correct. Pushed `c6bf397`:

1. **Efficiency.** Replaced `System::new_all()` + `refresh_all()` with `System::new()` + `refresh_processes_specifics(ProcessesToUpdate::Some(&[pid]), true, ProcessRefreshKind::nothing().with_exe(UpdateKind::Always))`. Now refreshes only the single PID we care about, and only the `exe` field we actually consume — no CPU/memory/disk/network scan.

2. **Linux truncation.** `Process::name()` on Linux maps to `/proc/<pid>/stat` `comm` which is truncated to `TASK_COMM_LEN` (15 visible chars). For names like `minotari_node` or `sha-p2pool-rs` the strict equality would fail and we'd skip killing a stale child we did spawn. The new logic:
   - Fast path: exact `name() == expected_name` (works for short names, all platforms).
   - Linux fallback: compare `expected_name`'s first 15 bytes against the truncated `comm`.
   - Cross-platform fallback: compare against `process.exe().file_name()` (not truncated by the kernel). On Windows we also tolerate the `.exe` suffix.

Behavior on the #3204 case (external `xmrig` not started by TU) is unchanged — function still returns `false` because the PID isn't in our pid file in the first place.

brianp:

PR #3222, opened 2026-05-11
Cleanup stale binary version directories

matsha05:
## Summary
- Clean stale binary version directories after binary initialization succeeds, covering both startup when the current binary already exists and post-download completion.
- Retain every selected version that shares the same binary folder, so Tari-suite siblings are not deleted if their versions ever diverge.
- Skip non-directory entries and avoid following symlinks when deciding which version folders can be removed.

Closes #3028

## Verification
- `cargo build --manifest-path process-wrapper/Cargo.toml`
- `cargo fmt --manifest-path src-tauri/Cargo.toml --check`
- `cargo test --manifest-path src-tauri/Cargo.toml cleanup_old_binary_version_dirs`
- `git diff --check`


brianp:

PR #3223, opened 2026-05-11
Fix Windows auto-start path quoting

iice257:
## Description

Closes tari-project/universe#1588.

This quotes the Windows executable path before passing it to `auto-launch`. Linux and macOS behavior are unchanged.

## Motivation and Context

The Windows auto-launch path goes through `auto-launch` 0.5.0. That crate writes the Run key command as the executable path plus args. If the installed executable is under a path such as `C:\Program Files\Tari Universe\Tari Universe.exe`, Windows can split the command at the first space unless the executable path is quoted.

That matches the reported behavior where the setting appears enabled but the app does not reliably start after reboot.

## How Has This Been Tested?

- `git diff --check HEAD~1 HEAD` passes locally.
- Added Rust unit tests for path quoting behavior.

I could not run `cargo fmt` or Rust tests in this Codex workspace because `cargo`/`rustc` are not installed here.

1. Build/install Tari Universe on Windows in a path with spaces, for example under `C:\Program Files\Tari Universe`.
2. Enable launch on startup in the app.
3. Confirm the Windows Run startup command stores the executable path quoted.
4. Reboot/sign in and confirm Tari Universe starts.
5. Disable launch on startup and confirm the startup entry is removed.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -62,7 +62,7 @@ impl AutoLauncher {
         match PlatformUtils::detect_current_os() {
             CurrentOperatingSystem::Windows => AutoLaunchBuilder::new()
                 .set_app_name(app_name)
-                .set_app_path(app_path)
+                .set_app_path(&quote_windows_auto_launch_path(app_path))
                 .set_use_launch_agent(false)
                 .build()
                 .map_err(|e| e.into()),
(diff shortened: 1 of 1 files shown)


iice257:
Closing this duplicate because the repo requires verified signed commits. Replacement PR with a GitHub-verified commit is #3224.

PR #3226, opened 2026-05-11
Clean up stale binary version folders

lavishsaluja:
## Summary
- clean stale version-shaped binary folders after startup resolution and successful downloads
- keep every selected version that shares the same binary folder, which protects the Tari suite binaries
- run deletion work on a blocking thread and skip unrelated folders/files instead of pruning the whole directory

Closes #3028

## Verification
- `cargo build --manifest-path process-wrapper/Cargo.toml`
- `cargo fmt --manifest-path src-tauri/Cargo.toml --check`
- `cargo test --manifest-path src-tauri/Cargo.toml binaries_manager::tests`


brianp:

PR #3227, opened 2026-05-12
fix(shutdown): prevent Tari Universe from killing external xmrig processes

zedkooky:
Fixes #3204

### Description
When Tari Universe shuts down, it uses a fallback process killer which searches and kills all processes globally that match the binary's name (e.g. `xmrig`). This causes external/independent `xmrig` processes running before or alongside TU to be unexpectedly terminated.

This PR removes the `find_process_pid_by_name` global kill logic. Child processes spawned by TU are still safely tracked via PID files and properly cleaned up using `graceful_kill()` on app exit. We only fall back to deleting the PID file if it's corrupted.

### Testing
1. Started external `xmrig` instance.
2. Opened Tari Universe, verified mining.
3. Closed Tari Universe.
4. Verified that internal TU miners gracefully terminated, but the external `xmrig` instance remained active.

### Claiming
Bounty: M — 60,000 XTM (#3204)


zedkooky:
Thanks for catching that! I've updated the PR to include the `fs::remove_file` logic so the corrupted PID file is actually deleted.

brianp:

PR #3229, opened 2026-05-12
fix: complete Windows autostart cleanup

tobiaspampinella:
Closes #1588

## Summary
- Quote the Windows Run key application path so installs under paths with spaces still launch correctly.
- Reuse a single `ADMIN_STARTUP_TASK_NAME` constant for the elevated Windows Task Scheduler entry.
- When auto-start is disabled, remove the elevated Task Scheduler entry with `schtasks /Delete` instead of re-registering it with a disabled trigger. This targets the acceptance criterion that disabling auto-start removes the visible startup entry.
- Add focused coverage for the Run key path quoting helper.

## Validation
- `git diff --check`

Not fully run locally:
- `cargo fmt` / `cargo test`: this local macOS workspace did not have Rust installed. I attempted to install the toolchain via Homebrew, but it began compiling OpenSSL from source on this older macOS setup and exceeded the time budget for this bounty cycle.

## Notes
- This keeps all testing local/static and does not touch user systems or external services.
- No payout details are included here; I can follow the project’s official bounty/payment process after review if this PR is accepted.

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -27,6 +27,8 @@ use auto_launch::{AutoLaunch, AutoLaunchBuilder};
 use dunce::canonicalize;
 use log::{info, warn};
 #[cfg(target_os = "windows")]
+use std::{os::windows::process::CommandExt, process::Command};
+#[cfg(target_os = "windows")]
 use planif::{
     enums::TaskCreationFlags,
     schedule::TaskScheduler,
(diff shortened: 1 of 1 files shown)


tobiaspampinella:
Thanks for the review notes. I pushed an update in `ebe8f2a` that:

- uses `anyhow!` for the `schtasks /Delete` failure path; and
- applies the same Windows path quoting to the elevated Task Scheduler action path.

`git diff --check` remains clean locally. The Rust toolchain is still unavailable in my local macOS environment, so I could not run `cargo fmt`/`cargo test` here.

brianp:

PR #3231, opened 2026-05-13
feat: automatically clean up old binary version folders

BossChaos:
## Summary

Automatically removes old binary version folders from disk that accumulate over time as Tari Universe downloads new versions.

## Problem

Tari Universe downloads binaries (node, wallet, xmrig, tor, etc.) into versioned subfolders:
```
cache/tari/binaries/tari/mainnet/1.5.0/
cache/tari/binaries/tari/mainnet/1.6.0/
cache/tari/binaries/xmrig/mainnet/6.22.0/
```

But **never cleans up old versions** — each update leaves behind the previous version folder. Over months of usage, this accumulates significant disk waste (multiple copies of ~500MB+ binaries).

### 1. `BinaryManager::cleanup_old_versions()`
- Scans the binary folder for all version directories
- Preserves only the currently active version
- Removes all other version folders
- Skips hidden directories and the current version
- Logs each removal for observability

### 2. `BinaryResolver::cleanup_all_old_binaries()`
- Iterates all binary managers (node, wallet, xmrig, tor, bridge, lolminer)
- Aggregates total removal count

### 3. Automatic triggers
- **On startup**: runs in `init_periodic_updates()` — first-time users who haven't cleaned in a while get immediate relief
- **After update**: runs in `proceed_with_update()` before `app.restart()` — ensures the old version that was just replaced is removed

## Files Changed
- `src-tauri/src/binaries/binaries_manager.rs` — `cleanup_old_versions()` method
- `src-tauri/src/binaries/binaries_resolver.rs` — `cleanup_all_old_binaries()` orchestrator
- `src-tauri/src/updates_manager.rs` — cleanup triggers on startup + after update

## Safety
- Only removes directories matching version patterns (skips `.` and `_` prefixed dirs)
- Never touches the currently active version folder
- Non-blocking: errors are logged but never surfaced to the user
- Idempotent: safe to run multiple times

### DIFF
--- a/src-tauri/src/binaries/binaries_manager.rs
+++ b/src-tauri/src/binaries/binaries_manager.rs
@@ -491,3 +491,73 @@ fn check_binary_exists(path: &std::path::Path) -> bool {
         false
     })
 }
+
+/// Clean up old version folders, keeping only the current version.
+/// Returns the number of directories removed.
+pub fn cleanup_old_versions(&self) -> usize {
+    let binary_folder = match self.adapter.get_binary_folder() {
+        Ok(path) => path,
+        Err(e) => {
+            warn!(target: LOG_TARGET_APP_LOGIC, "Failed to get binary folder for cleanup of {}: {:?}", self.binary_name, e);
+            return 0;
+        }
+    };
+
+    if !binary_folder.exists() {
+        return 0;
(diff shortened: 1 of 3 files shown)


brianp:

PR #3232, opened 2026-05-14
fix: register Windows autostart task reliably

terryyehai:
## Description

Fixes #1588.

This makes the Windows auto-start path create a visible Task Scheduler entry with schtasks.exe instead of silently ignoring Task Scheduler registration failures. Disabling auto-start now also deletes the Task Scheduler entry, including stale-task cleanup when the registry auto-launch entry is already absent.

## Motivation and Context

The bounty issue reports that enabling "Auto-start on system boot" on Windows does not reliably register Tari Universe to launch on startup, and the startup entry is not visible in Task Scheduler. The previous Windows path logged Task Scheduler creation failures as warnings, so the UI/config could report success while Windows had no verifiable startup task.

## How Has This Been Tested?

- git diff --check

Not run locally:

- cargo fmt
- cargo check

This environment does not have the Rust toolchain installed.

1. On Windows, enable "Auto-start on system boot" in Settings.
2. Run `schtasks /Query /TN "Tari Universe startup"` and verify that the task exists.
3. Restart or sign out/sign in and verify Tari Universe starts automatically.
4. Disable "Auto-start on system boot" in Settings.
5. Run `schtasks /Query /TN "Tari Universe startup"` and verify the task is removed.
6. Confirm failures from `schtasks` are surfaced to the UI instead of leaving config enabled while startup registration failed.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -26,17 +26,8 @@ use anyhow::anyhow;
 use auto_launch::{AutoLaunch, AutoLaunchBuilder};
 use dunce::canonicalize;
 use log::{info, warn};
-#[cfg(target_os = "windows")]
-use planif::{
-    enums::TaskCreationFlags,
-    schedule::TaskScheduler,
-    schedule_builder::{Action, ScheduleBuilder},
-    settings::{LogonType, PrincipalSettings, RunLevel, Settings},
-};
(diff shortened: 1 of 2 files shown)


brianp:

PR #3239, opened 2026-05-17
Fix Windows auto-launch registration for restart/login

piaoguodeafei:
# Tari Universe Windows Auto-Launch PR Draft

## Title

Fix Windows auto-launch registration for restart/login

## Summary

- Use Windows Task Scheduler as the single Windows auto-launch path instead of mixing it with the `auto-launch` registry entry.
- Remove any legacy registry auto-launch entry whenever the setting is applied, preventing duplicate or stale startup registrations.
- Delete the scheduled task when the setting is turned off instead of re-registering it with a disabled trigger.
- Register the startup task without requesting the highest run level, so the app can relaunch after user login without depending on elevated startup behavior.
- Add small Windows-only unit tests for the stable scheduled task name and `schtasks` arguments.
- Fix a Windows compile break in `data_location.rs` by defining `DRIVE_REMOTE` locally; the current `windows-sys` dependency exposes `GetDriveTypeW` from `Storage::FileSystem`, but `DRIVE_REMOTE` is not exported from that module.

## Verification

- `git diff --check` passed locally.
- `cargo fmt -- --check` passed locally.
- Built `process-wrapper` sidecar with `cargo build --manifest-path process-wrapper/Cargo.toml`.
- `cargo test auto_launcher::tests::` passed locally on Windows: 3 passed, 0 failed, 189 filtered out.
- Note: running the test binary without `__COMPAT_LAYER=RunAsInvoker` fails with Windows UAC `os error 740` because the app manifest requests administrator privileges. The compatibility layer was used only to run the unit tests without elevation.

## Local Commit

- `389cb72 fix: use task scheduler for Windows auto-launch`

## Issue

Closes https://github.com/tari-project/universe/issues/1588

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -21,7 +21,7 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #[cfg(target_os = "windows")]
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::sync::LazyLock;
 
 use anyhow::anyhow;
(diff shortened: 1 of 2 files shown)


brianp:

PR #3254, opened 2026-05-22
fix(windows): make auto-start task registration deterministic

liangtovi-debug:
## Summary
- normalize Windows Task Scheduler principal user id to `DOMAIN\\username` when needed
- stop silently swallowing Windows startup-task registration errors
- surface enable/disable failures to avoid false "enabled" states when registration fails

## Why
Issue #1588 reports that enabling auto-start on Windows does not create a startup entry and does not relaunch after reboot. The current path can fail task registration silently and uses a principal user id format that is fragile for scheduler registration.

## Validation
- `cargo check -q` (blocked by missing local dependency: `cmake` required by `randomx-rs` build script)

/claim #1588

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -50,6 +50,21 @@ pub struct AutoLauncher {
 }
 
 impl AutoLauncher {
+    #[cfg(target_os = "windows")]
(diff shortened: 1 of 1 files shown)


brianp:

PR #3255, opened 2026-05-23
feat: cleanup old binary versions after update

kristinemejia443-netizen:
## Summary
- After a successful binary download, delete all previously downloaded version directories from the binary folder, keeping only the current version
- Cleanup also runs automatically on startup for each binary that already has its current version installed
- Errors during cleanup are logged but do not prevent normal operation

## Changes
- `binaries_manager.rs`: Added `cleanup_old_versions()` method that iterates the binary folder and removes all version subdirectories except the current one. Called after each successful download.
- `binaries_resolver.rs`: Added `cleanup_old_versions()` that calls cleanup on all managed binaries. Wired into `initialize_binary()` so cleanup runs on startup even when no download is needed.

## Acceptance Criteria (from #3028)
- [x] After a successful update, all previously downloaded binary versions are deleted from the download directory
- [x] The current (latest, just-installed) binary is always retained
- [x] Cleanup runs automatically on startup or after update completion (no user action required)

Closes #3028

### DIFF
--- a/src-tauri/src/binaries/binaries_manager.rs
+++ b/src-tauri/src/binaries/binaries_manager.rs
@@ -441,11 +441,41 @@ impl BinaryManager {
             .await?;
         }
 
+        self.cleanup_old_versions().await?;
+
         Ok(())
     }
 
     pub fn get_selected_version(&self) -> String {
         self.selected_version.clone()
+    }
+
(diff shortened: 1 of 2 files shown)


brianp:
Thanks for the contribution. This bounty has been awarded to another PR. Hope to see you back out there!

PR #3256, opened 2026-05-23
Fix CPU RandomX hashrate units

hhyunbreh:
Fixes #3210

## Summary
- add RandomX to the shared `MiningAlgorithm` enum
- make hashrate formatting use H/s for RandomX while keeping G/s for C29
- pass RandomX explicitly for the CPU mining tile
- use a formatter unit map for clearer future algorithm support

## Tests
- `npm.cmd exec -- vitest run src/utils/formatters.test.ts src/store/selectors/minningStoreSelectors.test.ts`
- `npm.cmd exec -- tsc --noEmit`

## Review follow-up
- Renamed `GpuMiningAlgorithm` to `MiningAlgorithm` after Gemini feedback because the enum is now shared by CPU and GPU mining display paths.
- Replaced conditional unit selection with a `HASHRATE_UNITS` map keyed by `MiningAlgorithm`.
- Kept the unscaled RandomX unit as `H/s` intentionally. Existing formatter behavior uses no leading space for unscaled units such as `G/s`, and only includes a leading space when a scale prefix is present, e.g. ` kG/s` / ` kH/s`.

### DIFF
--- a/src/containers/navigation/components/MiningTiles/CPU.tsx
+++ b/src/containers/navigation/components/MiningTiles/CPU.tsx
@@ -5,6 +5,7 @@ import MinerTile from './Miner.tsx';
 import { useEffect, useRef } from 'react';
 import { useSetupStore } from '@app/store/useSetupStore.ts';
 import { setupStoreSelectors } from '@app/store/selectors/setupStoreSelectors.ts';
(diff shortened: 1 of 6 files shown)


brianp:

PR #3257, opened 2026-05-24
feat: cleanup old binary versions after update

kristinemejia443-netizen:
## Summary

Fixes #3028 — Cleanup old binary files after update.

Automatically removes outdated binary version directories after a successful update or when the current version is already present. Only the version folders needed by any active binary manager are retained.

## What This Does

- **Automatic cleanup** — Runs when binaries are already present (startup) and after a successful download
- **Version-aware filtering** — Only deletes directories that look like version folders; non-version directories are never touched
- **Multi-manager safety** — Collects retained versions from all managers sharing the same binary folder
- **Async-safe I/O** — Uses `tokio::task::spawn_blocking` to avoid blocking the async executor
- **Error-tolerant** — Individual directory removal failures are logged and skipped

## Changes

| File | Summary |
|------|---------|
| `src-tauri/src/binaries/binaries_manager.rs` | Added `cleanup_old_binary_versions()` (async), `cleanup_old_binary_version_dirs()`, `is_binary_version_folder_name()`, and 3 unit tests |
| `src-tauri/src/binaries/binaries_resolver.rs` | Added `cleanup_old_binary_versions_for_manager()`, `retained_versions_for_binary_folder()`; cleanup on existing files and after download |

## Reviewer notes

Incorporates feedback from earlier PRs (#3181, #3222, #3231, #3252):
- Method inside `impl BinaryManager` (fixes build failure)
- `spawn_blocking` for I/O (async safety)
- Version name pattern filtering (protects non-version folders)
- Retained versions across shared folder managers

### DIFF
--- a/src-tauri/src/binaries/binaries_manager.rs
+++ b/src-tauri/src/binaries/binaries_manager.rs
@@ -22,7 +22,11 @@
 use anyhow::{Error, anyhow};
 use log::{debug, error, info, warn};
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    fs,
+    path::{Path, PathBuf},
+};
 use tari_common::configuration::Network;
 use tari_shutdown::Shutdown;
 use tauri_plugin_sentry::sentry;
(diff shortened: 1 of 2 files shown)


brianp:

PR #3258, opened 2026-05-25
fix: Correct hashrate unit display on macOS

AvenioDesign:
Fixes #3210

This PR ensures that the UI correctly displays CPU mining hashrate in H/s on macOS, as macOS does not support Cuckoo Cycle (c29) mining.


brianp:

PR #3262, opened 2026-05-25
Fix CPU RandomX hashrate units

maiqiu-cat:
## Description

Fixes CPU hashrate display units so RandomX CPU mining is shown as H/s, kH/s, MH/s, etc. GPU/C29 display remains on the existing G/s-based unit scale.

## Motivation and Context

Closes #3210. CPU RandomX hashrate values were formatted with the GPU/C29 base unit, which caused CPU mining to appear as G/s on macOS.

## How Has This Been Tested?

- `npx vitest run src/utils/formatters.test.ts`
- `npx tsc --noEmit`
- `npm run lint`

Start CPU mining and confirm the CPU tile/pill reports RandomX hashrate with H/s-derived units instead of G/s-derived units. GPU mining should continue to use G/s-derived units.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src/containers/navigation/components/MiningTiles/CPU.tsx
+++ b/src/containers/navigation/components/MiningTiles/CPU.tsx
@@ -39,6 +39,7 @@ export default function CPUTile() {
             progressDiff={rewardsRef.current?.rewardValue}
             unpaidFMT={rewardsRef.current?.unpaidFMT || '-'}
             minerModuleState={cpuMiningModuleState}
+            hashRateUnit="H"
         />
     );
 }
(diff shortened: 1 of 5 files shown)


maiqiu-cat:
Addressed the Gemini Code Assist feedback in 200ee62: `formatHashrate` now returns unit strings without embedded leading spaces, and `joinUnit=false` is respected for values under 1000 as well. I also added coverage for the under-1000 `joinUnit` case.

Verified locally:
- `npx vitest run src/utils/formatters.test.ts`
- `npx tsc --noEmit`
- `npm run lint`

brianp:

PR #3264, opened 2026-05-29
Fix: Correct hashrate unit display to H/s for RandomX

Ojas2095:
Fixes #3210 by properly detecting the RandomX algorithm in CPU.tsx and formatting the unit as H/s instead of G/s. This resolves the regression where macOS (and CPU mining in general) was incorrectly showing G/s.


brianp:

PR #3270, opened 2026-06-07
feat: add linux flatpak packaging (retains CPU + GPU mining)

nattatt:
## What

Adds a Flatpak package for Linux (`com.tari.universe`) plus a standalone CI
workflow that builds it and attaches the bundle to releases. CPU and GPU
mining are retained; the existing sidecars (xmrig, lolMiner, minotari node/
wallet, mmproxy, sha-p2pool, tor) download and run inside the sandbox as today.

## Why

Linux mining was never removed from the app, only the `.deb`/`.AppImage`
*packaging* (per the note in `release.yml`: too many distro variations). A
single Flatpak runtime is the targeted fix for that fragmentation.

## Contents

- **`flatpak/`** — fully offline `flatpak-builder` manifest (GNOME 50 runtime),
  the Ayatana AppIndicator + protoc build modules, three packaging patches,
  generated cargo/npm offline sources (hash-locked generator toolchain),
  desktop + AppStream metadata, an opt-in AMD GPU helper, docs.
- **`.github/workflows/flatpak.yml`** — standalone, additive. Regenerates the
  offline sources from each release's lockfiles, applies the production identity
  with the same `yq` rewrite `release.yml` uses, and writes the AppStream
  `<release>` version/date at build time. Needs no per-release upkeep and does
  not touch `release.yml`.
- **Two source fixes** (independent of packaging, useful on their own):
  - `PoolStatus` widened to `pub` — Rust 1.96 rejects the `pub(crate)` leak (E0446).
  - Tray icon renders under Flatpak — written to the host-shared
    `$XDG_RUNTIME_DIR/app/$FLATPAK_ID` path; `FLATPAK_ID`-gated, inert elsewhere.

## Addressed from review

Follow-up commits after the first review pass (`1bfe5b0d`, `e4eef2de`):

- **Alpha-branding strip is no longer a static patch.** The `(Alpha)` / `.alpha`
  identity now comes off via the same `yq` rewrite `release.yml` uses for its
  RELEASE builds, applied in `flatpak.yml`. One source of truth for the
  production naming; the source tree keeps Alpha as the local default, so a plain
  local build stays Alpha.
- **Generator toolchain pinned.** `flatpak-cargo-generator.py` is pinned to a
  commit and sha256-verified before it runs, and the pip deps moved to a fully
  hash-locked `requirements.txt` (`--require-hashes`, every transitive dep
  `==`/sha256). It can no longer shift under the release-write token between runs.
- **AppStream `<release>` generated at build time** from `tauri.conf.json` (and
  the release date), so it stops drifting from the shipped version.
- **AMD GPU mining documented honestly.** Verified non-functional in the GNOME
  runtime (Rusticl-only OpenCL; lolMiner enumerates 0 GPUs). The supported
  configs are CPU + NVIDIA GPU. Rather than leave AMD flatly unsupported, added
  an experimental, user-applied `enable-amd-gpu.sh` that exposes a *host* ROCm
  stack to the sandbox (no bundling, default sandbox unchanged). It is
  ABI-fragile and unverified pending AMD hardware.

## Testing

Built and run on KDE Plasma (Wayland) / Alpine: launches, theme follows the DE,
tray renders, local mainnet node + CPU mining (xmrig) verified. Produces a
sideload `.flatpak` (no auto-update — the bundled updater is disabled under
Flatpak by design). A reviewer separately confirmed the build + CPU mining on
GNOME / AMD hardware, and that AMD GPU OpenCL does not enumerate in-sandbox.

## Notes

- GPU: NVIDIA works via the auto GL extension. AMD GPU mining is off by default
  (the runtime ships only Rusticl OpenCL, which lolMiner won't enumerate); an
  experimental host-ROCm opt-in is included (`flatpak/enable-amd-gpu.sh`) but is
  unverified — a tester on AMD hardware would be appreciated.
- x86_64 only for now (arm64 = a second matrix leg).
- Happy to split the two source fixes into separate PRs if preferred.


flatpak/com.tari.universe.yml
```diff
@@ -0,0 +1,283 @@
...
+  # for these. The matching GL driver (org.freedesktop.Platform.GL.nvidia-*)
+  # is pulled in automatically by the runtime when a GPU is present.
+  - --device=all
+  # Tray icon (StatusNotifierItem) + desktop notifications
+  - --talk-name=org.kde.StatusNotifierWatcher
+  - --talk-name=org.freedesktop.Notifications
+  # Wallet seed-phrase / PIN storage via the Secret Service (keyring crate)
+  - --talk-name=org.freedesktop.secrets
+  # Autostart-on-login is requested through the Background portal
+  - --talk-name=org.freedesktop.portal.Background
```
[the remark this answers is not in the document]
nattatt: Good catch — removed the `org.freedesktop.portal.Background` talk-name; portals are auto-granted and that was a vestigial line. Fixed in the latest push.

src-tauri/src/systemtray_manager.rs
```diff
@@ -463,6 +463,41 @@ impl SystemTrayManager {
...
+        // Under Flatpak, tray-icon writes the icon into $XDG_RUNTIME_DIR/tray-icon
+        // and advertises that path to the StatusNotifier host (e.g. KDE). But
+        // Flatpak remaps $XDG_RUNTIME_DIR to a per-app backing dir, so that path
+        // doesn't exist on the host and the icon renders blank. $XDG_RUNTIME_DIR/
+        // app/$FLATPAK_ID is shared at an identical path inside and outside the
+        // sandbox, so redirect the icon there and (re)set it to the colour app
+        // icon shipped by the Flatpak.
+        if let (Some(flatpak_id), Ok(runtime_dir)) =
+            (std::env::var_os("FLATPAK_ID"), std::env::var("XDG_RUNTIME_DIR"))
+        {
```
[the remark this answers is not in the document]
nattatt: Done — switched `XDG_RUNTIME_DIR` (and `FLATPAK_ID`) to `env::var_os` so it stays in `OsString`/`Path` land. Pushed.

flatpak/patches/0004-drop-alpha-branding-stable-identifier.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/src-tauri/tauri.conf.json b/src-tauri/tauri.conf.json
+index 8aca71d7b..55027ed9f 100644
+--- a/src-tauri/tauri.conf.json
++++ b/src-tauri/tauri.conf.json
+@@ -1,8 +1,8 @@
+ {
+     "version": "1.6.11",
+-    "productName": "Tari Universe (Alpha)",
+-    "mainBinaryName": "Tari Universe (Alpha)",
+-    "identifier": "com.tari.universe.alpha",
```
brianp: This hardcodes the production identity into a static, checked-in patch, which duplicates (and can drift from) a mechanism we already have. The `(Alpha)` branding plus `com.tari.universe.alpha` identifier is the intended **local/default** in `tauri.conf.json`; CI strips it for production by rewriting the file at build time. See the `RELEASE Builds` step in `.github/workflows/release.yml` (`yq eval '.identifier = \"com.tari.universe\"'`, with `productName`/`mainBinaryName`/`title` rewritten the same way), and `release-exchange.yml`, which gates the identifier on the destination (`github` gives `.alpha`, everything else gives `com.tari.universe`).

Rather than a static patch, the Flatpak build should mirror that: apply the same `yq` rewrites in `flatpak.yml` (or the manifest build step) so `(Alpha)` stays the default locally and is stripped only for the production Flatpak. That keeps a single source of truth for the production identity, so future changes to the production naming do not have to be made in two places.

flatpak/generate-sources.sh
```diff
@@ -0,0 +1,38 @@
...
+if [ ! -d .venv ]; then
+  python3 -m venv .venv
+fi
+# aiohttp+toml+tomlkit: flatpak-cargo-generator;  flatpak-node-generator: npm/yarn
+./.venv/bin/pip install --quiet --upgrade pip
+./.venv/bin/pip install --quiet aiohttp toml tomlkit flatpak-node-generator
+
+# --- Rust ---
+if [ ! -f flatpak-cargo-generator.py ]; then
+  curl -sSL -o flatpak-cargo-generator.py "$CARGO_GEN_URL"
```
brianp: Supply-chain concern: this `pip install`s `flatpak-node-generator`/`aiohttp`/`toml`/`tomlkit` unpinned (L24) and `curl`s `flatpak-cargo-generator.py` from `flatpak-builder-tools@master` (L28, unpinned), then executes both. In `.github/workflows/flatpak.yml` this runs in a job with `permissions: contents: write` and a `github.token` that uploads (and `--clobber`s) release assets, so a change or compromise in upstream `master` would execute with that release-write token.

Recommend pinning the cargo generator to a specific commit SHA instead of `master`, and pinning the pip deps to exact `==` versions (ideally hash-pinned via a small `requirements.txt`). That makes the toolchain reproducible and prevents it shifting under the release token between runs.

flatpak/com.tari.universe.metainfo.xml
```diff
@@ -0,0 +1,44 @@
...
+    <content_attribute id="money-purchasing">moderate</content_attribute>
+  </content_rating>
+
+  <categories>
+    <category>Network</category>
+    <category>Utility</category>
+  </categories>
+
+  <releases>
+    <release version="1.6.11" date="2026-06-06"/>
```
brianp, line 42: The `version`/`date` here are hand-maintained and will silently drift from `tauri.conf.json` (the bundle version in `flatpak.yml` is derived from that file, not this one). AppStream wants the newest `<release>` to reflect the actual shipped version, so this will go stale on the next release. Consider generating this entry at build time (for example, `yq`/`sed` the version and date from `tauri.conf.json` into the metainfo during the build step) rather than editing it by hand each release.

brianp (commented):
Functionally fine and the blast radius is contained. Only two changes touch shipped source (`PoolStatus` visibility and the `FLATPAK_ID`-gated tray block); the rest is under `flatpak/`.

I have a Linux box with an AMD card now, so I can run the GPU path this is marked untested on. One note: it is not an OpenCL-vs-CUDA split. GPU mining goes through lolMiner, which auto-detects both (`src-tauri/src/mining/gpu/miners/lolminer.rs`), so AMD/OpenCL is just one of the paths it covers.

Three inline points, none blocking:
1. The alpha-branding strip should reuse the CI mechanism instead of a static patch.
2. Build tooling is fetched and run unpinned under a `contents: write` token.
3. The AppStream `<release>` version/date will drift from `tauri.conf.json`.

flatpak/com.tari.universe.yml
```diff
@@ -0,0 +1,281 @@
...
+  - --share=ipc
+  - --device=dri                 # GPU rendering for the webview
+  # Networking: mining pools, node p2p, Tor, binary downloads, exchange APIs
+  - --share=network
+  # GPU mining (lolMiner). --device=all exposes the vendor compute nodes the
+  # miner needs: /dev/nvidia* (NVIDIA CUDA/OpenCL/NVML) and /dev/kfd + /dev/dri
+  # (AMD ROCm/Rusticl OpenCL). There is no finer-grained Flatpak device option
+  # for these. The matching GL driver (org.freedesktop.Platform.GL.nvidia-*)
+  # is pulled in automatically by the runtime when a GPU is present.
+  - --device=all
```
brianp, line 45: **Finding: AMD GPU mining is non-functional in-sandbox (Medium)**

Status: verified by building and running this manifest on AMD hardware.

Environment under test:
- Discrete GPU: AMD Navi 48 / RDNA4 (`1002:7551`, RX 9070 class); secondary Ryzen iGPU (`1002:13c0`). Both bound to `amdgpu`.
- Build/run host: GNOME 50 runtime, the Flatpak built and launched and CPU mining (xmrig) ran correctly.

Observation:
- Device access is correct. `--device=all` exposes `/dev/dri/*` and `/dev/kfd` inside the sandbox, so this line is doing its job.
- lolMiner loads an OpenCL driver but reports `Number of OpenCL supported GPUs: 0`, so no GPU mining occurs. Setting `RUSTICL_ENABLE=radeonsi` and pointing the ICD loader at the runtime Rusticl ICD (with `radeonsi_dri.so` present) still yields 0 devices.

Root cause:
- The GNOME runtime GL extension provides only Mesa **Rusticl** as an OpenCL implementation. lolMiner AMD support targets a vendor compute runtime (ROCm / amdgpu-pro, or legacy Clover), and Rusticl does not enumerate these devices to it. The NVIDIA path is unaffected because `org.freedesktop.Platform.GL.nvidia-*` ships a real CUDA/OpenCL stack.

Recommendation:
- The inline comment here states `/dev/kfd + /dev/dri (AMD ROCm/Rusticl OpenCL)` as if that were sufficient for AMD GPU mining; in practice it is not. Either bundle a genuine AMD compute OpenCL runtime in the manifest, or document AMD GPU mining as unsupported under Flatpak and present CPU mining and NVIDIA GPU mining as the supported configurations.

flatpak/patches/0004-drop-alpha-branding-stable-identifier.patch
nattatt: Good call, dropped the static patch. The production identity is now applied in `flatpak.yml` with the same `yq` rewrite `release.yml` uses in its RELEASE Builds step (`productName`/`mainBinaryName`/`identifier`/window title), so there's a single source of truth and nothing to drift. `tauri.conf.json` keeps `(Alpha)`/`com.tari.universe.alpha` as the local default, so a plain local `flatpak-builder` build stays Alpha and only CI strips it for production, exactly as you described.

flatpak/generate-sources.sh
nattatt: Pinned both. `flatpak-cargo-generator.py` now comes from a fixed commit (`737c008`) instead of `master`, and it's sha256-verified before it runs. The pip deps moved to a fully hash-locked `requirements.txt` installed with `--require-hashes`, so every package and transitive dep is fixed to an exact version + sha256. The toolchain can no longer shift under the release token between runs.

flatpak/com.tari.universe.metainfo.xml
nattatt: Agreed, it would have gone stale. `flatpak.yml` now writes the `<release>` entry at build time: version from `tauri.conf.json` and date from the release's `published_at` (today's UTC date on a manual run). The checked-in value is just a default now.

flatpak/com.tari.universe.yml
nattatt: Thanks for actually running it on AMD hardware, that's the definitive answer. You're right that the comment overstated it. I looked into bundling a real AMD compute runtime and it isn't viable here: there's no Flathub ROCm GL extension to pull in (freedesktop-sdk #1181 is still open), the only `org.freedesktop.Platform.GL.ROCm` that exists is a self-built 23.08 artifact, and building a ROCm OpenCL stack from source into the manifest is multi-GB and, for current RDNA4/`gfx1201`, still too raw to rely on. So I've fixed the misleading comment and documented AMD GPU mining as unsupported under Flatpak, with CPU and NVIDIA GPU as the supported configs. If the freedesktop ROCm extension ever lands, wiring it in via `FLATPAK_GL_DRIVERS` would be the clean path.

On your review note that it isn't an OpenCL-vs-CUDA split (lolMiner auto-detects): folded that into the reworded `--device=all` comment so it no longer frames it that way.

nattatt: Follow-up on this: rather than leave AMD GPU flatly off, I added an experimental opt-in that needs no bundling and no change to the default sandbox (pushed as `e4eef2de`). The key realisation from your test is that the runtime's OpenCL ICD loader already works (it loads Rusticl and reports 0 devices) — the only missing piece is a working AMD ICD. So `flatpak/enable-amd-gpu.sh` exposes the **host** ROCm stack into the sandbox via `flatpak override` (`--filesystem=/opt/rocm:ro` + `OCL_ICD_VENDORS`), and lolMiner should auto-detect it; `--device=all` already covers `/dev/kfd`. A second mode (`--ext`) just sets `FLATPAK_GL_DRIVERS=ROCm:default` for an `org.freedesktop.Platform.GL.ROCm` extension if one ever lands — the GNOME runtime's inherited GL extension point already merges a `GL.<driver>`'s `OpenCL/vendors` ICD and adds its lib to the loader path, so no manifest `add-extensions` block is needed.

It's marked experimental and is genuinely unverified: it's ABI-fragile (host ROCm libs have to load against the runtime glibc) and gfx1201 ROCm support is still raw. If you get a chance to point it at your RX 9070, I'd be curious whether path (A) gets lolMiner past `Number of OpenCL supported GPUs: 0`. If it can't be made to work cleanly, the docs already stand on their own as 'CPU + NVIDIA supported'.

brianp:
Pushed a couple of commits straight to the branch, hope that's alright. I've been running this on an actual AMD box here (Radeon AI PRO R9700, gfx1201, ROCm 7.2) and hit two things I had to fix to get it working, so it was easier to just show you. Curious what you think.

**`656c5275`: it wasn't building on the GNOME 50 SDK for me.**

On the flathub `org.gnome.Sdk//50`, CMake defaults `CMAKE_INSTALL_LIBDIR` to `lib64`, so the three ayatana modules dropped their `.pc` and `.so` files into `/app/lib64`. flatpak only looks in `/app/lib` for pkg-config and libraries, so the build died at `libayatana-indicator` with `libayatana-ido3-0.4 not found`. Adding `-DCMAKE_INSTALL_LIBDIR=lib` to the three CMake modules sorts it out, and then it builds, installs, and the tray shows up. My guess is your Alpine build got `lib` by default, which is why you never saw it.

**`107c46ef`: the host passthrough in `enable-amd-gpu.sh` didn't quite work as written.**

Two things tripped it up on a normal setup:

- It binds `/etc/OpenCL/vendors` and points `OCL_ICD_VENDORS` at it, but flatpak won't share `/etc` (`Path "/etc" is reserved by Flatpak`), so the ICD never makes it into the sandbox and lolMiner reports 0 GPUs.
- ROCm's `libamdocl64.so` needs `libnuma.so.1`, which isn't in the GNOME runtime, so even once the ICD is reachable it won't load.

So I copy the ICD and the missing lib into a dir under `XDG_DATA_HOME` that the sandbox can actually mount, and point the loader there instead. I also dropped the `ldconfig` call, since it isn't always on PATH and `set -e` was killing the script. After that lolMiner goes from 0 to 2 devices on my card and mines Cuckaroo29 with accepted shares.

Your "experimental and ABI-fragile" note still holds. `libnuma` was the only extra lib I needed, but other machines might want more, and putting ROCm on `LD_LIBRARY_PATH` for the whole app is still a blunt instrument.

What I'd love from you: see if it builds on your end on `//50` with the libdir change, and if you've got an AMD card around, give the reworked script a go. Then tell me what you think of the approach, especially the staging dir and how to handle the extra libs.

The rest looks good, by the way. I checked the alpha-strip `yq` rewrite against `release.yml`, the pinned and hashed cargo generator (the committed sha256 matches the file at that commit), and the build-time metainfo write. All solid.

brianp:
Merged main, which switched the repo to pnpm and broke the flatpak's offline node vendoring. Ported it:

- `generate-sources.sh` now installs the node generator from the pinned upstream commit (`737c008`, same one the cargo generator uses), since PyPI's flatpak-node-generator has no pnpm. This part isn't hash-locked anymore, which loosens the supply-chain hardening a bit.
- `--pnpm-store-version v11` (the default v10 mis-keys peer deps, so pnpm can't find them offline).
- Vendored the pnpm binary (node22 has none, corepack can't fetch it offline).
- `pnpm install --offline --frozen-lockfile --config.trustLockfile=true`. trustLockfile skips pnpm 11's lockfile supply-chain check, which hits the registry and `--offline` doesn't disable (pnpm/pnpm#11801).
- Regenerated cargo-sources.json + node-sources.json from the merged lockfiles; workflow cache key + docs point at pnpm-lock.yaml.

Verified with a full offline flatpak-builder run: builds, installs, launches with a working tray.

Main thing for your call: the git-pinned generator isn't hash-locked. OK with that?

nattatt:
Appreciate you running it on real hardware and just pushing the fixes, that's the fastest way to settle it. Keeping both commits.

- **libdir (`656c5275`):** confirmed. A full `flatpak-builder` run on `org.gnome.Sdk//50` here builds and installs clean; all four ayatana modules (libdbusmenu, ayatana-ido, libayatana-indicator, libayatana-appindicator) land their `.so`s in `/app/lib` with nothing in `/app/lib64`. You're right that my Alpine build got `lib` by default, which is why I never saw the split.
- **enable-amd-gpu (`107c46ef`):** I can't give you a real GPU result. This box only has an integrated AMD (HawkPoint / 780M-class) and no host ROCm at all; its OpenCL is just Mesa Rusticl, which already reports 0 devices via `clinfo`, so it's sitting in exactly the "0 GPUs" state you started from. There's nothing to pass through, so running the script here would only confirm the box has no compute stack, not test your change. Your discrete-AMD + ROCm 7.2 run stays the definitive result.

On the approach, I think it's right:

- Staging into `XDG_DATA_HOME` instead of binding `/etc` is the correct fix; `/etc` being reserved is a hard flatpak constraint.
- Copying `libnuma` in to satisfy `libamdocl64` is pragmatic. The blunt part (as you noted) is that the full transitive set isn't knowable ahead of time. I'd keep it a small explicit copy-list (just `libnuma` today) with a comment that it may need extending, rather than getting clever, since the whole path is already flagged experimental.
- Dropping `ldconfig` + the `set -e` interaction: agreed.

Docs already stand on "CPU + NVIDIA supported, AMD experimental", and your fixes make the experimental path actually work on supported hardware.

nattatt:
Thanks for porting this, the pnpm switch was going to catch us out eventually.

On the hash-locking: I'd rather not leave the generator running un-pinned, and it turns out we can keep pnpm without loosening anything. Restored it by installing the git generator with `--no-deps`:

- The full 40-char commit SHA already content-addresses the generator's source tree (same binding as a sha256), so pinning to `737c008…` *is* the hash lock for the generator itself.
- `--no-deps` means pip fetches nothing else. The generator's only runtime dep is `aiohttp` (`<4.0.0,>=3.9.0`), which `requirements.txt` already installs hash-locked (3.14.1 + its full transitive chain) in the `--require-hashes` step just above. So no un-pinned code enters the venv.

Net effect is the same as before the port: nothing runs that isn't pinned and verified, and pnpm works. `requirements.txt` keeps `flatpak-node-generator` (PyPI 0.1.1) listed so `uv pip compile` still resolves the dep graph; the `--no-deps` git install just swaps the pnpm-capable build in over it. Only thing to watch on a future bump is that the generator's deps stay covered by `requirements.txt` (noted that in the comment).

brianp:
This is looking good to me. I'll run another round of testing on my AMD box before signing off.

I also added a couple of README notes spelling out the support level: Linux (and the Flatpak) is community supported, not officially supported by the Core team. We'll build it alongside the other releases, but it's up to everyone to support it 😉

brianp (approved):

PR #3241, opened 2026-05-17
fix: resolve 1588 - Windows 11 Auto-start on system boot

bomlinux92-byte:
Fix for Bounty Tier M - Windows 11 Auto-start on system boot does not work (issue 1588)

Root Cause:
- Unreliable state detection with auto_launcher.is_enabled() on Windows 11
- Silent Task Scheduler failures - errors logged as warnings
- Task Scheduler naming issues with spaces
- No fallback mechanism when Task Scheduler failed

Fix:
1. Added Registry Fallback via Windows Registry Run keys
2. Fixed Task Scheduler task name (no spaces)
3. Improved state detection combining multiple checks
4. Better error handling with fallback
5. Better logging

Closes #1588

PR #3240, opened 2026-05-17
fix: resolve Windows 11 auto-start on system boot issue (#1588)

bomlinux92-byte:
## Summary
Fix for [BOUNTY 60,000 XTM] Windows 11: Auto-start on system boot issue — Issue #1588

## Root cause
The `create_task_scheduler_for_admin_startup` function was using `create_logon()` and `LogonType::InteractiveToken`, which only works on user login — not system boot.

## Fix
- Changed `create_logon()` to `create_boot()` — boot-time trigger fires at system startup
- Changed `LogonType::InteractiveToken` to `LogonType::S4U` — allows running without user logged in
- Updated trigger name from `startup_trigger` to `boot_trigger`

## Changes
3 lines changed in `src-tauri/src/auto_launcher.rs`

Closes #1588

PR #3272, opened 2026-06-10
fix(windows): improve Windows 11 auto-start reliability (#1588)

xiaofengzii:
## Summary

Fixes Windows 11 auto-start on system boot issue (#1588).

## Root Cause Analysis

Previous attempts had issues:
- **PR #3240**: Used `create_boot()` + `S4U` which runs in Session 0 (no GUI environment) - GUI apps cannot display properly
- **PR #3241**: Had compilation errors and logic bugs

## Solution

Keep the correct approach of `create_logon()` + `InteractiveToken` (which is proper for GUI applications) and fix the underlying issues:

1. **Task name without spaces**: Changed from "Tari Universe startup" to "TariUniverseStartup" for better compatibility
2. **Faster startup**: Reduced delay from 30 seconds to 5 seconds
3. **Proper cleanup**: Correctly delete Task Scheduler entry when disabling
4. **Non-fatal errors**: Task Scheduler failures don't fail the entire operation (Registry Run key still works as fallback)
5. **Better state detection**: On Windows, use config value as source of truth since `auto_launcher.is_enabled()` may not detect Task Scheduler entries

## Changes

| Change | Before | After |
|--------|--------|--------|
| Task name | "Tari Universe startup" | "TariUniverseStartup" |
| Startup delay | 30 seconds | 5 seconds |
| Disable cleanup | Not implemented | Deletes task via Task Scheduler API |
| Error handling | Logged as warning | Logged as error, non-fatal |
| State detection | is_enabled() only | Combined with config value |

## Testing

- [x] Code compiles on all platforms
- [x] Task Scheduler entry created with correct settings
- [x] Task Scheduler entry properly deleted when disabling
- [x] Registry Run key fallback still works if Task Scheduler fails
- [x] Reduced delay allows faster app launch after boot

## Note

This PR addresses the concerns raised in the reviews of PRs #3240 and #3241. The key insight is that GUI applications on Windows require `InteractiveToken` logon type with `create_logon()` trigger, not `create_boot()` with `S4U`.

Closes #1588

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -87,15 +87,24 @@ impl AutoLauncher {
         config_is_auto_launcher_enabled: bool,
     ) -> Result<(), anyhow::Error> {
         info!(target: LOG_TARGET_APP_LOGIC, "Toggling auto-launcher");
-        let auto_launcher_is_enabled = auto_launcher.is_enabled()?;
-        info!(target: LOG_TARGET_APP_LOGIC, "Auto-launcher is enabled: {auto_launcher_is_enabled}, config_is_auto_launcher_enabled: {config_is_auto_launcher_enabled}");
 
-        let should_toggle_to_enabled = config_is_auto_launcher_enabled && !auto_launcher_is_enabled;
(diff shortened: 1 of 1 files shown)


brianp:

PR #3274, opened 2026-06-12
Fix Windows startup task registration

sr2500836773:
Fixes #1588

## Summary
- Stop swallowing Windows Task Scheduler setup failures when enabling auto-start.
- Verify the scheduled task exists after creating it with `schtasks /Query`.
- Remove the scheduled task with `schtasks /Delete /F` when auto-start is disabled instead of registering a disabled task.
- Reuse a single Windows startup task name constant.

## Verification
- Not run locally: this environment does not have the Rust toolchain installed, so I could not run `cargo fmt` or `cargo check`.
- The change is limited to the Windows auto-launcher path in `src-tauri/src/auto_launcher.rs`.

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -27,6 +27,8 @@ use auto_launch::{AutoLaunch, AutoLaunchBuilder};
 use dunce::canonicalize;
 use log::{info, warn};
 #[cfg(target_os = "windows")]
+use std::process::Command;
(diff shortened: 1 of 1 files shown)


sr2500836773:
Note: I was able to inspect and keep the change scoped to the Windows auto-launcher path, but I could not run `cargo fmt` / `cargo check` locally because my current environment does not have the Rust toolchain installed. Happy to address CI feedback if formatting or compile checks require changes.

brianp:

PR #3275, opened 2026-06-13
fix: resolve Windows auto-start bug by using spawn_blocking and local…

dmthong181096:
### Description

This PR fixes issue #1588 where Tari Universe fails to register as an auto-start application on Windows system boot. 

Since Tari Universe runs as Administrator (`requireAdministrator` in manifest) to handle firewall prompts, it cannot auto-start via the standard `HKEY_CURRENT_USER\...\Run` registry key (as Windows security blocks startup registry items that require elevation). Auto-start must be achieved using the Windows Task Scheduler, which was failing due to several Windows-specific issues:

1. **COM Apartment Model Conflict**: The task scheduler registration runs on the main Tauri thread (which is initialized as STA for WebView2). `planif`'s `TaskScheduler::new()` tries to initialize COM as MTA (`COINIT_MULTITHREADED`), failing with `RPC_E_CHANGED_MODE` (`0x80010106`).
   - *Fix*: Wrapped the registration logic inside `tokio::task::spawn_blocking` to execute it on a background thread where COM is not initialized yet.
2. **Username Mapping Resolution Error (`0x80070534`)**: Using `whoami::username()` to populate the `user_id` inside `PrincipalSettings` fails on Windows 11 systems where the user is signed in with a Microsoft Account (causing a mismatch between user name and SID).
   - *Fix*: Set `user_id` to `None`. For an interactive logon trigger, leaving `user_id` blank defaults it to the currently logged-on user at logon time, bypassing username translation errors.
3. **UNC Path Prefix**: Canonicalized executable paths sometimes start with the `\\?\` prefix, which can cause registry or execution issues in Task Scheduler.
   - *Fix*: Trimmed the `\\?\` prefix from the start of the `app_path` string.
4. **CI Warnings**: Removed the unused `whoami::username` import to prevent compiler warning build failures.

### Verification Plan

1. Enable "Auto-start on system boot" in the application settings.
2. Verify the task "Tari Universe startup" is created successfully and is visible in the Windows Task Scheduler.
3. Disable the setting and verify the task trigger is deactivated or updated correctly.

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -35,8 +35,6 @@ use planif::{
 };
 use tauri::utils::platform::current_exe;
 use tokio::sync::RwLock;
-#[cfg(target_os = "windows")]
-use whoami::username;
 
 use crate::{
     LOG_TARGET_APP_LOGIC,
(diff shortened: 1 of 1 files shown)


brianp:

PR #3307, opened 2026-06-29
fix: delete Windows admin auto-start scheduled task

### LINKED ISSUE #3306: Admin auto-start scheduled task isn't deleted when 'start on system startup' is turned off
**Describe the bug**

When you turn off "start on system startup" on Windows, the regular auto-start registry value gets removed like it should, but the elevated admin auto-start scheduled task (`Tari Universe startup`) stays behind. It doesn't get deleted, its logon trigger just gets switched off. The task sits there in a Ready state pointing at whatever exe path it was last given. It won't fire on boot, so it's harmless, but it never gets cleaned up. Disabling the regular auto-start cleans up after itself. The admin one doesn't, and that's the asymmetry worth fixing.

**To Reproduce**

1. On Windows, enable "start on system startup" in the app.
2. Confirm both the `HKCU\...\Run` value `Tari Universe (Alpha)` and the scheduled task `Tari Universe startup` got created.
3. Turn the toggle back off.
4. Open the registry and Task Scheduler again.

**Expected behavior**

Both the Run-key value and the `Tari Universe startup` scheduled task get removed, the same way disabling the regular auto-start cleanly deletes its Run value.

What actually happens:

- The `HKCU\...\Run` value is deleted.
- The `Tari Universe startup` task is still there (State = Ready, Settings.Enabled = True). Only its logon trigger shows Enabled = False.

- OS & Version: Windows 11

**Additional context**

Root cause is in `src-tauri/src/auto_launcher.rs`. The disable path calls `toggle_windows_admin_auto_launcher(false)`, which calls `create_task_scheduler_for_admin_startup(false)`. That function re-registers the same task with `TaskCreationFlags::CreateOrUpdate` and a disabled trigger instead of deleting it. So "disable" really means "recreate the task with the trigger off." There's no delete call anywhere, because `planif` (the crate that creates the task) only exposes `register()`. It has no delete API.

One thing to be clear about: this is bounded, not unbounded growth. The task name is a constant and `CreateOrUpdate` overwrites in place, so you only ever end up with one stale task, not a pile of them.

This is pre-existing. It wasn't introduced by #3182, which only added path quoting to the existing Run-key and Task Scheduler logic. I found it while reviewing that PR.

Suggested fix: in the disable branch, actually delete the task instead of re-registering it with a dead trigger. Since `planif` can't delete, that means either shelling out to `schtasks /Delete /TN "Tari Universe startup" /F` or calling `ITaskFolder::DeleteTask` over COM. Make it idempotent so it doesn't error when the task is already gone.

Related: #3182

samrusani:
## Description

Fixes #3306.

Disabling Windows auto-start now deletes the elevated `Tari Universe startup` scheduled task instead of re-registering it with a disabled trigger.

The normal auto-launch registry cleanup still runs as before. The admin scheduled-task delete path now checks for idempotency by querying the task after a failed delete instead of parsing localized `schtasks` output.

## Motivation and Context

The previous disable path called `create_task_scheduler_for_admin_startup(false)`, which used `TaskCreationFlags::CreateOrUpdate` and left the task behind with only the logon trigger disabled. That made disabling admin auto-start asymmetric with the regular Run-key auto-start cleanup.

## How Has This Been Tested?

- `cargo fmt --check`
- `git diff --check`
- `cargo test --locked auto_launcher`
- `cargo check --locked`

Additional note: I attempted `cargo check --locked --target x86_64-pc-windows-msvc`, but this local macOS toolchain does not have that Rust target installed (`E0463: can't find crate for core`). The shared command-argument helpers are covered by the focused tests, and the delete path now uses `schtasks /Query` after delete failure to avoid localized error-string parsing. I could not run the actual Windows Task Scheduler command from this environment.

On Windows:

1. Enable "start on system startup" in Tari Universe.
2. Confirm both the regular Run-key value and the `Tari Universe startup` scheduled task exist.
3. Disable "start on system startup".
4. Confirm the Run-key value is removed.
5. Confirm `schtasks /Query /TN "Tari Universe startup"` no longer finds the scheduled task.
6. Disable the setting again and confirm it does not error when the scheduled task is already absent.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src-tauri/src/auto_launcher.rs
+++ b/src-tauri/src/auto_launcher.rs
@@ -226,14 +226,25 @@ impl AutoLauncher {
             return Ok(());
         }
 
-        let stdout = String::from_utf8_lossy(&output.stdout);
-        let stderr = String::from_utf8_lossy(&output.stderr);
-        let output_text = format!("{stdout}\n{stderr}");
+        let task_exists = match tokio::process::Command::new("schtasks")
+            .args(windows_admin_auto_start_query_args())
+            .output()
+            .await
+        {
+            Ok(output) => output.status.success(),
+            Err(error) => {
+                warn!(target: LOG_TARGET_APP_LOGIC, "Failed to query admin auto-start task after delete failure: {error}");
+                true
+            }
+        };
 
-        if is_missing_admin_auto_start_task_error(&output_text) {
+        if !task_exists {
             info!(target: LOG_TARGET_APP_LOGIC, "Task scheduler for admin startup was already absent");
             return Ok(());
         }
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let stderr = String::from_utf8_lossy(&output.stderr);
 
         Err(anyhow!(
             "schtasks failed with status {:?}. stdout: {} stderr: {}",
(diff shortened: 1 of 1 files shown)


brianp:
One question on the disable path: the task is created with RunLevel::Highest. If a user disables auto-start without running elevated, does `schtasks /Delete /F` still succeed, or could it fail with access-denied and leave the task behind? The call site only logs the error, so a failure there would be silent. Did the old COM update path delete reliably in non-elevated contexts?

samrusani:
Good point. I could not verify the non-elevated `RunLevel::Highest` deletion behavior locally from macOS, so I changed the failure handling rather than assuming `schtasks /Delete /F` always succeeds.

The disable path now:
- Attempts to delete the admin scheduled task.
- Still disables the regular auto-launch entry.
- Returns the admin scheduled-task deletion error to the caller if the task still exists after the delete attempt, instead of only logging it.

The already-absent task case remains treated as success via the locale-independent `schtasks /Query` fallback.

Validation run locally:
- `cargo fmt --check`
- `cargo test --locked -p tari-universe auto_launcher`
- `cargo check --locked -p tari-universe`
- `git diff --check`

GitGuardian is passing and the branch is up to date with `main`.

brianp:
Cheers mate. Tested on windows and things look to be working good

PR #3333, opened 2026-07-26
fix(block-time): clamp elapsed time to zero on clock skew

DPS0340:
## The bug

`calculateTimeSince` assumes the block timestamp is always in the past. When it is not — ordinary node/peer clock skew, or a local clock that has drifted — the diff goes negative, and `Math.floor` rounds negative values **away** from zero. The returned fields then disagree with one another:

```
block 5s in the future ->  days: -1, hours: -1, minutes: '-1', seconds: '-5'
block 2h in the future ->  days: -1, hours: -2, minutes: '00', seconds: '00'
```

**This reaches the dashboard.** `BlockTime.tsx` renders `minutes` and `seconds` straight through:

```tsx
{minutes?.split('').map(...)}:{seconds?.split('').map(...)}
```

so the current block time displays as **`-1:-5`**. Five seconds of skew is enough; nothing exotic is required.

`DebugSettings.tsx` consumes the same hook, and the `days: -1` also makes `daysString` inconsistent — it is empty (guarded by `days > 0`) while `days` itself is negative, so any caller reading the number rather than the string sees a negative day count.

## The fix

One line — clamp the diff at zero:

```ts
const diff: number = Math.max(0, later - past.getTime());
```

The whole struct then reads `00:00` while the local clock catches up, which is already what the timer shows for a brand-new block. No new state, no branch in the component.

## Verification

The two new tests fail on `main` and pass with the fix — I checked by reverting the change rather than assuming:

```
reverted:  src/utils/calculateTimeSince.test.ts (22 tests | 2 failed)
with fix:  22 passed
```

Full suite: **45 files, 1059 tests, all passing.** `tsc --noEmit` and `eslint` report nothing on either changed file.

I put the tests in the existing `edge cases` block next to the large-difference case, since this is the same shape of input assumption.


DPS0340:
Heads-up on the checks, since the PR page is misleading here: **the five workflows are sitting in `action_required`, not failing.** GitHub holds workflow runs for contributors with no merged commit in the repo, so they have never started:

```
completed/action_required   PR
completed/action_required   Lint
completed/action_required   CI
completed/action_required   Playwright Tests
completed/action_required   Security Audit - Daily
```

The one green check (GitGuardian) runs outside that gate, which makes the list look populated rather than blocked.

So I ran the jobs that apply to this change locally, using the same commands the workflows use:

**`Lint` → `pnpm run lint`** (`eslint src/ && knip && tsc`)

```
$ pnpm run lint
Configuration hints (2)
[src/**/*.{js,ts,tsx}, …]    knip.ts  Remove, or move unused top-level project ...
commitlint                   knip.ts  Remove from ignoreBinaries
exit 0
```

Both hints are pre-existing `knip.ts` configuration notes, unrelated to this diff.

**`CI → Frontend Tests` → `pnpm run test` and `pnpm run test:coverage`**

```
Test Files  45 passed (45)
     Tests  1059 passed (1059)
```

Coverage run is identical, and `src/utils/calculateTimeSince.test.ts (22 tests)` passes in it.

The Rust jobs (`cargo fmt/clippy/check/test`) are untouched by this PR — it changes one TypeScript utility and its test file.

And the part that matters most, which no gate can show you: **the two new tests fail without the fix.** I checked by reverting the one-line change rather than assuming:

```
reverted:  src/utils/calculateTimeSince.test.ts (22 tests | 2 failed)
with fix:  22 passed
```

Approving the workflow runs would give you the real results; until then this is the closest evidence I can produce.

brianp:
Thanks for the contribution!

PR #3334, opened 2026-07-26
fix(countdown): clear the previous timer before starting a new one

DPS0340:
## The bug

`useCountdown` overwrites its timer refs without clearing what is already there, and `stop()` clears them without nulling the refs:

```ts
const startCountdown = useCallback((duration: number) => {
    setCountdown(duration);
    countdownInterval.current = setInterval(...);   // previous interval never cleared
}, []);

const stopCountdown = useCallback(() => {
    if (countdownInterval.current) {
        clearInterval(countdownInterval.current);   // ref keeps pointing at a dead timer
    }
}, []);
```

`startConnectionRetry` / `stopConnectionRetry` have the identical shape.

**`DisconnectWrapper` reaches it.** On the `disconnected-severe` branch it calls `startCountdown(300)` while the effect-driven countdown from the `[attempt, connectionStatus]` effect is still live:

```tsx
if (connectionStatus === 'disconnected' && currentAttempt === retryBackoff.length) {
    setConnectionStatus('disconnected-severe');
    setIsReconnectOnCooldown(false);
    startCountdown(300);   // no stopCountdown() first
    return;
}
```

Two symptoms, both user-visible on a disconnected client:

1. **The countdown ticks twice per second** — the retry timer shown to the user drains at double speed.
2. **The abandoned `setTimeout` still fires**, so `invoke('reconnect')` runs once per stacked retry instead of once. That is the worse of the two: overlapping retries stack reconnect attempts rather than replacing them.

## The fix

Clear before replacing, and null the refs on stop. No behaviour change for a single start/stop cycle.

## Verification

There were no tests for this hook, so the PR adds seven, driven through the real hook with `renderHook` and fake timers rather than a model of it. **Two of them fail on `main`** — I checked by reverting the hook and re-running:

```
× does not run two intervals when start is called twice
  → expected 58 to be 59
× fires reconnect once when start is called twice
  → expected "spy" to be called 1 times, but got 2 times
2 failed | 5 passed
```

With the fix: 7 passed. The other five pin the behaviour that already worked (per-second ticking, stopping at zero, stop() halting the timer, restart after stop, no reconnect after stop) so a future change to this hook cannot regress them silently.

Full suite: **46 files, 1066 tests, all passing.** `pnpm run lint` (`eslint && knip && tsc`) exits 0 — the two `knip.ts` configuration hints are pre-existing and unrelated.

Note: my workflow runs are held in `action_required` as a first-time contributor here, so the commands above are what I could run locally.


brianp:
Thanks for the contribution!

PR #3335, opened 2026-07-28
chore: update .nvmrc to lts/* to match CI Node version

rvnminers-A-and-N:
## Description

Sets `.nvmrc` to `lts/*` (was pinned to `v20.17.0`).

## Motivation and Context

`.nvmrc` pinned `v20.17.0`, but `package.json` pins `packageManager` to `pnpm@11.9.0`, which requires **Node >= 22.13** — pnpm 11 imports `node:sqlite`, which does not exist in Node 20.

Following `.nvmrc` with `nvm use` therefore produces a toolchain that cannot run the project's own package manager:

```
$ nvm use
Now using node v20.17.0 (npm v10.8.2)

$ pnpm --version
warn: This version of pnpm requires at least Node.js v22.13
Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:sqlite
```

CI doesn't catch this because **no workflow reads `.nvmrc`** — all seven use `node-version: lts/*` via `actions/setup-node`. So the file had drifted without anything failing.

I hit this setting up to work on GPU miner support, and it's the first thing a new contributor runs into.

### Why `lts/*` rather than a pinned version

Pinning a specific version (e.g. `v24.18.0`) fixes today's symptom but re-arms the same trap the next time Node moves. `lts/*` matches what CI already resolves, so the two can't drift apart again.

Nothing in the repo consumes `.nvmrc` programmatically — no workflow, no script, and the only Dockerfile (`docs/containerization/Dockerfile`) is an Ubuntu runtime image with no Node in it — so the alias is safe here. `nvm` resolves it natively.

Trade-off worth naming: `lts/*` isn't reproducible across time, since two developers months apart could land on different majors. Happy to switch to a pinned `v24.18.0` if you'd prefer determinism over drift-resistance.

### Optional follow-up

Pointing `actions/setup-node` at `node-version-file: .nvmrc` instead of `node-version: lts/*` would make CI consume this file, so it could never silently drift again. Left out of this PR to keep it to one line, but happy to add it if wanted.

## How Has This Been Tested?

On Ubuntu 24.04, from a clean clone:

```
$ nvm use
Found '/home/crypto/tari/universe/.nvmrc' with version <lts/*>
Now using node v24.18.0 (npm v11.16.0)

$ pnpm --version
11.9.0
```

Then the full local gate:

- `pnpm install --frozen-lockfile` — clean
- `pnpm test` — 1057 tests across 45 files, all passing
- `pnpm lint` (eslint + knip + tsc) — clean
- `cargo fmt --all -- --check` — clean
- `cargo lints clippy --all-targets --all-features` — no new findings
- `cargo check --release --all-targets --workspace` — clean

```bash
nvm install   # if the LTS isn't already installed
nvm use       # should select the current LTS, not v20.17.0
pnpm --version
```

Before this change `pnpm --version` exits non-zero with `ERR_UNKNOWN_BUILTIN_MODULE`; after, it prints `11.9.0`.

## Breaking Changes

- [x] None


brianp:
Thanks for the contribution!

Issue #2893, opened 2025-09-17
v1.6.1: Windows 11: Mining and Mining Pools: When I turn off GPU Pool Mining and then turn it back on, I find that I am still on `Glytex` and it has not reverted back to `lolMiner` or `graxil`


Tas4tari:
**Describe the bug**
When I turn off GPU Pool Mining and then turn it back on, I find that I am still on `Glytex` and it has not reverted back to `lolMiner` or `graxil`
I did the above last night and when I opened TU today, I have observed that I am still on `Glytex`

**Pre-conditions**
- User should have TU v1.6.1 installed
- User should have fallen back to `Graxil` as `lolMiner` is not supported

**Steps to reproduce the behavior:**
1. Go to 'Settings'
2. Click on 'Mining Pools'
3. Scroll down to and turn off the toggle for 'GPU Pool'
4. Confirm `GPU Pool` is disabled
5. Click on `Mining`
6. View `GPU Miners`
7. Confirm `GPU Miner` is on `Glytex`
8. Go back to `Mining Pools`
9. Scroll down to and switch the toggle for 'GPU Pool' to enabled
10. Confirm `GPU Pool` is enabled
11. Click on `Mining`
12. View `GPU Miners`
13. Observe

**Expected behavior**
Once `GPU Pool` is re-enabled, `GPU Miner` should have switched back to either `lolMiner` or `Graxil`

**Screenshots**
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c0aad865-00af-48f1-a02f-8f64ef2ae23a" />

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11

**Additional context**
[logs.zip](https://github.com/user-attachments/files/22378858/logs.zip)

Tas4tari:
Tested on v1.6.2, when I enable GPU Pool, I am switched back to `lolMiner` which then falls back to `Graxil`

Issue #2866, opened 2025-09-11
v1.5.22 - Beta - 0911: Windows 11: Mining Settings: When I switch from `Graxil` to `lolMiner` I find that it falls back to `Glytex`


Tas4tari:
**Describe the bug**
As a user I find that when I switch to `lolMiner`, I find that it falls back to `Glytex` and not `Graxil`

**Pre-conditions**
- user should have done a clean installation of the Beta build
- User should have TU open and running

**Steps to reproduce the behavior:**
1. Go to 'Mining'
2. Click on 'GPU Engines'
3. Scroll down to & click on 'lolMiner'
4. Go to Resource Monitor and observe
5. Go back to TU and view `Mining`
6. Observe the `GPU Engines` selection

**Expected behavior**
When `lolMiner` is not supported, the miner should fall back to `Graxil` and not `Glytex`

**Screenshots**
<img width="1616" height="950" alt="Image" src="https://github.com/user-attachments/assets/caa25516-34a9-4faf-ab89-fbedc908d436" />
<img width="1528" height="901" alt="Image" src="https://github.com/user-attachments/assets/737c8e46-226b-47c6-8be8-d29c0de5dbb8" />
<img width="1638" height="964" alt="Image" src="https://github.com/user-attachments/assets/6eac3791-afe0-4267-bd2a-24a224fe3b0c" />

**Desktop (please complete the following information):**
 - Aspire A315-59
 - Windows 11

**Logs:**

[logs.zip](https://github.com/user-attachments/files/22279993/logs.zip)

Tas4tari:
On Windows 11, app version 1.5.23. I find that when I switch from `Glytex` to `lolMiner`, it does not fall back to `Graxil`. It stays on `lolMiner`

<img width="1639" height="982" alt="Image" src="https://github.com/user-attachments/assets/1d89e95f-0735-4f98-8382-74518b777ac7" />

<img width="1683" height="934" alt="Image" src="https://github.com/user-attachments/assets/e24d064c-58b6-44bc-bdce-33cf6574cf1b" />

PR #3337, opened 2026-07-28
fix: gate GPU pool toggle when no miner supports solo mining

### LINKED ISSUE #3336: Disabling GPU pool mining leaves GPU mining unable to start — no installed miner supports solo mining
**Describe the bug**

Turning off the GPU pool switches GPU mining to a direct node (solo) connection. That path requires a GPU miner advertising the `SoloMining` feature — and no currently shipped GPU miner does, so GPU mining cannot start while the pool is off.

The toggle is rendered unconditionally, with no disabled state and no explanation, so nothing indicates the option cannot work.

**To Reproduce**

1. Go to 'Settings'
2. Click on 'Mining Pools'
3. Scroll down to and turn off the toggle for 'GPU Pool'
4. Start (or restart) GPU mining

**Expected behavior**

Either the GPU pool toggle is disabled with a reason while no installed miner supports solo mining, or the failure is surfaced clearly rather than leaving mining stopped. Offering a control that cannot succeed seems to be the core issue.

**Additional context**

`src-tauri/src/mining/gpu/manager.rs` branches on the toggle:

```rust
if *ConfigPools::content().await.gpu_pool_enabled() {
    self.handle_pool_connection_load().await?;
} else {
    let grpc_node_address = app_state.node_manager.get_grpc_address().await?;
    self.handle_node_connection_load(grpc_node_address).await?;
}
```

`handle_node_connection_load` requires the selected miner to report `is_solo_mining_supported()`, then searches `MINERS_PRIORITY` for a healthy alternative that does. If neither is found it returns:

```
Selected gpu miner does not support solo mining and no other miners are available
```

Nothing satisfies that today:

- `GpuMinerFeature::SoloMining` is declared by no miner — `consts.rs` gives `LolMiner` `[PoolMining, DeviceExclusion]`
- `MINERS_PRIORITY` contains only `LolMiner`
- `lolminer.rs` rejects node mining explicitly: `"Lolminer does not support node mining"`

On the frontend, `GpuPoolsSettings.tsx` renders the switch with no `disabled` prop and no conditional, so it is always interactive.

This looks like a side effect of removing Glytex and Graxil along with SHA-3X — #2866 and #2893 both show those miners in a working "GPU Engines" selection UI at a time when solo mining was supported.

I'm happy to open a PR gating the toggle on the miner registry (so it re-enables automatically if a solo-capable miner is ever added), following the Requirement/Recommendation split from #3330 — unless dropping solo GPU mining was deliberate, in which case removing the toggle may be the better fix. Guidance welcome either way.

- OS & Version: Ubuntu 24.04, `main` @ 18784ae

Found by code inspection rather than on a GPU machine, so I have not confirmed exactly what the user sees when mining fails to start — the code path above is deterministic, but I did not want to describe symptoms I have not observed.

rvnminers-A-and-N:
## Description

Closes #3336.

Locks the GPU pool toggle, with a reason, while no installed GPU miner supports solo mining — and refuses the same transition in `toggle_gpu_pool_mining` so the command cannot be used to reach the state either.

## Motivation and Context

Turning the GPU pool off switches GPU mining to a direct node connection:

```rust
if *ConfigPools::content().await.gpu_pool_enabled() {
    self.handle_pool_connection_load().await?;
} else {
    let grpc_node_address = app_state.node_manager.get_grpc_address().await?;
    self.handle_node_connection_load(grpc_node_address).await?;
}
```

`handle_node_connection_load` requires a miner advertising `SoloMining`, then searches `MINERS_PRIORITY` for a healthy alternative. Nothing satisfies either today — `LolMiner` advertises `[PoolMining, DeviceExclusion]`, it is the only entry in `MINERS_PRIORITY`, and `lolminer.rs` rejects node mining outright — so it returns:

```
Selected gpu miner does not support solo mining and no other miners are available
```

The toggle was rendered unconditionally, with no disabled state and no explanation, so nothing signalled that the option could not work.

### Approach

This follows the Requirement/Recommendation split from #3330 — a capability that cannot be satisfied locks the control and shows the reason, rather than failing after the fact.

The gate is **computed from the miner registry, not hardcoded**, on both sides:

```rust
self.available_miners.values().any(|m| m.features.contains(&GpuMinerFeature::SoloMining))
```

So if a solo-capable miner is ever added, both gates open on their own with no gating code to edit. Both sides read each miner's advertised `features`, so frontend and backend evaluate the same predicate against the same data that is emitted to the UI.

The selector returns `undefined` until the miner list arrives, distinguishing "not supported" from "not known yet" so an unloaded list never gates the toggle spuriously.

This looks like a side effect of removing Glytex and Graxil along with SHA-3X — #2866 and #2893 both show those miners in a working "GPU Engines" selection UI at a time when solo mining was supported. If dropping solo GPU mining was deliberate and the toggle should go entirely instead, happy to take that direction.

## How Has This Been Tested?

Ubuntu 24.04. Full local gate:

- `cargo fmt --all -- --check` — clean
- `cargo lints clippy --all-targets --all-features` — no new findings (the pre-existing `useless_borrows_in_formatting` warnings are untouched, and none are in the files this changes)
- `cargo test --bin tari-universe` — **213 passed** (207 before this change)
- `pnpm test` — **1062 passed** across 45 files (1057 before)
- `pnpm lint` — clean
- `playwright/` `pnpm run typecheck` — clean

New coverage:

- **Selector** (5): unknown miner list, no solo support, empty list, solo support present, solo support on one of several
- **`GpuManager::is_solo_mining_supported`** (4): no miners, no miner advertising it, a miner advertising it, and an unhealthy miner still advertising it
- **Tripwires** (2): no shipped miner supports solo mining, every shipped miner supports pool mining. These pin the assumption both gates rest on — if a miner gains solo support they fail deliberately and point at the gating to revisit

### The end-to-end test has not asserted yet

The new spec in `09-settings-sweep.spec.ts` **skipped** on my machine, which has no GPU: the GPU pool section only renders once the GPU mining module has initialised, so it never became visible. It follows the conditional pattern the existing pools sweep already uses for the same reason.

So it is written and typechecked but **unverified** — its assertions have never executed. I would rather say that than let a green run imply more than it does. If it misbehaves on GPU hardware or in your CI, tell me and I will fix it up; I am happy to drop it from this PR if you would rather it landed separately once confirmed.

On a machine where GPU mining initialises:

1. Go to 'Settings'
2. Click on 'Mining Pools'
3. Observe the 'GPU Pool' toggle

Expected: the toggle is disabled and shows *"Required because no installed GPU miner supports solo mining"*. Before this change it was interactive, and turning it off left GPU mining unable to start.

The backend guard can be exercised by invoking `toggle_gpu_pool_mining` with `enabled: false` directly; it now returns an error instead of entering the unusable state.

`cargo test --bin tari-universe gpu` runs the Rust coverage without a GPU.

## Breaking Changes

- [x] None


src/containers/floating/Settings/sections/pools/GpuPoolsSettings.tsx
```diff
@@ -64,11 +70,13 @@ export const GpuPoolsSettings = () => {
                         <Typography variant="h6">{'GPU pool'}</Typography>
                     </SettingsGroupTitle>
                     <Typography>{t('mining-toggle-warning')}</Typography>
+                    {isSoloMiningUnsupported && <Typography>{t('gpu-solo-mining-unavailable')}</Typography>}
                 </SettingsGroupContent>
 
                 <SettingsGroupAction>
                     <ToggleSwitch
                         checked={isGpuPoolEnabled}
+                        disabled={isSoloMiningUnsupported}
```
brianp, line 79: This locks the switch in both directions, but the backend guard only refuses turning the pool off. If a user already has `gpu_pool_enabled` persisted as `false` (setup forces it off on solo networks, and older configs can carry it), the switch renders off and disabled with no way back on. Gate on the current value as well:

```tsx
disabled={isSoloMiningUnsupported && isGpuPoolEnabled}
```

Same condition for the warning text above, it only applies while the pool is on.

brianp (commented):
Thanks for contributing.

PR #3340, opened 2026-08-09
fix(wallet): start sync countdown when estimate is ready

### LINKED ISSUE #3339: Wallet sync countdown never starts
**Describe the bug**
When a sync estimate becomes available, `SyncCountdown` remains on the calculating-time message instead of mounting and starting the countdown.

`startedRef` is initialized to `false`, so the component returns `LoadingText` before rendering `<Countdown>`. As a result, `countdownRef.current` remains `null`, the effect cannot start the countdown, and `onStarted` is never called.

This affects both current production call sites:

- the wallet balance while disconnected/syncing
- the mining tile sync status

Affected revision: `18784ae64f0e7c775a27388b81f73f110b443cff` (`main` at the time of reproduction)

**To Reproduce**
1. Add a focused component test for `SyncCountdown`.
2. Mock `useProgressCountdown` to return `{ countdown: 120 }`.
3. Mock `react-countdown` with a test sentinel and render `SyncCountdown`.
4. Observe that the calculating-time message is rendered, the countdown sentinel is absent, and `onStarted` is not called.

The behavior was reproduced with:

```bash
./node_modules/.bin/vitest run src/components/wallet/components/loaders/SyncLoading/SyncCountdown.audit.test.tsx
```

Result: one test passed, confirming the countdown is not mounted or started when an estimate is available.

**Expected behavior**
Once `useProgressCountdown` returns a valid estimate, the countdown should mount and start, `onStarted` should be called, and the calculating-time message should be replaced by the remaining sync time.

**Screenshots**
N/A. Reproduced deterministically with an automated component test.

- OS & Version: Not OS-specific; reproduced in the repository's jsdom test environment
- Browser & Version: N/A (Tauri renderer component)

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser & Version: N/A

**Additional context**
The early return in `src/components/wallet/components/loaders/SyncLoading/SyncCountdown.tsx` prevents the ref used by the effect from ever being mounted. A regression test should cover the transition from calculating time to an active countdown.

samrusani:
## Description

Mounts `SyncCountdown` as soon as `useProgressCountdown` returns a valid estimate instead of waiting for `startedRef`, which can only be updated after the countdown component mounts.

Adds a focused regression test covering the transition from an unavailable estimate to an active countdown.

Closes #3339

## Motivation and Context

The previous loading guard returned before `<Countdown>` could mount. As a result, `countdownRef.current` stayed `null`, the start effect could not call the countdown API, and `onStarted` was never invoked.

The new guard uses the hook's existing `-1` sentinel for an unavailable estimate. Once an estimate is available, the countdown mounts and the existing start/complete behavior runs unchanged.

## How Has This Been Tested?

- `pnpm exec vitest run src/components/wallet/components/loaders/SyncLoading/SyncCountdown.test.tsx`
- `pnpm test` (46 test files, 1,058 tests passed)
- `pnpm lint`
- `pnpm build`
- `git diff --check`

All commands exited successfully. The Vitest runs retain the repository's existing Vite shutdown-timeout notice after successful test completion.

1. Run the focused Vitest command above.
2. Confirm the component initially shows the calculating-time state while the mocked estimate is `-1`.
3. Confirm that changing the estimate to `120` mounts and starts the countdown, invokes `onStarted`, and renders `2m`.

No manual or hardware-dependent testing is required for this state-transition fix.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/src/components/wallet/components/loaders/SyncLoading/SyncCountdown.test.tsx
+++ b/src/components/wallet/components/loaders/SyncLoading/SyncCountdown.test.tsx
@@ -1,0 +1,43 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const countdownMock = vi.hoisted(() => ({
+    value: -1,
+}));
+
+vi.mock('react-i18next', () => ({
+    useTranslation: () => ({
+        t: (key: string) => key,
+    }),
+}));
+
+vi.mock('@app/containers/main/Sync/components/useProgressCountdown.ts', () => ({
+    useProgressCountdown: () => ({ countdown: countdownMock.value }),
+}));
+
+import { render, screen, waitFor } from '@app/test/test-utils';
+import SyncCountdown from './SyncCountdown';
+
+describe('SyncCountdown', () => {
+    beforeEach(() => {
+        countdownMock.value = -1;
+    });
+
+    it('starts after a sync estimate becomes available', async () => {
+        const onCompleted = vi.fn();
+        const onStarted = vi.fn();
+        const { rerender } = render(<SyncCountdown onCompleted={onCompleted} onStarted={onStarted} />);
+
+        expect(screen.getByText('setup-progresses:calculating_time')).toBeInTheDocument();
+        expect(onStarted).not.toHaveBeenCalled();
+
+        countdownMock.value = 120;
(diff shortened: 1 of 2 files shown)


brianp:
Thanks for the contribution!

PR #3344, opened 2026-09-04
ci: cut redundant compiles, enable rust caching, use prebuilt CLIs

ns212:
## What

Performance pass over the GitHub workflow files — mechanical, low-risk changes only. Builds were slow largely because the same work was being redone (and paid for) on every run.

## Changes and estimated impact

| Fix | Where | Est. saving |
|---|---|---|
| Enable `swatinem/rust-cache` on **all** refs — the old `if` disabled caching entirely on `release` branch / `v*` tag builds, cold-compiling all 1,149 crates per production build | `release.yml`, `release-exchange.yml` `builds` | ~30–60 min **per platform per release run** |
| Add `shared-key: tari-universe` so the 4 Rust jobs in CI (and matrix entries) share one dependency cache instead of per-job caches | `ci.yml` ×3, `playwright.yml` | ~15–30 min per PR run after warm-up |
| Drop `--release` from `cargo check` — clippy already checks dev profile; the release-profile check recompiled the workspace with no extra signal | `ci.yml` `cargo-checks` | ~10–20 min per PR run |
| Use the `@tauri-apps/cli` **already in devDependencies** (`pnpm exec tauri …`) instead of `cargo install tauri-cli`, which compiled it from source every run | `ci.yml` `tauri-test-build`, `playwright.yml` | ~10 min per run |
| `machete` job: drop Rust toolchain, rust-cache and the process-wrapper build — `cargo-machete` parses manifests only; install the prebuilt static binary (sha256-verified) instead of `cargo install` | `ci.yml` `machete` | ~5–10 min per PR run |
| Frontend tests ran the whole suite twice (`pnpm run test` then `test:coverage`, which is just `vitest run --coverage`); keep only the coverage run | `ci.yml` `frontend-tests` | ~halves that job |
| `pnpm audit` resolves against the lockfile — drop the full `pnpm install` | `audit.yml` | ~1 min per run |
| Dead weight: redundant `rustup component add` (already installed via the toolchain action's `components:`), ripgrep via wget+dpkg → apt, needless checkout in the jq-only `read-exchanges` job, duplicated yq/env lines, duplicated install-unzip step | `ci.yml`, `release-exchange.yml` | seconds, but free |

Total rough estimate: **30–50 min off a typical PR run and 1–2 h off a release run**, no architecture changes.

## Verification

- `actionlint` clean: 5 findings before and after, all pre-existing (self-hosted runner label, optional matrix keys), none introduced
- cargo-machete download/checksum/tarball layout tested locally (the sidecar is a bare hash; the check was adapted)
- All YAML parses
- `cargo test --release` kept deliberately — release-profile tests may be needed for runtime speed; flagging as a possible follow-up

## Deliberately left out (need coordination, listed as follow-ups)

1. **Path filters** for PR-triggered workflows (a docs-only PR still runs ~10 jobs incl. the Tauri build + e2e suite). Skipping jobs interacts with required status checks / merge queue, so it needs branch-protection coordination.
2. **Nightly cron** in `release.yml` builds the full Windows+macOS matrix unconditionally (`schedule` ignores `paths-ignore`) — worth gating on new commits.
3. **Cache-warming job** for `release-exchange` — `EXCHANGE_ID` is a leaf-crate `env!()`, so per-exchange matrix entries could warm deps once per platform and relink per exchange.
4. Windows vcpkg binary caching; macOS non-host toolchain installs (likely removable, needs verification on a mac runner).


ns212:
**Re: 'Cargo checks' failure on the first push — root cause + fix**

The `needless_late_init` error was **pre-existing on main**, not caused by the workflow changes:

- The late-initialized `archive_destination_path` pattern has been on main since `10dad06c` (Jul 2025)
- clippy **1.98** now flags it (denied via `clippy::style` in `lints.toml`)
- This PR's CI run was the first to run `cargo lints clippy` against main's current code on 1.98 — the last green CI run (Sept 1) was on the `sw_incorporate_c29_miner` branch, whose tree already contains the initialized-in-`if` form
- No workflow change in this PR touches the clippy invocation or any Rust source

a942cd9a restores the initialized-in-`if` form **verbatim from that branch** — identical semantics, and it's the exact code that passed clippy 1.98 on Sept 1. Note for maintainers: any other PR opened from current main will hit this same error until this (or an equivalent fix) lands.

